PR #1608, opened 2014-08-10
Extract-o-matic: Extract event handlers to page JS

Shadowfiend:
The main functionality here moves event handler attachment from inline attributes to the page JavaScript that Lift 3 now supports. The reason for this is so that, out of the box, Lift will be compatible with very restrictive [Content Security Policy](http://en.wikipedia.org/wiki/Content_Security_Policy) settings even when using built-in Lift features.

To do this, we add a `lift.onEvent` function that has both jQuery and vanilla implementations. There are also a couple of fixes to existing JS functionality here.

Still two things missing that I know of:
 - [x] Extract `javascript://` `href`s from `a` elements into event handlers (these should `preventDefault()`).
 - [x] Extract `javascript://` `action`s from `form` elements into event handlers (these should also `preventDefault()`).

Looking for some early thoughts/feedback from folks.

### DIFF
--- a/web/webkit/src/main/resources/toserve/lift.js
+++ b/web/webkit/src/main/resources/toserve/lift.js
@@ -64,6 +64,9 @@
       ajaxGet: function() {
         consoleOrAlert("ajaxGet function must be defined in settings");
       },
+      onEvent: function(elementOrId, eventName, fn) {
+        consoleOrAlert("onEvent function must be defined in settings");
+      },
       onDocumentReady: function(fn) {
         consoleOrAlert("onDocumentReady function must be defined in settings");
       },
@@ -582,6 +585,7 @@
         doCycleIn200();
       },
       logError: function() { settings.logError.apply(this, arguments) },
+      onEvent: function() { settings.onEvent.apply(this, arguments) },
       ajax: appendToQueue,
       startGc: successRegisterGC,
       ajaxOnSessionLost: function() {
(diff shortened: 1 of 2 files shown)


Shadowfiend:
This fella's ready for a look, though I'm thinking we may want to do an M3 release before this goes in. This'll change behavior in a few ways and it'll be good to have a stable fallback in case the changed behavior breaks things.

fmpwizard:
@Shadowfiend the process we have in place to spin builds expects to either build from master or a branch (the branch hack was so I could build 2.6). I see that you were hoping to include this one on 3.0-M4 and not in M3, but it is now part of master, if this was the last PR merged into master, could you create a branch, something like `3.0-M3_pre` which would not include this PR and then I can spin a build based on that new branch. I like the idea of having `-M3` somewhat more stable than the m$ that will/may break things

Shadowfiend:
Waiiiiit… When in the world did this make it into master??

Shadowfiend:
BAH. I cherry-picked the CombParserHelperSpec onto the wrong branch. Hmm… Any complaints if I revert that merge altogether? This branch didn't get any review and I think it should before we merge it into master.

fmpwizard:
+1 to revert it

Shadowfiend:
Okay, reverted that and opened a new PR with these changes. I'll happily accept the award for screwup of the week, thank you…

fmpwizard:
I believe the old traddition in Lift land was this: http://twitpic.com/3zx3h0 :)

Shadowfiend:
Yeah, I was considering asking @dpp to :shipit: over here.

PR #1619, opened 2014-08-24
Comet fixes

pbrant:
This fixes a handful of bugs that when taken together kept things mostly working!

  - Unify lastListenerTime (which was previously always zero) with
    lastListenTime
  - Make lastRenderTime, delta timestamps, and lastListenerTime
    consistent with each other as they are directly compared
  - More accurately track whether any deltas have been sent
  - Pretend initial render happened at creation time to prevent a second
    render call from running on the initial Listen message
  - Rename devMode to alwaysReRenderOnPageLoad

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/CometActor.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/CometActor.scala
@@ -325,6 +325,7 @@ trait LiftCometActor extends TypedActor[Any, Any] with ForwardableActor[Any, Any
    * @return the last when sent from the listener
    */
   def lastListenerTime: Long
+  def lastRenderTime: Long
 
   private[http] def callInitCometActor(creationInfo: CometCreationInfo) {
     initCometActor(creationInfo)
@@ -451,7 +452,7 @@ trait CometActor extends LiftActor with LiftCometActor with CssBindImplicits {
   private val logger = Logger(classOf[CometActor])
   val uniqueId = Helpers.nextFuncName
   private var spanId = uniqueId
-  @volatile private var lastRenderTime = millis
+  @volatile private var _lastRenderTime = Helpers.nextNum
 
   /**
    * If we're going to cache the last rendering, here's the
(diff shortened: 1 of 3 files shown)


fmpwizard:
Any news on this PR? I saw the other day a report that we are missing updates sent from a cometactor and Antonio pointed to this PR as the possible solution.
Thanks

Shadowfiend:
Hmm… Yeah, it's a tough call here. I think some of the solutions/design aren't really optimal (that's the discussion I had with Peter above). But if Peter doesn't have time to improve the solution right now, we have to decide if we're okay with the suboptimal solution going in. If we do that, I think we should expect/plan for that solution to change by the final release of Lift 3… So basically we'd have one solution (implementing `partialUpdateStream_?`) for M3, and potentially another one (a new base trait) for M4.

One possible compromise is to immediately mark `partialUpdateStream_?` as deprecated and indicate that there is a cleaner solution coming, but that this is the thing to use until that happens.

fmpwizard:
Considering that right now Lift 3.0 comet support isn't working as we are used to (missing updates is pretty serious), I would say that the compromise solution of deprecating it as it goes in sounds good, so users will know what to expect on the following milestone.

pbrant:
Sorry, just saw this (the old problem of notifications going to a separate label rears its ugly head again).

I'm afraid I haven't had a chance to come back to this. I'm honestly not completely sure it's worth it. I agree with Antonio in principle, but on the other hand, it's a few lines out of a 1000+ lines of sometimes very dense code.

One compromise that would be easy to implement would be to make `partialUpdateStream_?'` lift-private, rename `CometActor` to e.g. `PartialUpdateCometActor` (suggestions on the name welcome) with `partialUpdateStream_? = true` and then have `CometActor` override it to false. That would at least give us a stable API going forward (albeit with the current implementation).

Thoughts?

Shadowfiend:
I dig that. I can then take a pass at reworking the whole structure for M4.

Lemme think on name and get back later tonight.

Shadowfiend:
How about `PartialUpdateCometActor` is instead `MessageCometActor` and we drop an alias for `partialUpdate` in there that is `pushMessage` or `sendMessage` (since “partial update” doesn't actually make sense if there's no “full update”, and the idea is this actor will underpin the rendering actor and its partial updates, rather than the other way around)? Also how about giving `render` a default implementation of `final def render = NodeSeq.Empty` for that actor as well?

pbrant:
Sounds good. Is this what you thinking?

Shadowfiend:
Oh yes oh yes :)

Shadowfiend:
Let's do this!

PR #1639, opened 2014-11-09
Recting and Reding: Fix ResponseShortcutException issues for AJAX calls

### LINKED ISSUE #1638: AJAX handlers can only use redirect shortcut responses
Right now, if you throw a `ResponseShortcutException` during an AJAX call, you can only use a redirect—anything else will result in a 500 error.

Shadowfiend:
Allow arbitrary ResponseShortcutExceptions in AJAX calls.

Fixes #1638 .


Shadowfiend:
Okay, going to merge this if I don't hear anything further tomorrow.

Shadowfiend:
In we go!

Issue #1645, opened 2014-12-05
Adding comets via comets doesn't work


gvdm:
As discussed in https://groups.google.com/forum/#!topic/liftweb/vxm9g_r8-qw

Not sure of the details but there is a suggested diagnosis and fix.

Shadowfiend:
Looks like this has been fixed for a while heh.

PR #1646, opened 2014-12-09
Comet, Comet on the Wall: Fix comets sent down via AJAX from within a comet

Shadowfiend:
This fixes a few issues:
 - We centralize comet command generation into `S.jsToAppend`.
 - We fix when we look up `S.jsToAppend` during AJAX processing in comet contexts so we don't miss anything that might be added during the actual AJAX function execution.
 - We make sure not to trigger a second comet request when setting up new comets client-side from within a comet response.

Fixes issue #1645 .


Shadowfiend:
K, fixed the one style thing I wanted to fix before this went in. Going to send it in tomorrow (hopefully heh) if nothing else comes up.

Shadowfiend:
Lighting this green.

PR #1648, opened 2014-12-10
This is Where the Function Stops: Map function on ajaxOnSubmit invocation.

### LINKED ISSUE #1644: S.oneShot doesn't work with SHtml.ajaxOnSubmit
As discussed in https://groups.google.com/forum/?fromgroups#!topic/liftweb/Oh7qNtj8G7g, Antonio did a great explanation why ```S.oneShot``` works with ```ajaxInvoke``` and ```ajaxSubmit``` but not ```ajaxOnSubmit```.

Shadowfiend:
Before, we bound the function when we processed the element. Because this
occurred in NodeSeq=>NodeSeq function, we had no guarantee that it would run in
the correct formGroup, oneShot, callOnce, etc contexts. To deal with the form
group part of this, we were capturing the form group and setting it when we
went to actually run the binding function. However, this left oneShot,
callOnce, and potential future contexts like them out in the dark.

We now run the fmapFunc call as soon as ajaxOnSubmit is called, and the
NodeSeq=>NodeSeq function is used solely to bind the resulting id to the
matching element(s). This means the function mapping happens in the calling
context, and we avoid all of the above issues.

There is a subtle shift in behavior: multiple matching elements will get the
same function id, where before they would get different ones. Realistically,
this shouldn't cause a big problem, since they'll be invoking the same function
anyway.

Fixes #1644 .

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/SHtml.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/SHtml.scala
@@ -1585,9 +1585,14 @@ trait SHtml {
    *  <pre>"type=submit" #> ajaxOnSubmit(() => Alert("Done!"))</pre>
    */
   def ajaxOnSubmit(func: () => JsCmd): (NodeSeq)=>NodeSeq = {
-    val fgSnap = S._formGroup.get
+    val functionId =
+      _formGroup.is match {
+        case Empty =>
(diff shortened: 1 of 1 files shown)


Shadowfiend:
Will probably merge this tomorrow in absence of further comments.

Shadowfiend:
In again!

PR #1649, opened 2014-12-11
Air Gap: Distinguish LiftSession uniqueId from underlyingId.

Shadowfiend:
What used to be uniqueId is now called underlyingId, to
indicate the fact that it is tied to the underlying host’s session
id. A new uniqueId is introduced, which is secure and unique,
and is not tied to the host id and cannot be used to look the
session up in the future.

The only place that is currently continuing to use uniqueId is
when we emit the id into the page as data-lift-session-id for
cache-busting purposes. The underlying id used to be
emitted into the page, which triggered some small security
concerns, so we switch to this model to be on the safe side.

See https://groups.google.com/forum/#!topic/liftweb/YwZgXuMOYrk
for the triggering discussion.


Shadowfiend:
Will probably merge this tomorrow in absence of further comments.

Shadowfiend:
And three!

PR #1650, opened 2014-12-15
Doc, Stock and Two Smoking HTML Tags: Cleanup pass in lift-common scaladocs

Shadowfiend:
Hopefully I'll get to do this with the other modules in short order as well. This pass
removes HTML from scaladocs, adds wiki markup in its stead, adds wiki markup
in places where we were using just text, adds examples here and there, and generally
tidies up the documentation.

It also adds the ability to link to Scala's APIs and `scala-xml` APIs from our scaladocs,
which required some tweaks to the build.

In the process, also did some very slight code cleanup. Each of these is split out into
a commit in case we want to selectively remove it. Of particular interest is that I
removed documentation that seemed to add nothing on top of the existing function
name or signature in https://github.com/lift/framework/commit/55fa831d9b848c5d37d376f56fa2a4ec82f2d8e7. Let me know if you think that's a problem.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,9 +12,9 @@ startYear in ThisBuild             := Some(2006)
 
 organizationName in ThisBuild      := "WorldWide Conferencing, LLC"
 
-scalaVersion in ThisBuild          := "2.11.2"
+scalaVersion in ThisBuild          := "2.11.3"
 
-crossScalaVersions in ThisBuild    := Seq("2.11.2")
+crossScalaVersions in ThisBuild    := Seq("2.11.3")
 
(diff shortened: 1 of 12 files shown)


Shadowfiend:
Any thoughts on the fixes for `FuncJBridge`? If those look good, I can include that commit here and then we can see about merging these changes.

fmpwizard:
looks good to me, :+1:

Shadowfiend:
Kk, pushing accordingly then. Will wait a couple of days for further comments before merging.

Shadowfiend:
Kk, seeing no further comments I'm going to go ahead and merge this feller in!

PR #1652, opened 2014-12-24
Unified parsers

joescii:
Enhancement for configuring parsers per [this thread](https://groups.google.com/forum/#!topic/liftweb/WnaUFd9Fw5E).

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
@@ -78,7 +78,7 @@ object ContentParser {
    * @param surroundFunction the function for surrounding the content returned by the `parseFunction`.
    *                         See [[defaultSurround]].
    */
-  def basic(templateSuffix: String,
+  def apply(templateSuffix: String,
             parseFunction: String=>Box[NodeSeq],
             surroundFunction: NodeSeq=>NodeSeq = defaultSurround):ContentParser =
     new ContentParser {
@@ -88,12 +88,12 @@ object ContentParser {
     }
 
   /**
-   * A full `ContentParser` which handles multiple filename suffixes, operates on an `InputStream`, and surrounds the
-   * top-level templates with the default surround
+   * A fully-specified `ContentParser` which handles multiple filename suffixes, operates on an `InputStream`, and
+   * surrounds the top-level templates with the default surround
    */
-  def full(templateSuffixesSeq: Seq[String],
+  def apply(templateSuffixesSeq: Seq[String],
            parseF: InputStream=>Box[NodeSeq],
-           surroundF: NodeSeq=>NodeSeq = defaultSurround):ContentParser =
+           surroundF: NodeSeq=>NodeSeq):ContentParser =
     new ContentParser {    
       override def templateSuffixes: Seq[String] = templateSuffixesSeq
       override def parse(content: InputStream): Box[NodeSeq] = parseF(content)
--- a/web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2007-2015 WorldWide Conferencing, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.liftweb
+package http
+
+import java.io.InputStream
+
+import net.liftweb.common.Box
+import net.liftweb.util.Helpers
+
+import scala.xml.NodeSeq
+
+/** Objects which can parse content should implement this trait. See LiftRules.contentParsers */
+trait ContentParser {
+  /** The template filename suffixes this parser is for, e.g. "html", "md", "adoc", etc. */
+  def templateSuffixes: Seq[String]
+
+  /** Given an InputStream to a resource with a matching suffix, this function should provide the corresponding
+    * NodeSeq ready to serve
+    * @return
+    */
+  def parse(content:InputStream): Box[NodeSeq]
+
+  /** Called to auto-surround the content when needed, i.e. when this content is a top-level template */
+  def surround(content:NodeSeq): NodeSeq
+}
+
+object ContentParser {
+  /** Convenience function to convert a simple String => Box[NodeSeq] content parser function into a
+    * InputStream => Box[NodeSeq] function for satisfying the ContentParser contract.
+    * @param f your String => Box[NodeSeq] content parser
+    * @return your parser wrapped up to handle an InputStream
+    */
+  def toInputStreamParser(f:String => Box[NodeSeq]):InputStream => Box[NodeSeq] = {
+    is:InputStream =>
+      for {
+        bytes <- Helpers.tryo(Helpers.readWholeStream(is))
+        elems <- f(new String(bytes, "UTF-8"))
+      } yield { elems }
+  }
+
+  val defaultSurround:NodeSeq => NodeSeq =
+    { elems => <lift:surround with="default" at="content">{elems}</lift:surround> }
+
+  /** A basic ContentParser which handles one template filename suffix, operates on a string, and surrounds the
+    * top-level templates with the default surround
+    */
+  def basic(templateSuffix:String,
+            parseF:String => Box[NodeSeq],
+            surroundF:NodeSeq => NodeSeq = defaultSurround):ContentParser =
+    new ContentParser {
+      override def templateSuffixes: Seq[String] = Seq(templateSuffix)
+      override def parse(content:InputStream): Box[NodeSeq] = toInputStreamParser(parseF)(content)
+      override def surround(content:NodeSeq):NodeSeq = surroundF(content)
+    }
+
+  /** A full ContentParser which handles multiple filename suffixes, operates on an InputStream, and surrounds the
+    * top-level templates with the default surround
+    */
+  def full(templateSuffixesSeq:Seq[String],
+           parseF:InputStream => Box[NodeSeq],
+           surroundF:NodeSeq => NodeSeq = defaultSurround):ContentParser =
+    new ContentParser {    
+      override def templateSuffixes: Seq[String] = templateSuffixesSeq
+      override def parse(content:InputStream): Box[NodeSeq] = parseF(content)
+      override def surround(content:NodeSeq):NodeSeq = surroundF(content)
+    }
+}
--- a/web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
@@ -24,59 +24,79 @@ import net.liftweb.util.Helpers
 
 import scala.xml.NodeSeq
 
-/** Objects which can parse content should implement this trait. See LiftRules.contentParsers */
+/**
+ * Objects which can parse content should implement this trait. See [[LiftRules.contentParsers]]
+ */
 trait ContentParser {
-  /** The template filename suffixes this parser is for, e.g. "html", "md", "adoc", etc. */
+  /**
+   * The template filename suffixes this parser is for, e.g. "html", "md", "adoc", etc.
+   */
   def templateSuffixes: Seq[String]
 
-  /** Given an InputStream to a resource with a matching suffix, this function should provide the corresponding
-    * NodeSeq ready to serve
-    * @return
-    */
+  /**
+   * Given an InputStream to a resource with a matching suffix, this function should provide the corresponding
+   * NodeSeq ready to serve
+   * @return
+   */
   def parse(content:InputStream): Box[NodeSeq]
 
-  /** Called to auto-surround the content when needed, i.e. when this content is a top-level template */
+  /**
+   * Called to auto-surround the content when needed, i.e. when this content is a top-level template
+   */
   def surround(content:NodeSeq): NodeSeq
 }
 
 object ContentParser {
-  /** Convenience function to convert a simple String => Box[NodeSeq] content parser function into a
-    * InputStream => Box[NodeSeq] function for satisfying the ContentParser contract.
-    * @param f your String => Box[NodeSeq] content parser
-    * @return your parser wrapped up to handle an InputStream
-    */
-  def toInputStreamParser(f:String => Box[NodeSeq]):InputStream => Box[NodeSeq] = {
+  /**
+   * Convenience function to convert a simple `String => Box[NodeSeq]` content parser function into a
+   * `InputStream => Box[NodeSeq]` function for satisfying the `ContentParser` contract.
+   * @param simpleParser your `String => Box[NodeSeq]` content parser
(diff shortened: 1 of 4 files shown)


Shadowfiend:
Yea verily I have not forgotten ye, and as consolation for last night I will give this some review tonight good sir.

joescii:
Carve it up like Cardele Jones did the Bama D on third downs. 

Joe

web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
```diff
@@ -0,0 +1,82 @@
...
+package http
+
+import java.io.InputStream
+
+import net.liftweb.common.Box
+import net.liftweb.util.Helpers
+
+import scala.xml.NodeSeq
+
+/** Objects which can parse content should implement this trait. See LiftRules.contentParsers */
```
Shadowfiend: Let's make sure to use Scaladoc markup appropriate. `[[LiftRules.contentParsers]]` I think would do the trick here.

web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
```diff
@@ -0,0 +1,82 @@
...
+    */
+  def parse(content:InputStream): Box[NodeSeq]
+
+  /** Called to auto-surround the content when needed, i.e. when this content is a top-level template */
+  def surround(content:NodeSeq): NodeSeq
+}
+
+object ContentParser {
+  /** Convenience function to convert a simple String => Box[NodeSeq] content parser function into a
+    * InputStream => Box[NodeSeq] function for satisfying the ContentParser contract.
```
Shadowfiend: To continue on the scaladoc vein, I've been wrapping types and other code-related stuff in Scaladoc in backticks so they're styled appropriately. See #1650 for more—definitely willing to discuss, mind you. This just makes things easier to read in the generated documentation, IMO, while not interfering too much with reading the docs in code.

web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
```diff
@@ -0,0 +1,82 @@
...
+    * @param f your String => Box[NodeSeq] content parser
+    * @return your parser wrapped up to handle an InputStream
+    */
+  def toInputStreamParser(f:String => Box[NodeSeq]):InputStream => Box[NodeSeq] = {
```
Shadowfiend: I'd suggest renaming `f` to `simpleParser` or something of the sort, personally.

Shadowfiend: Also, style throughout the Lift codebase is a single space after the `:` everywhere. I also think we typically try to list function types as a single type, ie. without spaces:

```scala
  def toInputStreamParser(simpleParser: String=>Box[NodeSeq]): InputStream=>Box[NodeSeq] = {
  ...
```

web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
```diff
@@ -0,0 +1,82 @@
...
+    is:InputStream =>
+      for {
+        bytes <- Helpers.tryo(Helpers.readWholeStream(is))
+        elems <- f(new String(bytes, "UTF-8"))
+      } yield { elems }
```
Shadowfiend: I've seen both styles in the Lift codebase, but I would do this:

```scala
  } yield {
    elems
  }
```

So that a quick scan makes it trivial to spot what's being returned from the `for`. Personal preference here, since there isn't a clear style preference in the Lift code that I've seen.

web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  val defaultSurround:NodeSeq => NodeSeq =
```
Shadowfiend: For now I won't annotate all the places where I've indicated previous deviations from style—I'll let you do a pass through first if that sounds good.

web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
```diff
@@ -0,0 +1,82 @@
...
+    { elems => <lift:surround with="default" at="content">{elems}</lift:surround> }
+
+  /** A basic ContentParser which handles one template filename suffix, operates on a string, and surrounds the
+    * top-level templates with the default surround
+    */
+  def basic(templateSuffix:String,
+            parseF:String => Box[NodeSeq],
+            surroundF:NodeSeq => NodeSeq = defaultSurround):ContentParser =
```
Shadowfiend: Seems like calling these `parsingFunction` and `surroundFunction` would trade off minimal typing for maximal clarity. Not sure what others think. Also, would it make sense to make this an `apply` instead of `.basic`? I feel like the `basic` vs `full` distinction doesn't matter that much to the user.

web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
```diff
@@ -0,0 +1,82 @@
...
+      for {
+        bytes <- Helpers.tryo(Helpers.readWholeStream(is))
+        elems <- f(new String(bytes, "UTF-8"))
+      } yield { elems }
+  }
+
+  val defaultSurround:NodeSeq => NodeSeq =
+    { elems => <lift:surround with="default" at="content">{elems}</lift:surround> }
+
+  /** A basic ContentParser which handles one template filename suffix, operates on a string, and surrounds the
```
Shadowfiend: I've embarked on unifying scaladoc style as I mentioned above, and am moving towards the currently-prevalent style:

```scala
/**
 * Docs…
 */
```

Since we do have Java files in the codebase, it seems like uniformity is an upside. I also have no clue why the Scaladoc folks decided to deviate from Javadoc style in such an insignificant way, but whatevs :p

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -1321,6 +1321,18 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
   val snippets = RulesSeq[SnippetPF]
 
   /**
+   * Handles the parsing of template content into NodeSeqs.
+   */
+  @volatile var contentParsers:Seq[ContentParser] = Seq(
+    new ContentParser {
```
Shadowfiend: Can't we just use `ContentParser.full` here? (Or `.apply` if we switch to that.)

web/webkit/src/main/scala/net/liftweb/http/Templates.scala
```diff
@@ -28,7 +28,10 @@ import java.io.InputStream
  * Contains functions for obtaining templates
  */
 object Templates {
-  private val suffixes = LiftRules.templateSuffixes
+  // Making this lazy to ensure it doesn't accidentally init before Boot completes in case someone touches this class.
+  private lazy val parsers = LiftRules.contentParsers.flatMap(parser =>
```
Shadowfiend: I'd prefer it if we were explicit about this and named it `parserForSuffix`. Though this brings up the question: what happens if two parsers sign up for the same suffix? Right now, we'll just lose the other one, and the order is reversed. This seems somewhat problematic… At minimum, I feel like we should:
 - Be prescriptive about what order they'll be considered in (i.e., in the `LiftRules` docs, say “if multiple things register for the same suffix, the (first|last) one will win”).
 - Think about alternate approaches to first/last one wins. Don't have any suggestions right now, and I don't think we need that solution before merging this stuff, but I think it's worth sticking it in the back of our minds and seeing if we can think of something.

web/webkit/src/main/scala/net/liftweb/http/Templates.scala
```diff
@@ -195,23 +183,21 @@ object Templates {
...
+            val se = parsers.iterator
             val sl = List("_" + locale.toString, "_" + locale.getLanguage, "")
 
             var found = false
             var ret: NodeSeq = null
 
             while (!found && se.hasNext) {
-              val s = se.next
+              val (suffix, parser) = se.next
               val le = sl.iterator
```
Shadowfiend: Oh dear. I'm going to have to take aim with my variable-name-o-smasher one of these days to fix this region and make it comprehensible.

web/webkit/src/main/scala/net/liftweb/http/Templates.scala
```diff
@@ -195,23 +183,21 @@ object Templates {
...
               while (!found && le.hasNext) {
                 val p = le.next
-                val name = pls + p + (if (s.length > 0) "." + s else "")
+                val name = pls + p + (if (suffix.length > 0) "." + suffix else "")
                 import scala.xml.dtd.ValidationException
                 val xmlb = try {
-                  LiftRules.doWithResource(name) { is =>
-                    if (s == "md") {parseMarkdown(is, needAutoSurround)} else
-                    parserFunction(is) } match {
+                  LiftRules.doWithResource(name) { parser.parse } match {
```
Shadowfiend: Can this just be `doWithResource(name)(parser.parse)`? That seems slightly clearer to me.

Shadowfiend:
Overall this looks phenomenal! Really dig the direction this takes us—when all you need from templates is content, who cares what they're written in?

joescii:
Awesome!  I knew the details would be all over the place wrong, but glad to hear the big picture is right.  Details are easier to clean up than pointing in a new direction altogether. :+1: 

I'll see if I can get around to the cleanup effort this afternoon.

web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
fmpwizard: +1 to the extra lines style, same reason as Antonio

web/webkit/src/main/scala/net/liftweb/http/Templates.scala
joescii: Good questions and suggestions.  My gut feel is we should make the last one win, so it's easier to override the default parsers without declaring the entire variable.  What do you think?

web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
joescii: No problem.  Will update accordingly.  My personal style has been to keep it on one line when the yield is a simple expression like this, but I haven't the slightest qualms with following the community style :+1:

web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
joescii: I'm good with the name change.

What I had in mind here is the `basic` is for your typical scenario builder (one suffix and a function of type `String=>Box[NodeSeq]`, whereas the `full` is the fully-specified to the interface of `ContentParser` builder.  The former gives the convenience of not needing a `Seq` for a single element, and eliminates the need to wrap the parser function with `toinputStreamParser`.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
joescii: Yes... in fact I think that was the intention, but I forgot to go back and do it.

web/webkit/src/main/scala/net/liftweb/http/Templates.scala
Shadowfiend: I guess it depends on how we suggest folks add to the list. Right now it's a `Seq`, which means there's not necessarily a prescribed way to add to it. If it were a `List`, the expectation would be that you'd use `::=` to add to it, thus meaning “first one wins” would actually make more sense. I can get behind that approach.

While I'm looking at it again, it's probably worth mentioning in the Scaladocs for `contentParsers` that it's read once when the first template is rendered and then ignored.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -1321,6 +1321,18 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
   val snippets = RulesSeq[SnippetPF]
 
   /**
+   * Handles the parsing of template content into NodeSeqs.
+   */
+  @volatile var contentParsers: Seq[ContentParser] = Seq(
+    ContentParser.full(
+      Seq("html", "xhtml", "htm"),
+      S.htmlProperties.htmlParser,
+      content => content // These templates are not surrounded  by default
```
Shadowfiend: Maybe make this a named parameter and use `identity` instead of `content=>content`?

web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
```diff
@@ -0,0 +1,102 @@
...
+      override def surround(content: NodeSeq): NodeSeq = surroundFunction(content)
+    }
+
+  /**
+   * A full `ContentParser` which handles multiple filename suffixes, operates on an `InputStream`, and surrounds the
+   * top-level templates with the default surround
+   */
+  def full(templateSuffixesSeq: Seq[String],
+           parseF: InputStream=>Box[NodeSeq],
+           surroundF: NodeSeq=>NodeSeq = defaultSurround):ContentParser =
```
Shadowfiend: To elaborate a bit on the `apply` suggestion, I came up with that idea mostly because you can distinguish between `full` and `basic` purely on parameter types. So you'd have:

```scala
  def apply(templateSuffix: String, parseFunction: String=>Box[NodeSeq], surroundFounction: NodeSeq=>NodeSeq = defaultSurround)
  def apply(templateSuffixes: Seq[String], parseFunction: InputStream=>Box[NodeSeq], surroundFounction: NodeSeq=>NodeSeq = 
```

Then you wouldn't have to worry about “am I building a basic or full” thing when invoking, you can just do `ContentParser(suffix, parser, surrounder)` as needed.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
joescii: I always forget about that...

web/webkit/src/main/scala/net/liftweb/http/ContentParser.scala
joescii: Got it! And I like it...

joescii:
Alright y'all, I think that wraps up the suggestions

Shadowfiend:
:+1: This looks pretty good to me; I'll give the other folks a day or two to look it over before merging.

fmpwizard:
Very cool contribution Joe! I had looked at it before and forgot to :+1:

joescii:
Thanks guys! However... I just remembered this morning that I need to address the ordering of the parsers as Antonio had mentioned. 

Joe

Shadowfiend:
Were you going to do something in particular, or just document the current behavior? I can give it a quick pass and implement whatever you wanted to if you don't feel like you have time this week.

joescii:
Hmm... just now noticed your comment.  After re-reading the thread here, I recall that we thought it better to have first-one-wins.  I just need to document it, as well as the fact that it is not re-read later.  I'll get this committed and pushed shortly.

Shadowfiend:
Did you want to move to `List` so that we explicitly encourage `::=` prepending? I think that'll make the first-found behavior more obvious.

joescii:
Sure thing.  Also, I tend to only use `Seq` when I anticipate random access.  One sec...

Shadowfiend:
Ookay, :+1: from me, will try and get it in tomorrow!

Shadowfiend:
Let's do it!

joescii:
Woohoo!  My first lil contribution all grown up now. :)

BTW, I really appreciate the feedback.  I know it's for the love of Lift, but for me it's the first time I've gotten a code review from someone with more Scala experience than myself.  It was eye-opening. :+1:

Shadowfiend:
Glad it was helpful, and feel free push back as needed. Ain't no geniuses here, just collaborative learning and development :)

joescii:
Roll Tide!

Shadowfiend:
lulwut

farmdawgnation:
gif_luka_brozzi350

PR #1654, opened 2015-01-03
fixes #1653

### LINKED ISSUE #1653: LAFuture.collect / LAFuture.collectAll with empty list parameter is never satified
Hi,

One of the usages of collect/collectAll is to pass a sequence using sequence-to-vararg conversion e.g. LAFuture.collect(seqOfFutures: _*). In the corner case the seqOfFutures could be an empty list. In this case collect/collectAll is never satisfied, because f.foreach will never execute given function.

I've prepared PR reproducing problem and fixing it: #1654.

ML discussion @ https://groups.google.com/forum/#!topic/liftweb/rsscoZ2BEYM

arkadius:


Shadowfiend:
Closing pending an ML thread.

Shadowfiend:
Reopening since the discussion seems to have been positive. Can we do a commit with a better message here? Then I think we put this on through.

arkadius:
I've mistakenly submited here pull request with my fork's master branch. So I'm closing it and reopen PR with only one commit with better message: #1664. Thanks for acceptance!

PR #1657, opened 2015-01-04
fixes #1656 LiftJavaScriptSpec fail on non-english environment

### LINKED ISSUE #1656: LiftJavaScriptSpec fail on non-english environment
`LiftJavaScriptSpec` have included expected message in english:
```javascript
"ajaxOnFailure": function() {alert("The server cannot be contacted at this time");}
```
, but `LiftJavaScript.settings` implementation generates localized message. It causes that tests fail.
My suggestion is to put localized message in expected message. 

Link to mailing list discussion:
https://groups.google.com/forum/#!topic/liftweb/TnfzvTKkdDQ

Pull request: #1657

arkadius:


Shadowfiend:
In this and the other PRs, can you include in the commit message what it's doing/why/etc? e.g., in this case something like:

```
Fix failing LiftJavaScriptSpec in non-English environments.

Generated Lift JS is localized based on the current environment. However,
LiftJavaScriptSpec checks for English messages, so running it under a non-English
environment causes it to fail due to the mismatch in expected message.
```

Additionally, I'm thinking the correct fix here is to force the locale to `en_US`, rather
than invoke all the i18n machinery in the spec.

Other opinions?

Shadowfiend:
(Oh one more thing, but this is personal preference which means you should do whatever feels best to you—I personally don't include issue numbers in my commit messages. The commit will be associated with the issue via PR, and the functionality that it implements and justification for implementing it should be summarized in the commit message itself, so external sources don't need to be consulted to figure out why something was done. If you do include the issue number, please include it somewhere other than the commit summary. Here's [some](http://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message) [more](http://chris.beams.io/posts/git-commit/) [reading](https://github.com/erlang/otp/wiki/Writing-good-commit-messages) on good commit messages.)

arkadius:
I've changed the message. It make history a bit uggly but after merge, blame should give better description of change.

According to test behaviour, I think that changing environment inside test is fixing implementation to work on only one environment. If we adopt this strategy in all tests we may release software usable in just one country :)

Shadowfiend:
That's a good point, but I think this isn't quite the right solution. There are two things we
need to check, and they're independent of each other:
 - Do we generate the expected JS object.
 - Do we internationalize that object.

This fix conflates them into the same test, which means a failure doesn't immediately
tell us what failed. So to address both issues, I'd say it would make sense to force the
locale to English to check that the JS object is generated as expected, and then force
the locale to a known second value to check that the result is localized. This decouples
the two verifications, which I think is a net win.

One of my rules of thumb with specs is to make sure the expected values are constants,
not generated through something else, and that's where this concern stems from.

Thoughts?

arkadius:
I will aggree with you with almost 100%. The only think I disagree is your last mentioned rule. I think that random tests are the best for some situations e.g. some computations or when we have imlemented some round-trips and we can use it for random tests e.g. export/import.

I will separete internationalization check and for all other cases use english locales.

arkadius:
You have 100% right. I will separate internalization test and for other cases will use english locales.

arkadius:
Done it.

Shadowfiend:
Dig it! :+1: from me, will give it a day so other committers get a chance to raise a flag if necessary.

Shadowfiend:
Whoops, got delayed on this guy. Let's do it!

Shadowfiend:
Going to do this guy on a new branch so we can get a cleaner history; please wait ;)

Shadowfiend:
Closed by merging 65fc960dee76ed3ba46363d05ed3a39c7cf5add4.

PR #1659, opened 2015-01-04
fixes #1658 (CombParserHelpersSpec fails when string with \n is probed)

### LINKED ISSUE #1658: CombParserHelpersSpec fails when string with \n is probed
"... provide an aNumber parser - returning an Int if succeeding" fails because "1\n" matches criteria aNumber but doesn't match regex expression "\\p{Nd}+.*". Because without DOTALL "\n" doesn't match ".*"

Link to mailing list discussion: https://groups.google.com/forum/#!topic/liftweb/hHzK9lgXyKs

Pull Request: #1659

arkadius:
fixes #1658

### DIFF
--- a/core/util/src/test/scala/net/liftweb/util/CombParserHelpersSpec.scala
+++ b/core/util/src/test/scala/net/liftweb/util/CombParserHelpersSpec.scala
@@ -75,7 +75,7 @@ object CombParserHelpersSpec extends Specification with ScalaCheck {
     "provide a digit parser - returning a String" in {
       val isDigit: String => Boolean =
         (s: String) => digit(s) match {
-          case Success(x, y) => s must beMatching ("\\p{Nd}.*")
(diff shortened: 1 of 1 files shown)


Shadowfiend:
As I mentioned in the other PR, a better commit message would be ideal here. Ideally it would include the information that's listed in the issue (a short summary that indicates what changed, e.g. “Fix CombParserHelperSpec for digits with newlines.”, then a description indicating why it was failing and what you did to fix it).

arkadius:
Done it.

Shadowfiend:
Lovely! I'll merge this in later today after rebasing a bit so we only include the good commit :)

Shadowfiend:
Okay, merged this via https://github.com/lift/framework/commit/f12a1d905b6aa79cfe7de6d46648c56861d667ba after making sure we only pulled the clean commit. I'll let #1664 bring in the updated contributors.md file.

PR #1660, opened 2015-01-05
TimeSpans with same milliseconds value are not equal fix

arkadius:


arkadius:
Closing it because it does not resolve the root of a problem according to mailing list discussion.

PR #1663, opened 2015-01-05
Make HLists covariant in their type parameters.

Shadowfiend:
Also add a spec for HList length.

I'm using this code in lift-formality and it works as advertised, but it's nicer if
I don't keep my own copy of it ;)

### DIFF
--- a/core/common/src/main/scala/net/liftweb/common/HList.scala
+++ b/core/common/src/main/scala/net/liftweb/common/HList.scala
@@ -69,19 +69,11 @@ object HLists {
    * }
    * }}}
    */
-  sealed trait HList {
-    type Head
-    type Tail <: HList
-
-    /**
-     * The length of the HList
-     */
-    def length: Int
-  }
+  sealed trait HList
 
   /**
    * The last element of an `HList`. This is the starting point for an `HList`,
-   * and carries a `:+:` method to start one:
+   * and you can use `[[HListMethods.:+: :+:]]` to start one based on it:
    *
    * {{{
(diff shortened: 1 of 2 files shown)


core/common/src/main/scala/net/liftweb/common/HList.scala
```diff
@@ -116,7 +96,7 @@ object HLists {
    * Carries the information about the type of this element, plus the `HList`
    * type of the rest of the list.
    *
-   * You can use `:+:` to make this `HList` longer:
+   * You can use `[[HListMethods.:+: :+:]]` to make this `HList` longer:
    *
    * {{{
```
fmpwizard: would you want to update the scaladoc style while you are here or are you planning on doing another massive pass at docs on a separate PR?

fmpwizard:
:+1:

core/common/src/main/scala/net/liftweb/common/HList.scala
Shadowfiend: I actually waited until the other docs PR for `lift-common` went in, so these docs should all be up-to-date, I think. Did you see an issue? I kind of had to move the tweaks I'd made for lift-formality into the already-properly-doced file, so it's possible I missed something.

Shadowfiend:
I'm not sure how much HLists are being used in general, but given one thumbs up I'll probably merge this tomorrow unless someone else protests vehemently :p

Shadowfiend:
I'm not sure when “tomorrow” turned into “in 6 days”, but in we go :p

PR #1665, opened 2015-01-07
Extract-o-matic: Extract event handlers to page JS (take 2)

Shadowfiend:
The main functionality here moves event handler attachment from inline
attributes to the page JavaScript that Lift 3 now supports. The reason for
this is so that, out of the box, Lift will be compatible with very restrictive
Content Security Policy settings when using built-in Lift features.

To do this, we add a lift.onEvent function that has both jQuery and vanilla
implementations. There are also a couple of fixes to existing JS functionality
here.

This fella's ready for a look, though I'm thinking we may want to do an M3
release before this goes in. This'll change behavior in a few ways and it'll be
good to have a stable fallback in case the changed behavior breaks things.

Also, this PR consists of a single commit because I accidentally merged
#1608  while cherry-picking another contribution into a branch that I merged
to master 😞 A cone of shame will ensue.

### DIFF
--- a/web/webkit/src/main/resources/toserve/lift.js
+++ b/web/webkit/src/main/resources/toserve/lift.js
@@ -643,8 +643,9 @@
 
   window.liftJQuery = {
     onEvent: function(elementOrId, eventName, fn) {
-      if (typeof elementOrId == 'string')
+      if (typeof elementOrId == 'string') {
         elementOrId = '#' + elementOrId;
+      }
 
       jQuery(elementOrId).on(eventName, fn);
     },
--- a/web/webkit/src/main/resources/toserve/lift.js
+++ b/web/webkit/src/main/resources/toserve/lift.js
@@ -65,8 +65,11 @@
       ajaxGet: function() {
         consoleOrAlert("ajaxGet function must be defined in settings");
       },
+      onEvent: function(elementOrId, eventName, fn) {
+        consoleOrAlert("onEvent function must be defined in settings");
+      },
       onDocumentReady: function(fn) {
-        consoleOrAlert("documentReady function must be defined in settings");
+        consoleOrAlert("onDocumentReady function must be defined in settings");
       },
       cometGetTimeout: 140000,
       cometFailureRetryTimeout: 10000,
@@ -554,7 +557,7 @@
         this.extend(settings, options);
 
         var lift = this;
-        options.onDocumentReady(function() {
+        settings.onDocumentReady(function() {
           var attributes = document.body.attributes,
               cometGuid, cometVersion,
               comets = {};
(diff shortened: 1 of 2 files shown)


web/webkit/src/main/resources/toserve/lift.js
```diff
@@ -584,7 +587,8 @@
           doCycleIn200();
         });
       },
-      logError: settings.logError,
+      logError: function() { settings.logError.apply(this, arguments) },
```
eltimn: I'm curious why this was necessary. Do the arguments not get passed along properly without it? Or was there some other reason?

web/webkit/src/main/resources/toserve/lift.js
```diff
@@ -638,6 +642,12 @@
   })();
 
   window.liftJQuery = {
+    onEvent: function(elementOrId, eventName, fn) {
+      if (typeof elementOrId == 'string')
+        elementOrId = '#' + elementOrId;
+
```
eltimn: I don't know about you but not using braces in JavaScript makes me nervous.

I also know it's easy to overlook when switching from Scala.

web/webkit/src/main/resources/toserve/lift.js
Shadowfiend: The commit message I wrote was:

```
We were tying settings.logError directly to lift.logError, but
this didn’t allow for the options passed to init to override the
implementation because lift.logError is bound before the init
function runs. We now delegate to settings.logError using
apply, ensuring we’re always seeing the latest version of it.
```

So I think the issue is that without that, sending in new settings with an updated `logError` won't work.

web/webkit/src/main/resources/toserve/lift.js
Shadowfiend: Haha it doesn't make me super-nervous but I can totally understand the nervousness; I'll fix it :)

web/webkit/src/main/resources/toserve/lift.js
eltimn: Got it. Thanks.

web/webkit/src/main/resources/toserve/lift.js
Shadowfiend: Fixed that for you :)

eltimn: Thanks. I guess more than anything it just doesn't look right to me without braces as I tend to use them as a rule in JavaScript.

eltimn:
:+1:

web/webkit/src/main/resources/toserve/lift.js
Shadowfiend: Yeah, completely understand—I use them 99% of the time as well. I pulled in some code for event handling and did some rearrangement here, so it must have just slipped through :)

Shadowfiend:
Kk, now that M3 is out, will give another day or two for folks to raise any concerns/remarks and then merge this guy.

Shadowfiend:
Aaaand movin' it.

PR #1662, opened 2015-01-05
travis integration

arkadius:
- build script
- build status in README

Jus sign in with github account, turn on project, additional turn on "Build only if .travis.yml is present" and send some triggering commit. You can also reconfigure notifications in .travis.yml if you prefer according to: http://docs.travis-ci.com/user/notifications/

discusiion link: https://groups.google.com/forum/#!topic/liftweb/hHzK9lgXyKs

Issue #1661, opened 2015-01-05
TimeSpans with same milliseconds value are not equal


arkadius:
TimeSpan created with the same milliseconds value but in two different ways (TimeSpan.apply and using TimeSpanBuilder) could be not equal for some time zones. This failes TimeSpanSpec for EST time zone.

Link to mailing list discussion: https://groups.google.com/forum/#!topic/liftweb/CgZFVwMEiqU

PR: #1660

PR #1603, opened 2014-08-03
Your data-lift is my data-lift: Reimplement data-lift as a data attribute parser.

farmdawgnation:
This PR reimplements data-lift as a data attribute parser. This stemmed from a Lift hack day a few months ago, and I'm just now getting around to wrapping this up...

Few points of concern:

* I want to go through at some point and clean up the copy-pasting of helper methods that I did wrt MetaData parsing.
* I'm not entirely sure why I'm having to manually check for the parallel attribute in my parser. My understanding was that if my parser generates a `lift:` node, that should be processed subsequently by the regular SnippetNode parser, which does checks for parallel processing.
* Can someone make sure my implementation of handling parallel is even functional while we're at it? lol

All specs are passing, though. So that's something.

closes #1582

PR #1666, opened 2015-01-08
Travis Integration

Shadowfiend:
Based off of #1662, but in one of our branches so I can mess with it ;)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+services:
+  - mongodb
+
+jdk:
+  - oraclejdk7
+
+script: ./liftsh test
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,6 @@ services:
 
 jdk:
   - oraclejdk7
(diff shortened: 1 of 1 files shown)


Shadowfiend:
Now that I look at it it looks like Travis will build the branch without a PR, so that's pretty handily-dandily handy-dandy…

arkadius:
Maybe it is a reason why: "Travis CI builds a pull request when it's first opened and when commits are added to the pull request throughout its lifetime.". I see that #1666 and #1603 is checked agains build. Because of #1661 they are failed. However discussion about TimeSpan problem fix doesn't involve too many persons on mailing list.

Shadowfiend:
Yeah, we'll see. I'm going to follow up on the TimeSpan discussion today if I get a chance.

Shadowfiend:
Okay okay okay, these now seem to be passing; cleaned up history a bit, will merge it in once travis gives the green light.

PR #1667, opened 2015-01-08
replaced TimeSpan with scala.concurrent.duration.Duration

arkadius:
Mailing list discussion: https://groups.google.com/forum/#!topic/liftweb/CgZFVwMEiqU

`TimeSpan` has ambiguous definition. It can behave like a joda `Period` (e.g. take care about timeshift)
but for most cases it should by only a wrapper for in-milliseconds duration. It breaks SRP. This change replace `TimeSpan` with `scala.concurrent.duration.Duration` deprecating the first one.

For most cases old code should work without changes - only import of `TimeSpanHelpers` will be neede and deprecation notice will appear. Change should be painless If user will want to clear it because of similar DSL for `Duration` building as in `TimeSpan`. The all what will be needed is to replace import with  `scala.concurrent.duration._.`

For a few places code will not compile until user make a migration because `TimeSpan` was used in abstract method e.g. `CometActor.lifespan: Box[TimeSpan]`

There is a little walk-around in `Schedule.schedule(f: () => Any, delay: Duration): ScheduledFuture[Unit]`. It takes now `f: () => Any` istead of `f: () => Unit`. Because of backward compatibility there is overloaded method with TimeSpan and it causes ambiguity if smb pass a function returning other thing than `Unit`.

### DIFF
--- a/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
+++ b/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
@@ -18,6 +18,7 @@ package net.liftweb
 package actor 
 
 import common._
+import scala.concurrent.duration._
 
 
 /**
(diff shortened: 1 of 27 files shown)


arkadius:
Closing because of idea of smooth transition with TimeSpan methods deprecations before.

PR #1668, opened 2015-01-13
TimeSpan represents duration in milliseconds.

arkadius:
Mailing list discussion: https://groups.google.com/forum/#!topic/liftweb/CgZFVwMEiqU

Before the change TimeSpan has many responsibilities - in equality behave as a joda Period (was checking duration field equality), can represent a Date (using toDate/toDateTime operations) or be used as a simple container for milliseconds. Now is responsible only for operations on milliseconds. Problematic methods have been deprecated (.ago/.later have been moved to implicit PeriodExtension). For backward compatibility there is introduced implicit conversion from Period to TimeSpan.

Period inside TimeSpan is holded now only for puropose of to-period conversion (for months/years builder). There has been used view bounds for purpose of nested conversions usage:
1.second.before
1 -> TimeSpan -> Period

This is WIP branch. Please send me feedback if my english has failed in deprecation descriptions :-)

### DIFF
--- a/core/util/src/main/scala/net/liftweb/util/TimeHelpers.scala
+++ b/core/util/src/main/scala/net/liftweb/util/TimeHelpers.scala
@@ -32,8 +32,8 @@ import common._
 object TimeHelpers extends TimeHelpers with ControlHelpers with ClassHelpers
 
 /**
- * The TimeHelpers trait provide functions to create TimeSpans (an object representing an amount of time), to manage date formats
- * or general utility functions (get the date for today, get year/month/day number,...)
+ * The TimeHelpers trait provide functions to create TimeSpans (an object representing duration in milliseconds),
+ * to manage date formats or general utility functions (get the date for today, get year/month/day number,...)
  */
 trait TimeHelpers { self: ControlHelpers =>
   // Logger must be lazy, since we cannot instantiate until after boot is complete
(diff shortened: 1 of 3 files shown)


core/util/src/main/scala/net/liftweb/util/TimeHelpers.scala
```diff
@@ -54,96 +51,86 @@ trait TimeHelpers { self: ControlHelpers =>
...
 
+    /**
+     * @return amount of milliseconds in duration
+     * @throws UnsupportedOperationException if was created by deprecated months/years builder
+     */
     def toMillis = millis
 
+    /**
+     * @return amount of milliseconds in duration
+     * @throws UnsupportedOperationException if was created by deprecated months/years builder
```
Shadowfiend: This is documenting existing behavior rather than a new exception, right?

core/util/src/main/scala/net/liftweb/util/TimeHelpers.scala
```diff
@@ -427,18 +424,29 @@ trait TimeHelpers { self: ControlHelpers =>
...
   def millis: Long
 }
 
+@deprecated("TimeSpan to Date/DateTime conversion will be removed for possibility of mistakes in on-duration operations")
 object ConvertableToDate {
+  @deprecated("TimeSpan to Date conversion will be removed for possibility of mistakes in on-duration operations")
   implicit def toDate(in: ConvertableToDate): Date = in.toDate
+  @deprecated("TimeSpan to DateTime conversion will be removed for possibility of mistakes in on-duration operations")
   implicit def toDateTime(in: ConvertableToDate): DateTime = in.toDateTime
   implicit def toMillis(in: ConvertableToDate): Long = in.millis
```
Shadowfiend: I'm thinking we deprecate this as well. Even though it doesn't convert `TimeSpan` to `Date`, it does facilitate what could be a pretty hard-to-trace usage.

core/util/src/main/scala/net/liftweb/util/TimeHelpers.scala
```diff
@@ -54,96 +51,86 @@ trait TimeHelpers { self: ControlHelpers =>
...
+    def hours = new TimeSpan(Left(Duration.standardHours(len)))
     def hour = hours
-    def days = new TimeSpan(Right(Duration.standardDays(len): Period))
+    def days = new TimeSpan(Left(Duration.standardDays(len)))
     def day = days
-    def weeks = new TimeSpan(Right(Duration.standardDays(len * 7L): Period))
+    def weeks = new TimeSpan(Left(Duration.standardDays(len * 7L)))
     def week = weeks
-    def months = new TimeSpan(Right((new Period().plusMonths(len.toInt))))
+    @deprecated("TimeSpan will not support operations on non-millis periods in future")
```
fmpwizard: I believe `@deprecated`  takes two parameters, one is the text you have here and the other is the version, so `3.0`

Shadowfiend: Let's use 3.0.0. I think it's worth going full semver for the 3 series, because why not :)

core/util/src/main/scala/net/liftweb/util/TimeHelpers.scala
arkadius: Yest, It's documenting existing behavior

core/util/src/main/scala/net/liftweb/util/TimeHelpers.scala
arkadius: I agree with it. The only one type conversion that exists now in TimeHelpers is joda Duration -> TimeSpan conversion. What do you think about deprecation of it?
Other implicit conversions are for class extensions.

core/util/src/main/scala/net/liftweb/util/TimeHelpers.scala
arkadius: done it

arkadius:
Tell me when you are filling that all looks fine. Then I will prepare branch with squashed changes.

Shadowfiend:
Sorry, I haven't had a chance to give this a second, deeper look. I'll try and do that tomorrow so we can roll out the M3 build! For time's sake, if everything looks reasonably good I'll probably make any tweaks to history and verbiage and merge it in at that time.

Shadowfiend:
Okay, I want most of this to land in M3, so I undid one thing I want to discuss a little more–namely the deprecations on `ConvertableToDate`. I'll bring it up on the list and we can figure that out and get any related changes in for M4.

I also reworded some of the deprecation notices and listed alternatives for all of them, and added a block about the deprecation to the `TimeSpan` scaladocs.

I put the result in #1670; let's try to merge that, close this, and resolve the remaining couple of questions on the list!

PR #1670, opened 2015-01-18
TimeSpan Deprecation

Shadowfiend:
Final version of #1668 .

Dropped deprecation on `ConvertableToDate` pending further discussion.

The current state is also a bit ~ because `5.days.ago` still throws a deprecation
warning… Need to discuss that a bit further.

Final version for M3 though.

Closes #1668.


fmpwizard:
:+1:

PR #1672, opened 2015-01-21
Fixed NPE in LiftRules:913

### LINKED ISSUE #1671: NullPointerException in LiftRules:913 when no "parallel"-attribute exists
In master, version: a21dfa04cef89ab7380c77a09cce53f18865c5d0

Line 913 in LiftRules looks like this
```
912:          val decodedMetaData = pairsToMetaData(encodedArguments.flatMap(_.roboSplit("[;&]")))
913:          if (decodedMetaData("parallel").headOption == Some(Text("true"))) {
```
The definition of MetaData.apply is:

```
  /**
   * Gets value of unqualified (unprefixed) attribute with given key, null if not found
   *
   * @param  key
   * @return value as Seq[Node] if key is found, null otherwise
   */
  def apply(key: String): Seq[Node]
```
Which makes line 913 throw NPE when the attribute="parallel" is not present.

Should be re-written to:
decodedMetaData.get("parallel").headOption

andreak:
Fixed NPE in LiftRules:913

fixed #1671


fmpwizard:
:+1:

PR #1673, opened 2015-01-23
Beautiful and Unique Snowflake: Try to ensure uniqueness of comet requests

Shadowfiend:
> ["You are not a beautiful or unique snowflake. You are the same decaying organic matter as everyone else, and we are all a part of the same compost pile."](http://en.wikiquote.org/wiki/Fight_Club_(novel))

(^ for the positive thought of the day :p)

We already did some work to make sure comet requests were unique;
this adds one more layer of checking around that. When we schedule a
new comet request, we include the current "comet request count". That
count is incremented when the request is actually fired off. This means
that scheduling a new request multiple times before the request is fired
will only result in one request, rather than multiple requests that will walk
all over each other.

Should fix issues from the ML threads at https://groups.google.com/forum/#!topic/liftweb/cREIVSEpDzQ
and https://groups.google.com/forum/#!topic/liftweb/vTUL2RW0SVc,
hopefully.

### DIFF
--- a/web/webkit/src/main/resources/toserve/lift.js
+++ b/web/webkit/src/main/resources/toserve/lift.js
@@ -293,7 +293,9 @@
     ///// Comet ////////////////////////////////////
     ////////////////////////////////////////////////
 
-    var currentCometRequest = null;
+    var currentCometRequest = null,
+        // Used to ensure that we can only fire one comet request at a time.
+        cometRequestCount = 0;
 
     // http://stackoverflow.com/questions/4994201/is-object-empty
     function is_empty(obj) {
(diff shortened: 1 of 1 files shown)


andreak:
+1

Shadowfiend:
Fixed a couple of issues codacy caught; going to let codacy and travis have their pass at it before merging. This seems to fix the reported issues as far as I can tell.

Shadowfiend:
Hmm… No clear indication of codacy check status or what have you, so I'm going to go ahead and put this in; asked them a question via their support stuff to see if there is any status information anywhere that I missed for future reference.

andreak:
Nice!

PR #1674, opened 2015-01-26
Locket Down: Add support for Content-Security-Policy and Strict-Transport-Security

Shadowfiend:
`SecurityRules` provides a way to set security rules like HTTPS
requirements and a content security policy, which are in turn
served with resources from Lift via headers. Right now, we support
`Content-Security-Policy` and `Strict-Transport-Security` headers.

Content security policy violations are by default reported to a URI
under the lift path, parsed into a case class, and passed to a `LiftRules`
function for handling. The default handler simply logs the violation and
returns a 200. If we get a violation report that doesn’t follow the expected
JSON format, we log an issue and return a 400 bad request.

Along the way, also renamed `BadResponse` to `BadRequest`, since
the previous name was wrong, and deprecated `BadResponse`, as
well as adding the ability to pass `BadRequest` a custom message.

Looking for some feedback here; the default `Content-Security-Policy`
isn’t quite loose enough for default Lift behavior. In particular, we’ll
probably need to allow `unsafe-eval` by default, due to the fact that Lift’s
JS responses need to be run through `eval` on the client.

This is, however, the initial API I have in mind. Let me know what you
think!

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftResponse.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftResponse.scala
@@ -107,13 +107,13 @@ case class ResetContentResponse() extends LiftResponse with HeaderDefaults {
  * the request appears incorrect. Use the `message` to indicate what was wrong
  * with the request, if that does not leak important information.
  */
-case class BadRequest(message: String = "") extends LiftResponse with HeaderDefaults {
+case class BadRequestResponse(message: String = "") extends LiftResponse with HeaderDefaults {
   def toResponse = InMemoryResponse(message.getBytes("UTF-8"), headers, cookies, 400)
 }
 object BadResponse {
-  @deprecated("Use BadRequest instead, as that is the correct name for this response.", "3.0.0")
+  @deprecated("Use BadRequestResponse instead, as that is the correct name for this response.", "3.0.0")
   def apply() = {
-    BadRequest()
+    BadRequestResponse()
   }
 }
 
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftResponse.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftResponse.scala
@@ -104,10 +104,17 @@ case class ResetContentResponse() extends LiftResponse with HeaderDefaults {
  * 400 Bad Request
  *
  * Your Request was missing an important element. Use this as a last resort if
- * the request appears incorrect.
+ * the request appears incorrect. Use the `message` to indicate what was wrong
+ * with the request, if that does not leak important information.
  */
-case class BadResponse() extends LiftResponse with HeaderDefaults {
-  def toResponse = InMemoryResponse(Array(), headers, cookies, 400)
+case class BadRequest(message: String = "") extends LiftResponse with HeaderDefaults {
+  def toResponse = InMemoryResponse(message.getBytes("UTF-8"), headers, cookies, 400)
+}
+object BadResponse {
+  @deprecated("Use BadRequest instead, as that is the correct name for this response.", "3.0.0")
+  def apply() = {
+    BadRequest()
+  }
 }
 
 /**
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
@@ -1654,7 +1654,6 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
    */
   val supplementalHeaders: FactoryMaker[List[(String, String)]] = new FactoryMaker(() => {
     ("X-Lift-Version", liftVersion) ::
-    ("X-Frame-Options", "SAMEORIGIN") ::
     securityRules().headers
   }) {}
 
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
@@ -264,7 +264,7 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
    */
   val beforeSend = RulesSeq[(BasicResponse, HTTPResponse, List[(String, String)], Box[Req]) => Any]
 
-  private lazy val defaultSecurityRules = SecurityRules()
+  private[this] lazy val defaultSecurityRules = SecurityRules()
   /**
    * The security rules used by Lift to secure this application. These mostly
    * relate to HTTPS handling and HTTP `Content-Security-Policy`. See the
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
@@ -264,6 +264,14 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
    */
   val beforeSend = RulesSeq[(BasicResponse, HTTPResponse, List[(String, String)], Box[Req]) => Any]
 
+  private lazy val defaultSecurityRules = SecurityRules()
+  /**
+   * The security rules used by Lift to secure this application. These mostly
+   * relate to HTTPS handling and HTTP `Content-Security-Policy`. See the
+   * `[[SecurityRules]]` documentation for more.
+   */
+  @volatile var securityRules: () => SecurityRules = () => defaultSecurityRules
+
   /**
    * Defines the resources that are protected by authentication and authorization. If this function
    * is not defined for the input data, the resource is considered unprotected ergo no authentication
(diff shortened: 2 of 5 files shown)


web/webkit/src/main/scala/net/liftweb/http/SecurityRules.scala
```diff
@@ -0,0 +1,423 @@
...
+    List(
+      "Content-Security-Policy" -> contentSecurityPolicyString,
+      "X-Content-Security-Policy" -> contentSecurityPolicyString
+    )
+  }
+  /**
+   * Returns the headers implied by this content security policy.
+   */
+  def headers(enforceInDevMode: Boolean = true): List[(String, String)] = {
+    if (! enforceInDevMode && Props.devMode) {
```
d6y: Rather than `enforceInDevMode` and `logInDevMode` would it make sense to allow the user to supply some function from `List[String,String] => List[String,String]` and let them do anything they want in that function? Log them, return `Nil`; filter them or generally make some kind of change and return some other List? 

I don't have a particular use case in mind, so consider this a random idea rather than some kind of must have thing.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -1634,9 +1644,39 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
...
+   * that's sent.
+   *
+   * Note that the headers for the applications `SecurityRules` are also set
+   * here, so if you override the supplemental headers, you should
+   * either refer back to the default set or make sure to include
+   * `LiftRules.securityRules.headers`.
+   */
+  val supplementalHeaders: FactoryMaker[List[(String, String)]] = new FactoryMaker(() => {
+    ("X-Lift-Version", liftVersion) ::
+    ("X-Frame-Options", "SAMEORIGIN") ::
```
d6y: Should we think of `X-Frame-Options` as a kind of security rule? A kind of content source restriction? Or not really a good fit there?

tjweir:
"Along the way, also renamed BadResponse to BadRequest, since
the previous name was wrong, and deprecated BadResponse, as
well as adding the ability to pass BadRequest a custom message.
"

Awesome!

web/webkit/src/main/scala/net/liftweb/http/SecurityRules.scala
Shadowfiend: You mean to change the output headers, rather than provide a first-class flag that tweaks the behavior?

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
Shadowfiend: It wouldn't be a content source restriction (`frameSources` is, but it's in a sense the mirror of `X-Frame-Options`), but moving it into security rules sounds like a great idea; I will do that.

Shadowfiend:
My crusade against misnomers continues ;)

web/webkit/src/main/scala/net/liftweb/http/SecurityRules.scala
d6y: I was thinking: function rather than flags. Then I can enable or disable them in any mode I like, log them or not log them, only log some of them....I don't know.  I suppose I'm trying to second guess what people might want to do with headers on a request-by-request basis. Like I said, random idea.

Shadowfiend: Yeah, I get what you mean—but I think they can do that kind of customization with `supplementalHeaders` directly, since that's where these headers go. I wanted this flag specifically basically because I wanted the default dev behavior to be report-only, and the default anything-else behavior to be enforce. So I figured it deserved its own first-class flag. Willing to change it if we think there's a better way though.

d6y: I'm happy if the user can change everything via `supplementalHeaders` then I'm cool with that.  Thanks for listening :-)

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
Shadowfiend: To clarify, Content-Security-Policy 2 does feature a `frame-ancestors` flag for this… But I'm aiming at CSP 1 support here for the time being.

Shadowfiend:
Three changes:
 - `X-Frame-Options` is now specified via a security rule, as Richard suggested.
 - Fixed the inheritance hierarchy for source restrictions so you can properly use `All` and friends for JS and stylesheet sources.
 - Changed script sources to include `UnsafeEval`. Most of Lift's JS stuff, including comet, relies on jQuery's `globalEval`, which in turn relies on `eval`. This is [will no longer be the case](https://github.com/jquery/jquery/pull/1449) in the next version of jquery, so I can back this change out when that comes out if we want to require it for Lift 3. May be worth asking on the ML? I also need to double-check that script injection is possible without `unsafe-inline`… Not sure if that's the case. If it is, maybe we can roll our own globalEval based on script injection (it's a fairly simple function).

Shadowfiend:
Also worth mentioning: we'll have to either remove the `ajax-loader.gif` in the default template, or we'll have to change its inline styles to something in a stylesheet.

Shadowfiend:
Ok, so script injection is understandably blocked if `unsafe-inline` is not included… I think `unsafe-eval` is the lesser of the two evils for now, but to really allow proper usage of `Content-Security-Policy`, we'll want to offer a more-limited channel that is made up entirely of JSON exchange—something that should be perfectly easy to do.

Shadowfiend:
(Er, but outside the scope of this PR—for now I'll leave `unsafe-eval` as a default, and once we introduce JSON-based messaging we can turn it off in `ContentSecurityPolicy.secure`'s defaults, along with a note as to what that could cause.

web/webkit/src/main/scala/net/liftweb/http/LiftResponse.scala
```diff
@@ -104,10 +104,17 @@ case class ResetContentResponse() extends LiftResponse with HeaderDefaults {
  * 400 Bad Request
  *
  * Your Request was missing an important element. Use this as a last resort if
- * the request appears incorrect.
+ * the request appears incorrect. Use the `message` to indicate what was wrong
+ * with the request, if that does not leak important information.
  */
-case class BadResponse() extends LiftResponse with HeaderDefaults {
-  def toResponse = InMemoryResponse(Array(), headers, cookies, 400)
+case class BadRequest(message: String = "") extends LiftResponse with HeaderDefaults {
```
farmdawgnation: Shouldn't this really be `BadRequestResponse`? (cf. `OkResponse`, `RedirectResponse`, `JsonResponse`, `CreatedResponse`)

web/webkit/src/main/scala/net/liftweb/http/SecurityRules.scala
```diff
@@ -0,0 +1,461 @@
...
+ * @param logInDevMode If true, dev mode violations of security policies are
+ *        logged by default. Note that if you override
+ *        `[[LiftRules.contentSecurityPolicyViolationReport]]` or otherwise
+ *        change the default Lift policy violation handling behavior, it will
+ *        be up to you to handle this property as desired.
+ */
+final case class SecurityRules(
+  https: Option[HttpsRules] = None,
+  content: Option[ContentSecurityPolicy] = Some(ContentSecurityPolicy()),
+  frameRestrictions: Option[FrameRestrictions] = Some(FrameRestrictions.SameOrigin),
```
farmdawgnation: These all reduce to headers, don't they? Do the disparate types gain us anything, or could we define a common trait that will have a method headers?

web/webkit/src/main/scala/net/liftweb/http/LiftResponse.scala
Shadowfiend: A perfectly reasonable point, will make that change.

web/webkit/src/main/scala/net/liftweb/http/SecurityRules.scala
Shadowfiend: Yes, they give us proper naming and enumeration of the available kinds of security rules. Also, their purpose is not solely providing headers; for example, we currently check early in request servicing if https rules are enabled but set to log-only for dev mode, and log a warning when we get a non-https request. This log-only behavior is not supported by the HTTP header.

farmdawgnation: Rockin'

Shadowfiend:
All right, what are we thinking here? Ready for merge, or are we worried about issues?

Shadowfiend:
Whoops, except for the point about `BadRequestResponse`, which I forgot about. Will make that change here shortly ;)

fmpwizard:
After updating to `BadRequestResponse`, :+1:

Shadowfiend:
Hopefully travis confirms we're in bidness :)

farmdawgnation:
What Diego said.

fmpwizard:
This is why I didn't like the idea of adding the "build <status>" to the README, false fails would make it look like there is a problem :(

Shadowfiend:
Mmkay, I'm going to ignore the Codacy warnings about volatile vars if we agree that's the right move.

As for the Travis build, I'm going to look into a suggested fix at https://github.com/travis-ci/travis-ci/issues/1441 for keeping a local cache in travis.

fmpwizard:
:+1:

Shadowfiend:
Heh. I may have confused the travis builder with a force-push :innocent:

Shadowfiend:
Okay, I think we're in business here. Anyone interested in merging? :)

fmpwizard:
let's do this!

Shadowfiend:
! Now the excitement starts :p

PR #1678, opened 2015-01-31
LAFuture.map/.flatMap returns Future satisfied with Failure when Exception in transforming function occurs

arkadius:
Before the change LAFuture.map/.flatMap in this case return never satisfied Future. After the change, returned Future is satisfied by Failure.

This change also contains minor fix: get(timeout) was returning Empty instead of Failure if LAFuture was aborted.

Also added extensions providing `toBox` conversions from `scala.Option` and from `scala.util.Try`.

Mailing list link: https://groups.google.com/forum/#!topic/liftweb/dl0nNfwt4EU


core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
```diff
@@ -106,19 +106,31 @@ class LAFuture[T](val scheduler: LAScheduler) {
...
+        wrapWithBox(f)(v) match {
+          case Full(successfullyComputedFuture) => successfullyComputedFuture.onComplete(v2 => ret.complete(v2))
+          case e: EmptyBox => ret.complete(e)
+        }
       case e: EmptyBox => ret.complete(e)
     })
     ret
   }
 
+  private def wrapWithBox[A](f: T => A)(t: T): Box[A] = {
```
Shadowfiend: Isn't this just `tryo`?

arkadius: It's allmost the same but there is a little differences: it takes a function not a call-by-name parameter. In tryo you must write:
```scala
box.flatMap { v =>
  tryo(f(v))
}
```

instead of:
```scala
box.flatMap(wrapWithBox(f))
```

The second thing is that module *util* uses module *actor* - at least in *Mailer* there is used *SpecializedLiftActor*. But we can solve this, moving tryo into module *common* (which is IMO the better place for it).

arkadius:
What do you think about an idea of moving `tryo` from *util* to *common*? It is Box-related feature so maybe it would be a better place? After that I could reuse it in my change.

Reporeted bug is quite serious because client code in many case can rely on assumption that LAFuture will eventually by completed e.g. quite idiomatic is to create scheduled job that will be periodically repeated after future completion.

Shadowfiend:
I could see moving `tryo` to the `Box` singleton and making the versions in `Helpers` stubs that refer back to the `Box` singleton's stuff.

arkadius:
I've prepared suggested change but it causes ambiguousity errors when smb import both Box._ and Helpers._ (Box has also e.g. implicit conversions from Option). I fixed it by extracting new Tryo object. What do you think about that?

Shadowfiend:
Hmm… A fair point. Not sure there's a good reason to ever import `Box._` though… The implicits are in implicit resolution scope by virtue of being defined on the `Box` singleton, and the non-implicit functions are perhaps “best” invoked on `Box` directly.

Can't rule out that people are using it this way though. I'll follow up on the list, want to see what other folks think is the past path.

arkadius:
For minimizing the mess I removed Tryo object and added extensions for Option/Try adding explicit toBox conversion. If smb was using import of Box._ for conversion Option -> Box, now can repleace it with toBox. What do you think about that?

arkadius:
I haven't seen your answer on mailing list. If you think that those extensions are the bad idea, I can remove it from PR.

Shadowfiend:
I actually dig it. My bad for taking so long to reply on the list :)

Shadowfiend:
Okay, this looks good from where I stand, would like at least one other pair of eyeballs on the changes though. Any volunteers?

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
```diff
@@ -1840,8 +1839,8 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
     val isForm = !attrs.get("form").toList.isEmpty
 
     val eagerEval: Boolean =
-      (attrs.get("eager_eval").map(toBoolean) or
-        findNSAttr(attrs, "lift", "eager_eval").map(toBoolean) or
+      (attrs.get("eager_eval").map(toBoolean) orElse
```
fmpwizard: is this a change users will also have to make? (`s/or/orElse/g`)

fmpwizard:
could you add to the PR description that you also added an implicit from Try to Box? seems like something wee will want to highlight on release notes

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
arkadius: If they will import both Helpers._ and Box._, the have to. Other possibility is to explicitly convert first element of chain and other will be implicitly converted without Box._ import. eg.
```scala
ttrs.get("eager_eval").map(toBoolean).toBox or  findNSAttr(attrs, "lift", "eager_eval").map(toBoolean)
```

arkadius:
Done it, thanks for review.

fmpwizard:
:+1:

Shadowfiend:
Sweet, two thumbs, I think we're lookin' good for this one.

Issue #60, opened 2009-09-15
Setting cookies does not work in an ajax response


harryh:
I have an SHtml.ajaxSelect that, when executed sets a cookie:

S.addCookie(HTTPCookie("CITYID", city.id.toString))

and then returns a JsCmds.RedirectTo(uri) command.  The cookie isn't
actually being set (if I look at the live http headers the Set-Cookie
line isn't sent down in the http response to the ajax call).

http://groups.google.com/group/liftweb/browse_thread/thread/a9a7de2cb1f05544/

PR #1679, opened 2015-02-07
Fixed missing `s` in string interpolation

fmpwizard:
as discussed here
https://groups.google.com/d/topic/liftweb/jjV3vFhIReA/discussion

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/Paginator.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/Paginator.scala
@@ -315,7 +315,7 @@ trait SortedPaginatorSnippet[T, C] extends SortedPaginator[T, C] with PaginatorS
     val headerTransforms =
       headers.zipWithIndex.map {
         case ((binding, _), colIndex) =>
-          ".$binding *" #> { ns: NodeSeq =>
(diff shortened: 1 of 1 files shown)


farmdawgnation:
This looks :+1: but openjdk doesn't look happy. But that's an issue on Travis isn't it?

fmpwizard:
looking at the travis log, I see it could not find:
org.scala-lang#scala-compiler;2.10.4!scala-compiler.jar


and it's weird how here it shows fail, but I got an email from travis saying  : 

**Passed: lift/framework#60 (diego_issue_string_inter - dc632a9)**

Shadowfiend:
Yeah, so travis runs builds both on branch push and on PR open I believe. So you probably got a successful run of the branch push and a failed run on the PR.

I've seen that scala compiler resolution issue before… Not sure why we're even depending on the scala 2.10.4 compiler JAR—presumably there's a transitive dependency somewhere. The OpenJDK build seems a little questionable in general; it's also hung for >10 minutes causing travis to kill the build due to lack of activity. May be worth discussing turning it off unless/until someone is willing to give it some attention.

fmpwizard:
the scala compiler 2.10.4 is the one sbt 0.13 needs :(
ok, merging this ...

PR #1682, opened 2015-02-10
Fix Comet clean-up bug

pbrant:
Comet actors which defined a life span were being half-cleaned up with correspondingly strange results. 

The cause was java.util.Map#remove taking an Object, not a K with the result that this was missed when migrating from a tuple to CometId. I assume remove(Object key) was for backwards compatibility reasons, but seriously, come on!


Shadowfiend:
-.-

:+1:

farmdawgnation:
Using `Object`:

fail

PR #1683, opened 2015-02-14
Documentation Ape: Documentation updates to JsonAST and a few small code optimizations.

farmdawgnation:
dawn-planet-apes-140709-700

This PR does some updates to the documentation in JsonAST.

ML Thread about these changes: https://groups.google.com/d/msg/liftweb/b-xB7zVNzUQ/4z3485XEti0J

### DIFF
--- a/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
+++ b/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
@@ -314,6 +314,9 @@ object JsonAST {
      * > JArray(JInt(1) :: JInt(2) :: Nil).children
      * List(JInt(1), JInt(2))
      * }}}
+     *
+     * @return Direct children of this `JValue` if it is a `[[JObject]]` or
+     * `[[JArray]]`, or `[[JNothing]]` otherwise.
      */
     def children: List[JValue] = this match {
       case JObject(l) => l map (_.value)
@@ -366,7 +369,7 @@ object JsonAST {
     }
 
     /**
-     * Return a new `JValue` resulting from applying the given function to each value.
+     * Return a new `JValue` resulting from applying the given function to each value, recursively.
      *
      * If this function is invoked on a `[[JObject]]`, it will iterate over the field values of that `JObject`.
      * If this function is invoked on a `[[JArray]]`, it will iterate over the values of that `JArray`.
@@ -393,7 +396,7 @@ object JsonAST {
 
     /**
      * Return a new `JValue` resulting from applying the given function to each `[[JField]]` in a `[[JObject]]` or a
-     * `[[JArray]]` of `JObject`.
+     * `[[JArray]]` of `JObject`, recursively.
      *
      * Example:
      *
@@ -403,6 +406,8 @@ object JsonAST {
      *   case x => x
      * }
      * }}}
+     *
+     * @see transformField
      */
     def mapField(f: JField => JField): JValue = {
       def rec(v: JValue): JValue = v match {
@@ -413,39 +418,71 @@ object JsonAST {
       rec(this)
     }
 
-    /** Return a new JValue resulting from applying the given partial function <code>f</code>
+    /** Return a new `JValue` resulting from applying the given partial function `f``
      * to each field in JSON.
-     * <p>
-     * Example:<pre>
+     * 
+     * Example:
+     * {{{
      * JObject(("age", JInt(10)) :: Nil) transformField {
      *   case ("age", JInt(x)) => ("age", JInt(x+1))
      * }
-     * </pre>
+     * }}}
      */
     def transformField(f: PartialFunction[JField, JField]): JValue = mapField { x =>
       if (f.isDefinedAt(x)) f(x) else x
     }
 
     /**
-     * Return a new `JValue` resulting from applying the given partial function to each value.
+     * Return a new `JValue` resulting from applying the given partial function
+     * to each value within this `JValue`.
      *
-     * Example:
+     * If this is a `JArray`, this means we will transform each value in the
+     * array and return an updated array.
+     *
+     * If this is a `JObject`, this means we will transform the value of each
+     * field of the object and the object in turn and return an updated object.
+     * 
+     * If this is another type of `JValue`, the value is transformed directly.
+     *
+     * Note that this happens recursively, so you will receive both each value
+     * in an array ''and'' the array itself, or each field value in an object
+     * ''and'' the object itself. If an array contains arrays, we will recurse
+     * into them in turn.
+     *
+     * Examples:
      *
      * {{{
-     * JArray(JInt(1) :: JInt(2) :: Nil) transform {
+     * > JArray(JInt(1) :: JInt(2) :: Nil) transform {
      *   case JInt(x) =>
      *     JInt(x+1)
      * }
+     * res0: net.liftweb.json.JsonAST.JValue = JArray(List(JInt(2), JInt(3)))
      * }}}
+     *
+     * Without type matching, notice that we get the result of the transform
+     * replacing the array:
+     *
+     * {{{
+     * > JArray(JInt(1) :: JInt(2) :: Nil) transform {
+     *   case _ =>
+     *     JString("hello")
+     * }
+     * res0: net.liftweb.json.JsonAST.JValue = JString("hello")
+     * }}}
+     *
+     * @return This `JValue` with its child values recursively transformed by
+     *         the given `PartialFunction`, when defined. If the
+     *         `PartialFunction` is undefined, leaves the child values
+     *         untouched.
      */
     def transform(f: PartialFunction[JValue, JValue]): JValue = map { x =>
       if (f.isDefinedAt(x)) f(x) else x
     }
 
     /**
-     * Return a new JValue resulting from replacing the value at the specified field
-     * path with the replacement value provided. This has no effect if the path is empty
-     * or if the value is not a `[[JObject]]` instance.
+     * Return a new `JValue` resulting from replacing the value at the specified field
+     * path with the replacement value provided. This has no effect if the path
+     * is empty or if the value is not a `[[JObject]]` instance.
      *
      * Example:
      *
@@ -479,16 +516,17 @@ object JsonAST {
      *
      * When invoked on a `[[JObject]]` it will first attempt to see if the `JObject` has the field defined on it.
      * Not finding the field defined, this method will recurse into the fields of that object and search for the
-     * value there. When invoked on or enccountering a `[[JArray]]` during recursion this method will run its search
-     * on each member of the JArray.
+     * value there. When invoked on or encountering a `[[JArray]]` during recursion this method will run its search
+     * on each member of the `JArray`.
      *
      * Example:
      *
      * {{{
-     * JObject(("age", JInt(2))) findField {
-     *   case (n, v) =>
+     * > JObject(JField("age", JInt(2))) findField {
+     *   case JField(n, v) =>
      *     n == "age"
      * }
+     * res0: Option[net.liftweb.json.JsonAST.JField] = Some(JField(age,JInt(2))
      * }}}
      */
     def findField(p: JField => Boolean): Option[JField] = {
@@ -507,7 +545,8 @@ object JsonAST {
      * Example:
      *
      * {{{
-     * JArray(JInt(1) :: JInt(2) :: Nil) find { _ == JInt(2) } == Some(JInt(2))
+     * > JArray(JInt(1) :: JInt(2) :: Nil) find { _ == JInt(2) }
+     * res0: Option[net.liftweb.json.JsonAST.JValue] = Some(JInt(2))
      * }}}
      */
     def find(p: JValue => Boolean): Option[JValue] = {
@@ -524,40 +563,64 @@ object JsonAST {
     }
 
     /**
-     * Return a List of all fields that match the given predicate.
+     * Return a `List` of all fields that match the given predicate. Does not
+     * recurse into child elements, so this will only check a `JObject`'s field
+     * values.
      *
      * Example:
      *
      * {{{
-     * JObject(("age", JInt(10)) :: Nil) filterField {
-     *   case ("age", JInt(x)) if x > 18 =>
+     * > JObject(JField("age", JInt(10))) filterField {
+     *   case JField("age", JInt(x)) if x > 18 =>
      *     true
      *
      *   case _ =>
      *     false
      * }
+     * res0: List[net.liftweb.json.JsonAST.JField] = List()
+     * > JObject(JField("age", JInt(10))) filterField {
+     *   case JField("age", JInt(x)) if x < 18 =>
+     *     true
+     *
+     *   case _ =>
+     *     false
+     * }
+     * res1: List[net.liftweb.json.JsonAST.JField] = List(JField(age,JInt(10)))
      * }}}
+     *
+     * @return A `List` of `JField`s that match the given predicate `p`, or `Nil`
+     *         if this `JValue` is not a `JObject`.
      */
     def filterField(p: JField => Boolean): List[JField] = 
       foldField(List[JField]())((acc, e) => if (p(e)) e :: acc else acc).reverse
 
     /**
-     * Return a List of all values which matches the given predicate.
+     * Return a List of all values which matches the given predicate, recursively.
      *
      * Example:
      *
      * {{{
-     * JArray(JInt(1) :: JInt(2) :: Nil) filter {
+     * > JArray(JInt(1) :: JInt(2) :: Nil) filter {
      *   case JInt(x) => x > 1
      *   case _ => false
      * }
+     * res0: List[net.liftweb.json.JsonAST.JValue] = List(JInt(2))
+     * }}}
+     *
+     * This operates recursively, so nested objects work too:
+     * {{{
+     * > ((("boom" -> ("slam" -> "hello")) ~ ("shap" -> 3)): JObject) filter {
+     *   case JString("hello") => true
+     *   case _ => false
+     * }
+     * res0: List[net.liftweb.json.JsonAST.JValue] = List(JString(hello))
      * }}}
      */
     def filter(p: JValue => Boolean): List[JValue] =
       fold(List[JValue]())((acc, e) => if (p(e)) e :: acc else acc).reverse
 
     /** 
-     * Create a new instance of [[WithFilter]] for Scala to use when using
+     * Create a new instance of `[[WithFilter]]` for Scala to use when using
      * this `JValue` in a for comprehension.
      */
     def withFilter(p: JValue => Boolean) = new WithFilter(this, p)
@@ -576,7 +639,7 @@ object JsonAST {
      *
      * {{{
      * > JArray(JInt(1) :: JInt(2) :: Nil) ++ JArray(JInt(3) :: Nil)
-     * JArray(List(JInt(1), JInt(2), JInt(3)))
+     * res0: JArray(List(JInt(1), JInt(2), JInt(3)))
      * }}}
      */
     def ++(other: JValue) = {
@@ -597,8 +660,8 @@ object JsonAST {
      * Example:
      *
      * {{{
-     * JObject(("age", JInt(10)) :: Nil) removeField {
-     *   case ("age", _) => true
+     * > JObject(JField("age", JInt(10))) removeField {
+     *   case JField("age", _) => true
      *   case _          => false
      * }
      * }}}
@@ -614,7 +677,8 @@ object JsonAST {
      * Example:
      *
      * {{{
-     * JArray(JInt(1) :: JInt(2) :: JNull :: Nil).remove(_ == JNull)
+     * > JArray(JInt(1) :: JInt(2) :: JNull :: Nil).remove(_ == JNull)
+     * res0: net.liftweb.json.JsonAST.JValue = JArray(List(JInt(1), JInt(2), JNothing))
      * }}}
      */
     def remove(p: JValue => Boolean): JValue = this map {
@@ -626,21 +690,17 @@ object JsonAST {
      * Extract a value into a concrete Scala instance from its `JValue` representation.
      *
      * Value can be:
-     *
-     * <ul>
-     *   <li>case class</li>
-     *   <li>primitive (String, Boolean, Date, etc.)</li>
-     *   <li>supported collection type (List, Seq, Map[String, _], Set)</li>
-     *   <li>any type which has a configured custom deserializer</li>
-     * </ul>
-     *
+     *  - a case class
+     *  - a primitive (String, Boolean, Date, etc.)>
+     *  - any type which has a configured [[TypeHints custom deserializer]]
+     *  - a supported collection type of any of the above (List, Seq, Map[String, _], Set)
      *
      * Example:
      *
      * {{{
      * > case class Person(name: String)
      * > JObject(JField("name", JString("joe")) :: Nil).extract[Person]
-     * Person("joe")
+     * res0: Person("joe")
      * }}}
      */
     def extract[A](implicit formats: Formats, mf: scala.reflect.Manifest[A]): A =
@@ -653,12 +713,10 @@ object JsonAST {
      * it will return a `[[scala.None]]` instead of throwing an exception as `[[extract]]` would.
      *
      * Value can be:
-     * <ul>
-     *   <li>case class</li>
-     *   <li>primitive (String, Boolean, Date, etc.)</li>
-     *   <li>supported collection type (List, Seq, Map[String, _], Set)</li>
-     *   <li>any type which has a configured custom deserializer</li>
-     * </ul>
+     *  - a case class
+     *  - a primitive (String, Boolean, Date, etc.)>
+     *  - any type which has a configured [[TypeHints custom deserializer]]
+     *  - a supported collection type of any of the above (List, Seq, Map[String, _], Set)
      *
      * Example:
      *
@@ -681,20 +739,17 @@ object JsonAST {
      * the default value instead.
      *
      * Value can be:
-     *
-     * <ul>
-     *   <li>case class</li>
-     *   <li>primitive (String, Boolean, Date, etc.)</li>
-     *   <li>supported collection type (List, Seq, Map[String, _], Set)</li>
-     *   <li>any type which has a configured custom deserializer</li>
-     * </ul>
+     *  - a case class
+     *  - a primitive (String, Boolean, Date, etc.)>
+     *  - any type which has a configured [[TypeHints custom deserializer]]
+     *  - a supported collection type of any of the above (List, Seq, Map[String, _], Set)
      *
      * Example:
      *
      * {{{
      * > case class Person(name: String)
      * > JNothing.extractOrElse(Person("joe"))
-     * Person("joe")
+     * res0: Person("joe")
      * }}}
      */
     def extractOrElse[A](default: => A)(implicit formats: Formats, mf: scala.reflect.Manifest[A]): A =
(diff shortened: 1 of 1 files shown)


core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -19,32 +19,97 @@ package json
 
 import scala.language.implicitConversions
 
+/**
+ * This object contains the abstract snytax tree (or AST) for working with JSON objects in lift-json.
```
Shadowfiend: Syntax*

Also, these lines seem a bit on the long side?

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -19,32 +19,97 @@ package json
...
+ *
+ * The purpose of the JsonAST is to represent and manipulate JSON by leveraging Scala language features like types,
```
Shadowfiend: Seems like “purpose of the JSON AST” would make more sense here, since you're referring to the concept rather than the class/singleton?

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -19,32 +19,97 @@ package json
...
+ * case classes, etc. The AST should allow you to represent anything you could imagine from JSON land using the Scala
+ * type system.
+ *
+ * Everything in the AST has a single root: JValue. A JValue could, quite literally, be anything. It could be an
+ * an object (represented by [[JObject]]), a string ([[JString]]), a null ([[JNull]]), and so on. So, when constructing
```
Shadowfiend: In #1650 I both linked and monospaced all class names (via backticks). Pretty easy change, just sticking backticks outside of the braces.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -19,32 +19,97 @@ package json
...
+ * {{{
+ * JObject(JField("bacon", JBool(true)) :: JField("spinach", JBool(false)))
+ * }}}
+ *
+ * Once serialized to the string representation of JSON you would end up with the following:
+ *
+ * {{{
+ * {
+ *   "bacon":true,
+ *   "spianch":false
```
Shadowfiend: spinach? :)

Shadowfiend: Though I see that's the same here and above, so that may be on purpose.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -19,32 +19,97 @@ package json
...
+ */
 object JsonAST {
   import scala.text.{Document, DocText}
   import scala.text.Document._
 
-  /** Concatenates a sequence of <code>JValue</code>s.
-    * <p>
-    * Example:<pre>
+  /**
+    * Concatenate a sequence of `JValue`s together.
```
Shadowfiend: Link?

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -19,32 +19,97 @@ package json
...
+    *
+    * This would be useful in the event that you have a handful of `JValue` instances that need to be
+    * smacked together into one unit.
```
Shadowfiend: It's not really for one “unit” generally, but specifically to combine a set of values into a JSON array, right?

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -19,32 +19,97 @@ package json
...
-  def concat(xs: JValue*) = xs.foldLeft(JNothing: JValue)(_ ++ _)
+  def concat(values: JValue*) = values.foldLeft(JNothing: JValue)(_ ++ _)
 
   object JValue extends Merge.Mergeable
 
   /**
-   * Data type for JSON AST.
+   * The base type for all things that represent distinct JSON entities in the AST.
+   *
+   * Most members of the AST will extend this class. The one exception is [[JField]] which does not extend this class
```
Shadowfiend: Won't remark on the rest of the backticks around links, just noting that there are more here though.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -19,32 +19,97 @@ package json
...
+   * because it really <em>can't</em> properly exist as a first-class citizen of JSON.
```
Shadowfiend: I think you can use double single-quotes ('') on either side instead of `em` tags.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -19,32 +19,97 @@ package json
...
     type Values
 
-    /** XPath-like expression to query JSON fields by name. Matches only fields on
-     * next level.
-     * <p>
-     * Example:<pre>
-     * json \ "name"
-     * </pre>
+    /**
+     * Find a child of a `JObject` or a `JArray` of `JObject` by name, and return `JNothing` for anything else.
```
Shadowfiend: I think losing the “XPath-like expression” bit might be a mistake, and that we may want to keep that as the summary/intro to the more detailed documentation you've added. The point of `\` and `\\` is precisely to parallel XPath syntax, so I think keeping that up front makes sense.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -19,32 +19,97 @@ package json
...
+     * }
+     * }}}
+     *
+     * If for some reason you're interested in taking a look at Joe's catchphrase, you can query it using the `\` method
+     * to find it like so:
+     *
+     * Example:
+     *
+     * {{{
+     * > json \ "catchphrase"
```
Shadowfiend: I've been keeping the `scala` in front of the `>` for Scala prompts, as a way to make it clear that we're evaluating something on the Scala REPL, fwiw.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -19,32 +19,97 @@ package json
...
+     * JObject(List(JField("name", JString("Alabama Cheer")), JField("value", JString("Roll tide"))))
```
Shadowfiend: Likewise, I've been including the `res0: `-type stuff the REPL adds for the same reason. Basically I copy-paste from the REPL :p

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -19,32 +19,97 @@ package json
...
+     * }}}
+     *
+     * Likewise, if you wanted to find Joe's name you could do the following:
+     *
+     * {{{
+     * > json \ "name"
+     * JString("Joe")
+     * }}}
+     *
+     * The result could be any kind of `JValue` that you could imagine.
```
Shadowfiend: I think the phrasing “that you could imagine” is kind of misleading. `JValue` is a sealed trait with clearly delineated subtypes, and we lose some precision in the informal language. Saying “The result could be any subclass of `JValue`” seems like a better approach.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -19,32 +19,97 @@ package json
...
+     * In the event that the `JValue` you're operating on is actually an array of objects, you'll get back a `JArray` of
+     * whatever the result of querying each of those object is. In the event nothing is found, you'll get a `JNothing`.
```
Shadowfiend: Maybe “get back a `JArray` containing the results of querying each object in turn”?

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
Shadowfiend: Also, for this particular description, may be worth being a little clearer on the `JNothing` bit and explicitly saying, in a separate sentence, “if you call `\` on something other than a `JObject` or `JArray`, you will always get a `JNothing` as there are no sub-properties to query”?

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -73,47 +138,68 @@ object JsonAST {
...
-      * </pre>
-      */
+    /**
+     * Find all children of a `JObject` with the matching name, returning an empty `JObject` is no matches are found.
+     *
+     * Example:
+     *
+     * {{{
+     * json \\ "name"
+     * }}}
```
Shadowfiend: Seems like this example might use some fleshing out.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -73,47 +138,68 @@ object JsonAST {
...
     }
 
-    /** XPath-like expression to query JSON fields by name. Returns all matching fields.
-      * <p>
-      * Example:<pre>
-      * json \\ "name"
-      * </pre>
-      */
+    /**
+     * Find all children of a `JObject` with the matching name, returning an empty `JObject` is no matches are found.
```
Shadowfiend: Does this one not work on `JArray`s?

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -73,47 +138,68 @@ object JsonAST {
...
-              a ::: find(value)
-            }
-        }
-        case JArray(l) => l.foldLeft(List[JField]())((a, json) => a ::: find(json))
-        case _ => Nil
+        case JObject(fields) =>
+          fields.foldLeft(List[JField]()) {
+            case (matchingFields, JField(name, value)) =>
+              matchingFields :::
+              List(JField(name, value)).filter(_.name == nameToFind) :::
```
Shadowfiend: You are a superhero for reformatting this :)

I would indent things past the first line, though, to make it clear it's part of the same expression.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -73,47 +138,68 @@ object JsonAST {
...
+          fields.foldLeft(List[JField]()) { (matchingFields, children) =>
+            matchingFields ::: find(children)
+          }
+
+        case _ =>
+          Nil
       }
+
       find(this) match {
         case JField(_, x) :: Nil => x
```
Shadowfiend: Random question: I can't remember if there's an issue about changing this so that it consistently returns `JObject`… Certainly seems odd that it doesn't.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -73,47 +138,68 @@ object JsonAST {
...
-      * <p>
-      * Example:<pre>
-      * json \ classOf[JInt]
-      * </pre>
-      */
+    /** 
+     * Find immediate children of this `JValue` that match the specific type.
+     *
+     * This is only particularly useful if you have a `JObject` and are interested in finding all children that are,
+     * say, integers. YOu might do so like this:
```
Shadowfiend: Stray caps!

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -73,47 +138,68 @@ object JsonAST {
...
+     *
+     * {{{
+     * json \ classOf[JInt]
```
Shadowfiend: This example could probably use a little more fleshing out (perhaps with some small sample data and an expected result).

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -73,47 +138,68 @@ object JsonAST {
...
 
-    /** XPath-like expression to query JSON fields by type. Matches only fields on
-      * next level.
-      * <p>
-      * Example:<pre>
-      * json \ classOf[JInt]
-      * </pre>
-      */
+    /** 
+     * Find immediate children of this `JValue` that match the specific type.
```
Shadowfiend: Maybe “that are of a particular `JValue` subclass”, and then lose the last sentence?

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -73,47 +138,68 @@ object JsonAST {
...
       findDirect(children, typePredicate(clazz) _).asInstanceOf[List[A]] map { _.values }
 
-    /** XPath-like expression to query JSON fields by type. Returns all matching fields.
-      * <p>
-      * Example:<pre>
-      * json \\ classOf[JInt]
-      * </pre>
-      */
+    /**
+     * Find all descendents of this `JValue` that match the specific type.
```
Shadowfiend: descendants*

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -73,47 +138,68 @@ object JsonAST {
...
+     *
+     * This is only particularly useful if you have a `JObject` and are interested in finding all descendents that are
+     * of a particular type. This method functions a lot like its single-backslash cousin, excepting that it will
+     * dive deep into the `JObject` to find matches.
+     *
+     * You might use this as follows:
+     *
+     * {{{
+     * json \\ classOf[JInt]
```
Shadowfiend: Again re: example—I actually really like this style of describing the analogue between `\` and `\\`, and I think it may be the right way to document the `String`-based `\\` above as well.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -73,47 +138,68 @@ object JsonAST {
...
-      * <p>
-      * Example:<pre>
-      * json \\ classOf[JInt]
-      * </pre>
-      */
+    /**
+     * Find all descendents of this `JValue` that match the specific type.
+     *
+     * This is only particularly useful if you have a `JObject` and are interested in finding all descendents that are
+     * of a particular type. This method functions a lot like its single-backslash cousin, excepting that it will
```
Shadowfiend: Could link to the single-backslash cousin, something like `[[\ single-backslash cousin]]` or `[[JValue.\ single-backslash cousin]]`, I think. Scaladoc's occasionally finicky about weird links to operators and overloads, though, so I may be wrong.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -122,28 +208,44 @@ object JsonAST {
...
 
-    /** Return nth element from JSON.
-      * Meaningful only to JArray, JObject and JField. Returns JNothing for other types.
-      * <p>
-      * Example:<pre>
-      * JArray(JInt(1) :: JInt(2) :: Nil)(1) == JInt(2)
-      * </pre>
-      */
+    /**
+     * Return `n`th element from a `JArray` or `JObject` and return `JNothing` for any other kind of `JValue`.
```
Shadowfiend: I guess if we're going to indicate `n`th, we should make the parameter name `n`, or make this say “`i`-th”. Also would suggest a dash there to be clear on the fact that it refers to the parameter, rather than just being the word “nth”.

Also may be worth clarifying:
 - What happens for a `JObject`.
 - What happens for non-`JArray`/`JObject` (presumably `JNothing`).

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -122,28 +208,44 @@ object JsonAST {
...
     def apply(i: Int): JValue = JNothing
 
-    /** Return unboxed values from JSON
-      * <p>
-      * Example:<pre>
-      * JObject(JField("name", JString("joe")) :: Nil).values == Map("name" -> "joe")
-      * </pre>
-      */
+    /**
+     * Return unboxed values from JSON.
```
Shadowfiend: May be worth expanding slightly on what “unboxed values” means here (i.e., the “contained” value in a native Scala type).

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -122,28 +208,44 @@ object JsonAST {
...
+     * This method is useful for getting all the values of a `JObject` or `JArray` and will return them as a
+     * `List[JValue]`. If the `JValue` you invoke this method on is not a `JObject` or `JArray` you will instead
+     * get `Nil`.
+     *
+     * Example:
+     *
+     * {{{
+     * > JArray(JInt(1) :: JInt(2) :: Nil).children
+     * List(JInt(1), JInt(2))
+     * }}}
```
Shadowfiend: Just remembered this—the cases where return values will diverge by the type of this `JValue` are good cases to use the `@return` tag, e.g.:

```scala
  /**
   * ...
   *
   * @return Direct children of this `JValue` if it is a `[[JObject]]` or `[[JArray]]`, or
   * `[[JNothing]]` otherwise.
   */
```

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -172,8 +277,11 @@ object JsonAST {
       rec(z, this)
     }
 
-    /** Return a combined value by folding over JSON by applying a function <code>f</code>
-     * for each field. The initial value is <code>z</code>.
+    /**
+     * Fold over a sries of `JField`s applying a function to each one.
```
Shadowfiend: series*

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -172,8 +277,11 @@ object JsonAST {
...
+     *
+     * @param z The initial value for the fold.
+     * @param f The Function to apply, which takes an accumulator as its first param and the next field as its second.
```
Shadowfiend: Is `Function` capitalized for a reason here, or is that a typo?

May also be worth saying “as its first parameter” rather than “as its first param”.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -239,26 +357,34 @@ object JsonAST {
...
 
-    /** Return a new JValue resulting from applying the given partial function <code>f</code>
-     * to each value in JSON.
-     * <p>
-     * Example:<pre>
-      * JArray(JInt(1) :: JInt(2) :: Nil) transform { case JInt(x) => JInt(x+1) }
-      * </pre>
-      */
+    /**
+     * Return a new `JValue` resulting from applying the given partial function to each value.
```
Shadowfiend: Each value in what? I think it may be worth enumerating `JObject` vs `JArray` behavior here as is done above, assuming it's different.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -239,26 +357,34 @@ object JsonAST {
...
+     *
+     * Example:
+     *
+     * {{{
+     * JArray(JInt(1) :: JInt(2) :: Nil) transform {
+     *   case JInt(x) =>
+     *     JInt(x+1)
+     * }
```
Shadowfiend: Probably worth showing this as a REPL session with the result.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -239,26 +357,34 @@ object JsonAST {
...
-      * <p>
-      * Example:<pre>
-      * JObject(List(JField("foo", JObject(List(JField("bar", JInt(1))))))).replace("foo" :: "bar" :: Nil, JString("baz"))
-      * // returns JObject(List(JField("foo", JObject(List(JField("bar", JString("baz")))))))
-      * </pre>
-      */
+    /**
+     * Return a new JValue resulting from replacing the value at the specified field
+     * path with the replacement value provided. This has no effect if the path is empty
+     * or if the value is not a `[[JObject]]` instance.
```
Shadowfiend: Maybe it might be clearer to say “this `JValue` will be returned unchanged if the path is empty or if this `JValue` is not a `[[JObject]]` instance”?

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -295,48 +432,65 @@ object JsonAST {
...
       }
+
       find(this)
     }
 
-    /** Return a List of all fields that match the given predicate.
-     * <p>
-     * Example:<pre>
+    /**
+     * Return a List of all fields that match the given predicate.
```
Shadowfiend: May be worth explicitly saying this doesn't recurse into child objects and such?

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -366,82 +522,107 @@ object JsonAST {
...
-      * case class Person(name: String)
-      * JObject(JField("name", JString("joe")) :: Nil).extract[Person] == Person("joe")
-      * </pre>
-      */
+    /**
+     * Extract a value into a concrete Scala instance from its `JValue` representation.
+     *
+     * Value can be:
+     *
+     * <ul>
```
Shadowfiend: You can use:

```
  * Value can be:
  *  - a case class
  *  - a JSON primitive (`String`, `Boolean`, `Date`, etc)
  *  - supported ...
```

To avoid HTML here.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -366,82 +522,107 @@ object JsonAST {
...
+     *   <li>case class</li>
+     *   <li>primitive (String, Boolean, Date, etc.)</li>
+     *   <li>supported collection type (List, Seq, Map[String, _], Set)</li>
+     *   <li>any type which has a configured custom deserializer</li>
```
Shadowfiend: Maybe worth linking to the deserializer-related class here?

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -366,82 +522,107 @@ object JsonAST {
...
-      * Example:<pre>
-      * case class Person(name: String)
-      * JObject(JField("name", JString("joe")) :: Nil).extractOpt[Person] == Some(Person("joe"))
-      * </pre>
-      */
+    /**
+     * Optionally extract a value into a concrete Scala instance from its `JValue` representation.
+     *
+     * This method will attempt to extract a concrete Scala instance of type `A`, but if it fails it will return
+     * a `[[scala.None]]` instead.
```
Shadowfiend: Maybe worth saying “instead of throwing an exception as `[[extract]]` would”?

Shadowfiend:
This looks really solid! Left various remarks throughout, mostly nitpicking to make sure things are consistent with some of the previous doc cleanup work and the kind of typo-spotting that always happens with a second pair of eyeballs :)

I did notice a couple of things:
 - Some inconsistency between monospacing and linking (linked things aren't monospaced, unliked things are)—suggested above that we do both whenever we're referring to a class name.
 - Some inconsistency with linking in general—there are a several places where we refer to another class but don't link it. Consistently linking is kind of annoying, but I think it's better than doing so inconsistently.
 - Some inconsistency in examples having vs not having the REPL-style “here's the code, here's the result”.

Except for typos I don't think most of these should prevent the PR from going in unless you feel motivated to deal with them now—this is already a big improvement over what was there.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
farmdawgnation: Honestly, I've always applied the same semantics to cons-like operators that I apply to `andThen` and `&` in CSS selectors. Is that not the style we want to use here?

Shadowfiend: I'll let someone else weigh in. I only apply that (admittedly strange) stylistic choice to `&` and `andThen`, and with `andThen` I only apply it when it's done alongside `&`, not in other cases.

Even done that way, I'm slowly changing my mind about it, as it removes a visual hint about what is supposed to be one statement vs what is a series of them.

farmdawgnation: I guess I've always looked at it as dependency between lines? So the fact that they're one statement to the compiler is immaterial to me if the individual lines could be removed and it would still have a valid effect.

fmpwizard: I have been slowly adopting the indentation of the second line too, specially on code from lift-json when using  `~` mostly because the IDE does it automatically and I noticed people new to scala think it is strange we align them all the left, instead of making the second line have more indentation.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
farmdawgnation: Not really guaranteed that you'll get a `JArray` out. See the implementation of `++`:

```scala
        case (JNothing, x) => x
        case (x, JNothing) => x
        case (JArray(xs), JArray(ys)) => JArray(xs ::: ys)
        case (JArray(xs), v: JValue) => JArray(xs ::: List(v))
        case (v: JValue, JArray(xs)) => JArray(v :: xs)
        case (x, y) => JArray(x :: y :: Nil)
```

So if you have a nothing on one side you just get back the thing that wasn't a nothing. I feel like saying "You'll get back a JArray" would be misleading there.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
farmdawgnation: Eh. The idiom "in turn" leaves a bad taste in my mouth, but I'll make an attempt to make this clearer all the same.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
farmdawgnation: So it does. Weird.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
farmdawgnation: Dunno....

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
farmdawgnation: Yeah I tried this a few time. It blew up in my face. I took five minutes to try and understand what it wanted me to do. Not understanding it after that amount of time I gave up. heh.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
farmdawgnation: Wow. Looks like the original docs were wrong:

```
scala> val obj = JObject(JField("bacon", JInt(2)) :: JField("apple", JInt(3)) :: Nil)
obj: net.liftweb.json.JsonAST.JObject = JObject(List(JField(bacon,JInt(2)), JField(apple,JInt(3))))

scala> obj(1)
res1: net.liftweb.json.JsonAST.JValue = JNothing
```

So that's a fun discovery. Updated accordingly.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
farmdawgnation: I don't quite know where that is at the moment to be honest.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
farmdawgnation: That was not working when I tried it.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
Shadowfiend: Yeah, that's fair. Linking in Scaladoc can be a bit finnicky and sometimes I've gotten to a point where I just can't get it to work. Kind of annoying -.-

Shadowfiend:
Okay, pushed some further fixes/tweaks for the rest of my comments. Will merge this in the next couple of days barring any complaints.

fmpwizard:
Rememer to wait until I push the security patches to master before getting this to master. I'm also changing the milestone, as M4 is already out

Shadowfiend:
Yep!

Shadowfiend:
Looks good, let's do it.

Issue #1641, opened 2014-11-23
Split LiftSession into multiple source files


farmdawgnation:
The LiftSession source is current 3,281 lines long. Notably the LiftSession implementation alone is > 2k lines, so this is probably something we could split up into different traits to make it a bit easier to grok.

farmdawgnation:
We're already late in the Milestone cycle for 3.0 and I don't feel comfortable doing such a big refactor at this point. I vote we punt until 3.1.

fmpwizard:
Sounds good

dpp:
FWIW... when we split Helpers to multiple files/traits it was a huge disaster. People use some but not all the traits. Navigating the traits is a pain rather than searching for a function in a file (less of an issue with an IDE). There were all forms of problems with inner classes.

I'd vote :-1: on splitting LiftSession.

farmdawgnation:
I think given the feedback on this I've come to the following conclusion.

First, I think we have places in `LiftSession` where the semantic coupling that's in place just doesn't make sense. I'd argue that splitting `LiftSession` wouldn't be the same as splitting `Helpers` because Helpers was and is kind of a catch-all semantically speaking.

Second, I think I've decided doing this work doesn't make sense as a discrete unit of work. I've come to the opinion that we'll deliver better results by separating these out _when_ that semantic coupling itself becomes problematic or awkward for some other goal we're trying to achieve. This ticket was refactoring for refactoring's own sake in some ways - which can have its virtues, but I'm less convinced the payoff is there than I was when I opened the ticket.

I'm going to close this for now, but if there's disagreement on this point we can have a discussion on the ML and reopen later.

Shadowfiend:
:100: from me on the conclusion. Targeted, opportunistic refactoring of bad coupling will reap a lot of rewards… As you mentioned, much of `LiftSession` is a tangle of bad coupling, and we've already taken some swipes at it. I'll be taking some more as I continue working on refactoring snippet handling (for #1799 and friends).

PR #1684, opened 2015-02-14
Session IPA: Split out SessionMaster and ScopedLiftActor from LiftSession

farmdawgnation:
session ipa

My eyes were bigger than my stomach in #1641 – the changes that are going to be required to do that properly are going to require a lot of discussion on the list so for now I say we consider that ticket closed after this PR is merged and kick of a discussion sometime in the future about how to look at LiftSession again. For this one, there were two things in LiftSession that were relatively easy to move:

- SessionMaster because it is sufficiently complex imo to warrant its own file.
- ScopedLiftActor because it's apparently not used anywhere in session management period. It's just kind of there, and warrants its own file just by being not fundamentally related to LiftSession.


Shadowfiend:
These look like straight copy-paste jobs, no extraneous changes, right?

Shadowfiend:
(If so, I'm :+1: !)

farmdawgnation:
That's correct.

fmpwizard:
I think I know the answer (no) but is there a way that we could keep the history of those two new files?, as it stands on this PR, it will look as if the code for those two files started on this PR

Shadowfiend:
I don't think there's a way to keep traceability, no, though the individual lines of code may well be traceable through file changes.

fmpwizard:
tnx, alright, :+1:  from me too so merging in ...

PR #1688, opened 2015-03-13
Get S._attrs via safe accessor, avoiding NPE.

steven-sortable:
As the code in the accessor makes clear, _attrs.value can be null.


Shadowfiend:
As per our [contributing guidelines](https://github.com/lift/framework/blob/master/CONTRIBUTING.md) we ask that non-committers start a thread on the [mailing list](https://groups.google.com/forum/#!forum/liftweb) before opening tickets. If you can open up a thread we can confirm the issue and ask for the PR. :)

If there happens to already be a discussion that I missed (not at all impossible), please link it in the description. Thanks!

PR #1692, opened 2015-03-21
Midair Collision: Fix issue where binding an `href` and an `onclick` together could fail.

### LINKED ISSUE #1691: onclick event-handlers are not proparly applied on a-tags when href="#"
Ref: https://groups.google.com/forum/#!topic/liftweb/Ks3MIfqepRQ

Shadowfiend:
This only manifested when the `onclick` came before the `href` on an
`a` element, and probably also manifested if you bound an `onsubmit`
on a form that had an `action` attribute. The cause was just that I had
forgotten to properly pass on already-seen event attributes when
processing the href attributes, so we lost them.

Fixes #1691 .


andreak:
+1

Works for me.

Shadowfiend:
Okay, all passing, so let's do it.

PR #1693, opened 2015-03-21
Implement ParamFailure chaining for ~>

### LINKED ISSUE #1680: Box method ~> behaviour on ParamFailure
Current behaviour of ~> on a ParamFailure is to override the parameter with the new parameter passed to ~>. This means that when applying two successive ~>, the last application will be used. This is in contrast to the current behaviour of ?~, which only use the first application, and ignore the following ones (if the Box is already a Failure, applying ?~ will return the same failure).

A use case illustrating the poential issue:
for {
  id <- S.param("id") ?~ "Missing id" ~> 400
  item <- (DB.find(id) ?~ "Item not found" ~> 404).filter(canViewItem) ?~ "Unauthorized access to item" ~> 403
} yield ???

In case the item id is not in the DB, the message will be "item not found" along with error code 403, which is very counter-intuitive.

Discussion started in the google group: https://groups.google.com/forum/#!topic/liftweb/IMJqy2aMtXk

farmdawgnation:
Now, when ~> is invoked in a ParamFailure, the existing ParamFailure
will be chained, thereby preserving the original param that was
attached.

closes #1680

### DIFF
--- a/core/common/src/main/scala/net/liftweb/common/Box.scala
+++ b/core/common/src/main/scala/net/liftweb/common/Box.scala
@@ -1037,6 +1037,9 @@ final class ParamFailure[T](override val msg: String,
       case null => 0
       case x => x.hashCode()
     })
+
+    override def ~>[T](errorCode: => T): ParamFailure[T] =
+      ParamFailure(msg, exception, Full(this), errorCode)
   }
 
 /**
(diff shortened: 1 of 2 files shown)


Shadowfiend:
Looks like all of these are failing due to Iran's switch to daylight savings happening on the 22nd of March… Going to look at it again tomorrow and see if things are behaving better.

farmdawgnation:
Going to wait a few more days before merging this. Would love some feedback if you folks have time, but barring that I'm going to merging this soon!

farmdawgnation:
```
[info] LAFuture should
[info] x map to failing future if transforming function throws an Exception
[error]    the value is not equal to 'true' (LAFutureSpec.scala:27)
```

Only on openjdk 7.... that's weird.

Shadowfiend:
Lookin goood!

Issue #1689, opened 2015-03-16
NoSuchElementException in LiftSession:1012 (2.6)


andreak:
In LiftSession.postPageJavaScript we make a wrong assumption of a previously checked Some:
```
2015-03-13 09:14:14,390 ERROR [liftExecutor-1-thread-918] - [] []  ActorLogger                    - Actor threw an exception
java.util.NoSuchElementException: None.get
        at scala.None$.get(Option.scala:347)
        at scala.None$.get(Option.scala:345)
        at net.liftweb.http.LiftSession.run$1(LiftSession.scala:1012)
        at net.liftweb.http.LiftSession.postPageJavaScript(LiftSession.scala:1024)
        at net.liftweb.http.LiftSession$$anonfun$postPageJavaScript$1.apply(LiftSession.scala:1040)
        at net.liftweb.http.LiftSession$$anonfun$postPageJavaScript$1.apply(LiftSession.scala:1039)
        at scala.collection.immutable.List.flatMap(List.scala:327)
        at net.liftweb.http.LiftSession.postPageJavaScript(LiftSession.scala:1039)
        at net.liftweb.http.S$$anonfun$2.apply(S.scala:845)
        at net.liftweb.http.S$$anonfun$2.apply(S.scala:844)
        at net.liftweb.common.Full.map(Box.scala:610)
        at net.liftweb.http.S$class.jsToAppend(S.scala:844)
        at net.liftweb.http.S$.jsToAppend(S.scala:47)
        at net.liftweb.http.LiftServlet$$anonfun$21$$anonfun$apply$14$$anonfun$apply$15.apply(LiftServlet.scala:769)
        at net.liftweb.http.LiftServlet$$anonfun$21$$anonfun$apply$14$$anonfun$apply$15.apply(LiftServlet.scala:769)
        at net.liftweb.http.CoreRequestVarHandler$$anonfun$generateSnapshotRestorer$1$$anonfun$apply$8$$anonfun$apply$9$$anonfun$apply$10$$anonfun$apply$11.apply(Vars.scala:546)
        at net.liftweb.util.ThreadGlobal.doWith(ThreadGlobal.scala:71)
        at net.liftweb.http.CoreRequestVarHandler$$anonfun$generateSnapshotRestorer$1$$anonfun$apply$8$$anonfun$apply$9$$anonfun$apply$10.apply(Vars.scala:545)
        at net.liftweb.util.ThreadGlobal.doWith(ThreadGlobal.scala:71)
[...]

PR #1694, opened 2015-03-21
Null-Anti-Null: Don't assume postPageFunctions are Some.

Shadowfiend:
When generating post-page JavaScript, we were assuming that because our first
access of the post-page functions was non-empty, all of them would be. This led
to occasional .get-related NullPointerExceptions. We now allow for later calls
to the post-page functions to be None, and simply assume we have no more
functions to fetch once we get a None for them.

Fixess #1689 .


Shadowfiend:
Once this guy gets passing, I'm going to go ahead and merge 'im.

Shadowfiend:
Success!

PR #1696, opened 2015-04-13
Replace defunct scala-tools.org reference

natekupp:
scala-tools.org is no longer active. Pointing to sonatype.org for the jar instead.

### DIFF
--- a/core/json/README.md
+++ b/core/json/README.md
@@ -60,7 +60,7 @@ Add dependency to your pom:
 
 Download following jars:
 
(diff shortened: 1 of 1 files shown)


farmdawgnation:
Generally, it's a good idea to make sure that you've vetted a change on the mailing list before opening a PR and link to the ML thread from the PR. This is small enough and obvious enough that I'm not going to pitch too much of a fit about it, though. Just for future reference.

Thanks for the contribution.

PR #1697, opened 2015-04-16
Added extra arguments to method to allow for lifecycle support

dpp:
Added optional extra arguments to `serverActorForClient` to allow for lifecycle support.

Made `buildAndStoreComet` public so that any app can add a programmatically created CometActor


dpp:
Wait... need to change some params

PR #1698, opened 2015-04-17
Enhanced Support for browser Actor proxies

dpp:
Added support to `serverActorForClient` to register for life cycle events as well as filtering the data before
it's sent to the client.

Opened and documented `serverActorForClient` so folks can add custom CometActors that
are programmatically generated rather than named.

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
@@ -2047,8 +2047,8 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
    * @return a server-side Actor that is a proxy for the client-side Actor/Function call
    *
    */
-  def serverActorForClient(toCall: String, setupFunc: Box[() => Unit] = Empty,
-                            shutdownFunc: Box[() => Unit] = Empty,
+  def serverActorForClient(toCall: String, setupFunc: Box[LiftActor => Unit] = Empty,
+                            shutdownFunc: Box[LiftActor => Unit] = Empty,
                             dataFilter: Any => Any = a => a): LiftActor = {
     testStatefulFeature{
       val ca = new CometActor {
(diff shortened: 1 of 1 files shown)


web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
```diff
@@ -2372,7 +2395,20 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
...
+   *
+   * This is dangerous stuff, so use with caution.
+   *
+   * @param newCometFn a Function that takes `CometCreateInfo` and returns a CometActor
+   *                   instance
+   * @param creationInfo the info to send to the above function
+   * @tparam T A LiftCometActor or better
+   * @return The created, registered, etc. CometActor
+   */
+  def buildAndStoreComet[T <: LiftCometActor](newCometFn: (CometCreationInfo)=>Box[T])(creationInfo: CometCreationInfo): Box[T] = {
```
Shadowfiend: This is like way late for me to be asking this, @dpp, but what was the reasoning behind exposing this instead of just using `S.findOrCreateComet[T]`, which also handles ensuring that names are unique in the session? Wanted to drop a note about it in the release notes.

dpp: I needed a place to dynamically create CometActors rather than refer to a CometActor that had been compiled with the project. Now how can one dynamically subclass a JVM class without compiling it with the project? Yep... it was some of the Scala/Clojure/Lift/Reagent inter-op stuff I've been doing.

But more broadly, I realized that there might be times even in pure Scala-land that one might want to do a:

```
new CometActor { ..... }
```

Rather than creating a separate CometActor subclass in a file with a name.

Shadowfiend: Aha! So the key here is the access to `newCometFn` that can produce an arbitrary `T <: LiftCometActor`, right? Can you drop a sample use here so I can use it in docs/release notes, by any chance? If not I'll just do some digging around myself :)

dpp: Okay... here's an example https://github.com/dpp/devoxx_pl_2015_scala_clojure/blob/alt_comet/src/main/scala/sb/snippet/Actorize.scala#L29

I wound up extending `serverActorForClient` to include the lifecycle call-backs... but I liked the idea of being able to create an arbitrary CometActor class and maybe close over local scope rather than do the wonky "send a message and put the variables from the message in `var`s" paradigm. So I left `buildAndStoreComet` public.

PR #1699, opened 2015-04-19
Test Fixes: Drop some sleeps in LAFutureSpec.

Shadowfiend:
A couple of these specs are behaving intermittently in the CI environment, and
the sleeps seem beside the point to these two specs.

There's still an intermittent failure related to the Iran/Persian timezone that I'm
going to try and nail down before merging this, I think.


Shadowfiend:
Going to go ahead and merge this even though it doesn't seem to fix the intermittent failures :/

PR #1701, opened 2015-04-25
Add data-lift-fixedeventattribute-<event-name> for each replaced event-attribute

andreak:
Add data-lift-fixedeventattribute-<event-name> for each replaced event-attribute, with settings in LiftRules: LiftRules.includeFixedEventAttributesAsDataAttributes_? (default false)

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
@@ -562,44 +562,63 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
   @volatile var displayHelpfulSiteMapMessages_? = true
 
   /**
-   * Set to true if you want to add an attribute of the form data-lift-removed-attributes="removedattribute1 removedattribute2 ..."
-   * for each event-attribute (on*) which is replaced as part of the CSP.
-   * <br/>
-   * Example for the onclick attribute.
+   * The attribute used to expose the names of event attributes that
+   * were removed from a given element for separate processing in JS.
+   * By default, Lift removes event attributes and attaches those
+   * behaviors via a separate JS file, to avoid inline JS invocations so
+   * that a restrictive Content-Security-Policy can be used.
    *
-   * <pre>
-   * An element looking like this
+   * You can set this variable so that the resulting HTML will have
+   * attribute information about the removed attributes, in case you
+   * need them for e.g. CSS selector matching. The attribute will
+   * contain a space-separated list of JS attributes that were removed
+   * by Lift's processing.
    *
-   * &lt;span onclick="someStuff"&gt;
+   * For example, if you needed to match elements with an `onclick`
+   * attribute in CSS, you would usually do:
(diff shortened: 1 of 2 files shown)


Shadowfiend:
Two broad thoughts on the feature:
 - Switch instead to `data-lift-removed="<attribute name> <attribute name> <attribute name>"`.
 - Switch instead to adding a *class* `lift-removed-<attribute> lift-removed-<attribute>`.

Thought on both fronts is that having one attribute to represent these would be nice. Using a class is nice because there's such strong and optimized support for selection on classes in browsers.
Thoughts?

andreak:
I somewhat think that "removed" is a little bit too general in this case to I propose using
 ```data-lift-removedattributes="onclick onmouseover"```

Ok?

andreak:
A colleague of mine pointed out that the name should be ```data-lift-removed-attributes```so that it becomes el.dataset.liftRemovedAttributes in javascript

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -562,6 +562,44 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
...
+   * }
+   * </pre>
+   * with similar matching the data-lift-removed-attributes="onclick":
+   * <pre>
+   * [data-lift-removed-attributes~=onclick] {
+   *   text-decoration: underline;
+   * }
+   * </pre>
+   */
+  @volatile var includeFixedEventAttributesAsDataAttributes_? = false
```
Shadowfiend: Two questions:
 - Should this be `includeRemovedEventAttributesAsDataAttributes_?`?
 - Let's use Scaladoc markup rather than HTML (triple-backticks for multiline code blocks, single backticks for inline code blocks, empty lines for line breaks, etc).

andreak: 1. Yes, Fixed in last commit.
2. I'm not able to make this change in a way that IntelliJ will recognize the markup and show the same as the HTML-version. Can you  make the changes you suggest?

andreak:
I added a configurable setting, LiftRules.removedEventAttributesAttributeName = "data-lift-removed-attributes", so that one can change that if wanted.

Shadowfiend:
I'm going to merge the two `LiftRules` into `attributeForRemovedEventAttributes` and make that an `Option`, if that's cool by you—a `None` will simply not emit the removed attributes into the markup. Sound good?

andreak:
That's of course much better, thanks!

Shadowfiend:
Schweet, give that a shot and see what you think!

andreak:
Just tested it and it works great!
Merge it in:-)

Shadowfiend:
:thumbsup: I'll leave it here for another day or two in case anyone else wants to make some remarks and then :shipit: .

Shadowfiend:
All righty let's do it!

andreak:
:thumbsup:

PR #1702, opened 2015-04-27
Dcb issue 1677, send pre-creation CometActor messages in the order received

### LINKED ISSUE #1677: LiftSession.cometSetup wrong message order
When you send messages to a Comet Actor with session.sendCometActorMessage which isn't set up yet, Lift stores messages in LiftSession.cometSetup List in wrong order.

```scala
def setupComet(theType: String, name: Box[String], msg: Any) {
  testStatefulFeature {
    cometSetup.atomicUpdate(v => (Full(theType) -> name, msg) :: v)
  }
}
```
Later, when the Comet Actor is set up, the messages are sent back to the actor in LiftSession.findComet:

```scala
for {
  actor <- ret
  (cst, csv) <- cometSetup.is if cst == what
} actor ! csv
```

Notice that Lift prepends messages to cometSetup List and reads those from head to tail, hence it sends messages in reverse order. Which is, well, not what you expect from an actor.

dcbriccetti:
fixes #1677


web/webkit/src/test/scala/net/liftweb/http/LiftSessionSpec.scala
```diff
@@ -0,0 +1,57 @@
...
+    "Send accumulated messages to a newly-created comet actor in the order in which they arrived" in {
+
+      val session = new LiftSession("Test Session", "", Empty)
+      S.init(Empty, session) {
+        val cometName = "TestCometActor"
+        val sendingMessages = 1 to 20
+        sendingMessages foreach (message =>
+          session.sendCometActorMessage(cometName, Full(cometName), message))
+        session.findOrCreateComet[TestCometActor](Full(cometName), NodeSeq.Empty, Map.empty).map(comet =>
+          comet !? NoOp /* Block to allow time for all messages to be collected */)
```
Shadowfiend: Can you just mix in `MockLiftActor` to the test comet to avoid having to block?

Shadowfiend: (Conveniently, it also provides access to the list of received messages with no additional plumbing.)

dcbriccetti: MockLiftActor is appealing, but even if I can get it to compile, I don’t see anything in it that will block the call to get all the messages until they have all arrived. I added the NoOp message because without it (or a sleep), I was calling for the messages before they had arrived. 

  class TestCometActor extends MockLiftActor with CometActor {
    def render = NodeSeq.Empty
  }

Error:(28, 9) overriding method messageHandler in class MockSpecializedLiftActor of type => PartialFunction[Any,Unit];
 method messageHandler in trait BaseCometActor of type => PartialFunction[Any,Unit] has weaker access privileges; it should be public
  class TestCometActor extends MockLiftActor with CometActor {
        ^

Shadowfiend: Oof. Yeah, that's annoying.

`MockLiftActor` obviates the need for sleeping by making the comet message-send operation a function call rather than a multi-threaded message send.

We may want to introduce a `MockCometActor` that does very similar stuff, or a `BaseMockLiftActor` that doesn't implement the message handler… @farmdawgnation thoughts?

dcbriccetti: Nothing from Matt, and I’d like to close this before I get busy again. Maybe we can refactor later.

farmdawgnation: Sorry for the delay. I, uh, have been too aggressive in clearing out my inbox lately it seems. Yeah, I think a reorg of that mock actor code makes sense. I'd basically want to mirror however the real actors do it, so I'd need to take a look at that. Either way, if this is a fix for the `cometSetup` issue, let's not block that on a mock reorg.

Shadowfiend: Alll righty, that sounds good.

Shadowfiend:
So I guess this should be good to go then? I'll wait a bit to see if anyone else remarks and otherwise merge.

Shadowfiend:
Uh-oh, looks like this build is failing due to issues with the sbt JAR file. Do we need to bump our sbt version or something?

dcbriccetti:
Good to go as far as I’m concerned.

Shadowfiend:
Yeah, any sbt change wouldn't be related to this branch. Let's do it.

PR #1703, opened 2015-04-28
Fixed the new bind-less password change in ProtoUser

dpp:
Fixed an issue with the way CSS Selectors work so that password change in ProtoUser works again.

### DIFF
--- a/persistence/proto/src/main/scala/net/liftweb/proto/ProtoUser.scala
+++ b/persistence/proto/src/main/scala/net/liftweb/proto/ProtoUser.scala
@@ -990,10 +990,12 @@ trait ProtoUser {
         }
       }
 
+      val passwordInput = SHtml.password_*("",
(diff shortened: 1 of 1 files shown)


Shadowfiend:
Just to check here, the issue at hand was that the CSS bind will bind two separate password inputs that will each receive one value, instead of creating one password input bound in two places, right?

dpp:
Yep... 2 different functions, 2 different values... neither sets the right thing in the duplicate password in the User

Shadowfiend:
Yep, I've been bitten by the laziness of CSS selector binds like this before… Obviously I haven't learned my lesson yet :disappointed: 

:+1: on the change, let's get it in.

hedefalk:
Ah, got it! Thanks!

And sorry for taking your time, I just _hate_ to drop it when I don't understand something…

Shadowfiend:
As well it should be :)

PR #1705, opened 2015-05-23
Log if JSON parsing fails when handling a jsonCall invocation.

### LINKED ISSUE #1615: We don't indicate any sort of error if you make an AJAX request to an endpoint requiring JSON with invalid JSON
We ran into an annoying edge case in how we handle JSON AJAX requests.

We were using `SHtml.jsonCall` to wire up some stuff on the client to the server. This is all fine and dandy until you submit something that doesn't result in proper JSON to the function name. No error whatsoever is generated on the server or the client.

That... should probably be fixed.

farmdawgnation:
This PR is designed to address an issue where the call sent to a function wired up to a `jsonCall` isn't something that the JSON parser can actually parse. This tends to happen in cases where developers are doing some custom things with the `jsonCall` system - a bit more custom than just wiring it up to a button directly in a CSS transform.

To test this, you can set up a local Lift project and bind a JSON call to the browser somewhere, then invoke it from the console.

screen shot 2015-05-23 at 11 08 33 am

In my code, I had it print what it got from the client if it succeeded. If it failed, the framework would log its error to the console. Like so:

screen shot 2015-05-23 at 11 08 24 am

closes #1615

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/SHtml.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/SHtml.scala
@@ -42,7 +42,7 @@ object SHtml extends SHtml
  * The SHtml object defines a suite of XHTML element generator methods
  * to simplify the creation of markup, particularly with forms and AJAX.
  */
-trait SHtml {
+trait SHtml extends Loggable {
 
   /**
    * Convert a T to a String for display in Select, MultiSelect,
@@ -169,6 +169,21 @@ trait SHtml {
     ajaxCall_*(jsCalcValue, jsContext, SFuncHolder(func))
 
   /**
+   * A wrapper around parseOpt for use in the jsonCall variants that will notify the developer on the console when
+   * parseOpt is unsuccessful at parsing the JSON.
+  **/
+  private def parseOptOrLog(rawJsonText: String) = {
+    parseOpt(rawJsonText) match {
+      case None =>
+        logger.error(s"Browser sent back something that wasn't JSON to a jsonCall function: $rawJsonText")
+        None
+
+      case anythingElse =>
+        anythingElse
+    }
+  }
+
+  /**
    * Build a JavaScript function that will perform a JSON call based on a value calculated in JavaScript.
    * This method uses the Lift-JSON package rather than the old, slow, not-typed JSONParser.  This is the preferred
    * way to do client to server JSON calls.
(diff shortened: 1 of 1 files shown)


web/webkit/src/main/scala/net/liftweb/http/SHtml.scala
```diff
@@ -169,6 +169,21 @@ trait SHtml {
     ajaxCall_*(jsCalcValue, jsContext, SFuncHolder(func))
 
   /**
+   * A wrapper around parseOpt for use in the jsonCall variants that will notify the developer on the console when
+   * parseOpt is unsuccessful at parsing the JSON.
+  **/
+  private def parseOptOrLog(rawJsonText: String) = {
```
farmdawgnation: Is this something that'd be useful in the JSON parser itself?

farmdawgnation:
@Shadowfiend Hey, so in the ticket that I mentioned above you requested an overload of `jsonCall` that allows you to feed in a fallback function to provide a hook for the developer to handle the case the JSON couldn't be parsed.

The problem with that is that it'll take us from three overloads of `jsonCall` to six unless we make breaking changes to the API or we're horribly inconsistent. I'm honestly not in love with the idea. I'm kind of thinking it may be worth waiting to see how common it is that people need to do this and say in the meantime use `ajaxCall` and parse your own JSON. If we find people asking for this a lot we can always come back and add it, but I'm not sure it's going to be useful enough weighed against the API complexity.

If you disagree we could take it to the ML and discuss there.

farmdawgnation:
Still waiting for some feedback from ya here @Shadowfiend

web/webkit/src/main/scala/net/liftweb/http/SHtml.scala
Shadowfiend: I think if this went into `lift-json` it would best be `parseOptOr(rawJsonText: String)(alternative: =>Option[JValue])` or something of the sort.

Shadowfiend:
I think this is definitely good to start with, though I already know I'd probably want to customize that behavior. Logging doesn't tell the user anything, after all :)

Shadowfiend:
Any chance we can merge master in here and get a good CI build?

Shadowfiend:
The answer, apparently, was “yes, we can merge master, no, we can't get a good CI build” -.-

Anyway, :+1: let's get 'er goin'.

PR #1706, opened 2015-05-30
Optimize JsonAST.appendEscapedString character insertion

chriswebster:
Submitting a pull request based on mailing list discussion https://groups.google.com/forum/#!topic/liftweb/PD5GJYCBmx4 . 

https://github.com/lift/framework/blob/master/contributors.md already contains the appropriate info.

### DIFF
--- a/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
+++ b/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
@@ -852,7 +852,7 @@ object JsonAST {
   private def appendEscapedString(buf: StringBuilder, s: String) {
     for (i <- 0 until s.length) {
       val c = s.charAt(i)
-      buf.append(c match {
+      val strReplacement = c match {
         case '"'  => "\\\""
         case '\\' => "\\\\"
         case '\b' => "\\b"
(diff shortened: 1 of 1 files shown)


farmdawgnation:
Can you please add yourself to the contributor file for me per the contributing guidelines? :)

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -861,8 +861,9 @@ object JsonAST {
         case '\r' => "\\r"
         case '\t' => "\\t"
         case c if ((c >= '\u0000' && c < '\u0020')) => "\\u%04x".format(c: Int)
-        case c => c
-      })
+        case c => ""
+      }
+      if (strReplacement.isEmpty) buf.append(c) else buf.append(strReplacement)
```
farmdawgnation: Why do we need the if-statement? Is this part of ensuring the correct `append` is invoked?

chriswebster:
I am currently on line 23 of the contributor agreement . https://github.com/lift/framework/blob/master/contributors.md#name-23

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
chriswebster: The `isEmpty` branch allows the character to be inserted directly instead of getting converted to a string, which incurs the additional overhead.

Shadowfiend:
I'm going to fix the formatting on the if statement and add a comment indicating what it's doing there, and then merge manually.

PR #1707, opened 2015-06-13
SBT Fixes

farmdawgnation:
Looks like SBT changed some things about the way you have to download the tool. I've updated the liftsh shell script to compensate for it. This should get things in travis passable again. I also added a comment to the top of `liftsh.cmd` explaining that it's probably broken. Going to shoot something out to the mailing list and see if I can get some love on getting that updated by a Windows user.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,9 @@ services:
 jdk:
   - oraclejdk7
   - openjdk7
+branches:
+  only:
+    - master
+    - lift_26
 
 script: ./liftsh test
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,5 @@ services:
 jdk:
   - oraclejdk7
   - openjdk7
-branches:
-  only:
-    - master
-    - lift_26
 
 script: ./liftsh test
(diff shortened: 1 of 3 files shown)


Shadowfiend:
🎉 and also ew at the change. I'm down when Travis says yay 👍 (where yay is really just running anything—so intermittent failures aside).

farmdawgnation:
Also fixing the error in Travis instructions that caused *two* builds for each PR, which was unneeded work.

Shadowfiend:
Hopefully you kept the per-branch one, not per-pr? Seems extra useful >_>

Shadowfiend:
Ah I see this was a code change, sorry. Yeah I think this is opposite of what we want tbh. I like that I don't have to open a pr to get build feedback...

farmdawgnation:
> Hopefully you kept the per-branch one, not per-pr?

Eh? I don't understand...

farmdawgnation:
> Ah I see this was a code change, sorry. Yeah I think this is opposite of what we want tbh. I like that I don't have to open a pr to get build feedback...

I actually hate the fact that I'll get build feedback before having a PR. Heh. :/ "Yes, I pushed it but that doesn't mean the damn thing is ready. Leave me alone." XD

farmdawgnation:
Grrr. So we either accept duplicate builds or we don't get builds for third-party PRs. Epic fail. Whatever...

Shadowfiend:
Let's do it. :+1:

PR #1709, opened 2015-07-03
Add LiftSession.onFunctionOwnersRemoved.

Shadowfiend:
These functions are invoked with the set of function owners that had
all of their remaining associated functions evicted from a session in a
given run. A function owner is typically a page’s RenderVersion, though
there are cases (e.g., comets) where this is not necessarily the case.

@andreak have a first look and see what you think!

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,2 +1,2 @@
 # Deprecate using build.properties, use -Dsbt.version=... in launcher arg instead
-sbt.version=0.13.5
+sbt.version=0.13.8
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
@@ -70,6 +70,17 @@ object LiftSession {
   var onSetupSession: List[LiftSession => Unit] = Nil
 
   /**
+   * Holds user's functions that will be called when a function owner is
+   * no longer has functions registered for callbacks.
+   *
+   * Note that owners are '''typically''', though not always, the
+   * [[S.renderVersion render version]] of a page, so this is a good way
+   * to find out when a page no longer has any client-side functions
+   * (e.g., form fields) bound on the server.
+   */
+  var onFunctionOwnersRemoved: List[Set[String] => Unit] = Nil
+
+  /**
    * Holds user's functions that will be called when the session is about to be terminated
    */
   var onAboutToShutdownSession: List[LiftSession => Unit] = Nil
@@ -354,6 +365,7 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
   private val fullPageLoad = new BooleanThreadGlobal
 
   private val nmessageCallback: ConcurrentHashMap[String, S.AFuncHolder] = new ConcurrentHashMap
+  private val functionOwnerRemovalListeners = LiftSession.onFunctionOwnersRemoved
 
   @volatile private[http] var notices: Seq[(NoticeType.Value, NodeSeq, Box[String])] = Nil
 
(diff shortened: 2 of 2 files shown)


andreak:
Cool!
I'll have a look at it tomorrow, it's getting late here...

andreak:
My thoughts was that one could, in a Snippet or comet, register a function for the given S.renderVersion, kind of like this:
```
S.registerFunctionOwnersRemoved((renderVerson: String) => {
	println(s"Removed $renderVerson and all references in database")
})
```
and that S.registerFunctionOwnersRemoved used S.renderVersion behind the scenes and sent that as the argument in to the registered function.

When the function is executed it must also be removed from the list of registered functions, or else we accumulate lots of these and leak memory.

I see that onFunctionOwnersRemoved has the following signature:
var onFunctionOwnersRemoved: List[Set[String] => Unit] = Nil

Can you elaborate on how you think this should be used?

Thanks!

Shadowfiend:
This is basically a lower-level version of that, and leaves management of which render versions to watch to the implementing function.

Basic use:

```scala
// Boot.scala
LiftSession.onFunctionOwnersRemoved ::= { removedOwners =>
  println(removedOwners) // prints all owners who were removed for any session, when it happens
}
```

The key here is that not all owners are render versions, and some owners have session lifetime. For example, inside a comet, function ownership is a little stranger, from my brief investigation while I did this work. Functions bound in comets may have session lifetime unless you explicitly instruct otherwise, if I'm not mistaken.

The most straightforward way of achieving the thing you want to with the above API is:

```scala
object pageIdWatchers extends java.util.concurrent.ConcurrentHashMap[String,()=>Unit]

// Boot.scala
LiftSession.onFunctionOwnersRemoved ::= { removedOwners =>
  for {
    owner <- removedOwners
    watcher <- Option(pageIdWatchers.get(owner))
  } yield {
    watcher()
  }
}

// ...
// snippet
def cleanupPage = {
  cachedThingies.foreach(_.remove)
}
pageIdWatchers.put(S.renderVersion, cleanupPage)
```

andreak:
Cool, I'll try it out later.

andreak:
This is what I ended up with:

```scala
object pageIdWatchers extends java.util.concurrent.ConcurrentHashMap[String,List[()=>Unit]] {
	def addCleanupFunc(func: () => Unit): Unit = {
		put(S.renderVersion, List(func) ::: pageIdWatchers.getOrDefault(S.renderVersion, Nil))
	}
}
```
In Boot:

```scala
LiftSession.onFunctionOwnersRemoved ::= { removedOwners =>
	for {
		owner <- removedOwners
		watcher <- Option(pageIdWatchers.get(owner))} {
		watcher.foreach(_())
		pageIdWatchers.remove(owner)
	}
}
```

In my snippet:
```scala
def cleanupPage(): Unit = {
	println(s"Running cleanupPage for ${S.renderVersion}.")
}
println(s"Adding cleanupFunc for ${S.renderVersion}")
pageIdWatchers.addCleanupFunc(cleanupPage)
```

AFAICS this fully solves my use-case, thanks!

Shadowfiend:
A piece of advice to that: I would make `addCleanupFunc` wrap the passed function in a function runs it then removes it from the map, that way the caller doesn't need to worry about it.

That said, rock on! Another thing worth noting: I'm not sure what `S.renderVersion` does in weird cases (e.g., no `S` context, other contexts, etc).

We'll give this a couple of days for feedback and then merge it into master, I think!

andreak:
As a developer might call ```pageIdWatchers.addCleanupFunc``` multiple times in a snippet there might be a list of functions registered as callbacks for the given ownerID, hence the signature ```ConcurrentHashMap[String,List[()=>Unit]]```. Removing a function from this list for the ownerID-key when the function is finished and testing if there are any functions left for then removing the key introduces unnecessary complexity in ```addCleanupFunc``` I think. The implementation of ```onFunctionOwnersRemoved``` in Boot removes the ownerID-key, removing all the functions after finished executing. I feel that the burden of removing the funcs is sufficiently removed from the user this way.

In Comet S.renderVersion returns ```nextFuncName``` which is useless in this context. What is the ```removedOwners``` List[String] for functions registered in Comet-context, and how to I obtain them in ```addCleanupFunc```?

andreak:
After testing more from CometActors, which doesn't bind AJAX-funcs using S.renderVersion, I found out that the AJAX-funcs' owner is the CometActor.uniqueId, so this is a better suited signature for ```addCleanupFunc```:

```scala
object pageIdWatchers extends java.util.concurrent.ConcurrentHashMap[String,List[()=>Unit]] {
	def addCleanupFunc(func: () => Unit): Unit = {
		S.currentCometActor.map(_.uniqueId).orElse(S.request.map(_ => S.renderVersion)).foreach(functionOwner =>
			put(functionOwner, List(func) ::: pageIdWatchers.getOrDefault(functionOwner, Nil))
		)
	}
}
```

Shadowfiend:
Sweet, sounds good. I'm going to put a note on the relevant ML thread about all of this, it'd be cool if you mentioned your implementation of `pageIdWatchers` and use thereof on that thread as well so there's a reference for a possible use.

andreak:
Thanks, I will.

andreak:
Merge-time?

Shadowfiend:
Yep yep, thanks for the reminder :)

PR #1710, opened 2015-07-10
Eliminate the need for Document during JSON Serialization

### LINKED ISSUE #1681: Investigate a better printer to replace the deprecated Document
Document is deprecated. Let's take a look at using a printer more geared toward JSON itself to replace it.

farmdawgnation:
**This is a work-in-progress branch.**

This branch is focused on hitting two birds with one stone:

1. Removing the use of the deprecated `scala.text.Document` intermediary format for JSON.
2. Providing more flexible JSON rendering settings that permit options like having spaces between field names and values in JSON objects.

We already put in a good bit of work to implement `compactRender` in the `JsonAST`, and with some trivial changes it was possible to implement a `prettyRender` that uses the same pipeline with different options. My hypothesis is that by eliminating the need for an intermediary format we'll see similar performance gains during pretty rendering that we did for compact rendering using `compactRender`.

I'm opening this early so people can start taking a look if they are interested, but there's still a lot I need to do here before I consider this ready to ship. Roughly in this order:

- [x] Quantify the performance impact of these changes.
- [x] Open a thread on the ML to discuss the removal of any intermediary format, and solicit feedback on the performance impact. [See the thread.](https://groups.google.com/d/msg/liftweb/UGwy7T9-Y6s/yDUOdBuSKxkJ)
- [x] Figure out a good way to write JSON to a `Writer`.
- [x] Figure out the upgrade path from 2.6, which will likely include adding back some code I've removed with deprecation warnings.
- [x] File a ticket for Lift 3.1 to rip out the things we mark as deprecated here.
- [x] Get tests working again.

More to come!

closes #1681

### DIFF
--- a/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
+++ b/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
@@ -932,19 +932,36 @@ object JsonAST {
   private def bufRenderObj(fields: List[JField], buf: StringBuilder, settings: RenderSettings, indentLevel: Int): StringBuilder = {
     buf.append("{") //open bracket
 
+    val currentIndent = indentLevel + settings.indent
+
     if (! fields.isEmpty) {
+      if (settings.lineBreaks_?)
+        buf.append("\n")
+
       fields.foreach {
         case JField(name, value) if value != JNothing =>
+          (0 until currentIndent).foreach(_ => buf.append(" "))
+
           bufQuote(name, buf)
           buf.append(":")
-          bufRender(value, buf, settings, indentLevel)
+
+          bufRender(value, buf, settings, currentIndent)
           buf.append(",")
+
+          if (settings.lineBreaks_?)
+            buf.append("\n")
 
         case _ => // omit fields with value of JNothing
       }
 
-      if (buf.last == ',')
-        buf.deleteCharAt(buf.length - 1) //delete last comma
+      // Detect and eliminate dangling commas.
+      if (! settings.lineBreaks_? && buf.last == ',') {
+        buf.deleteCharAt(buf.length - 1)
+      } else if (settings.lineBreaks_? && buf.charAt(buf.length - 2) == ',') {
+        buf.deleteCharAt(buf.length - 2)
+      }
+
+      (0 until indentLevel).foreach(_ => buf.append(" "))
     }
 
     buf.append("}") //close bracket
--- a/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
+++ b/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
@@ -878,28 +878,36 @@ object JsonAST {
     case JNothing      => sys.error("can't render 'nothing'") //TODO: this should not throw an exception
   }
 
-  private def bufRenderArr(xs: List[JValue], buf: StringBuilder, settings: RenderSettings, indentLevel: Int): StringBuilder = {
+  private def bufRenderArr(values: List[JValue], buf: StringBuilder, settings: RenderSettings, indentLevel: Int): StringBuilder = {
     buf.append("[") //open array
-    if (!xs.isEmpty) {
-      xs.foreach(elem => Option(elem) match {
-        case Some(e) =>
-          if (e != JNothing) {
-            bufRender(e, buf, settings, indentLevel)
-            buf.append(",")
-          }
-        case None => buf.append("null,")
-      })
+
+    if (! values.isEmpty) {
+      values.foreach { elem =>
+        Option(elem) match {
+          case Some(elem) =>
+            if (elem != JNothing) {
+              bufRender(elem, buf, settings, indentLevel)
+              buf.append(",")
+            }
+
+          case None =>
+            buf.append("null,")
+        }
+      }
+
       if (buf.last == ',')
         buf.deleteCharAt(buf.length - 1) //delete last comma
     }
+
     buf.append("]")
     buf
   }
 
-  private def bufRenderObj(xs: List[JField], buf: StringBuilder, settings: RenderSettings, indentLevel: Int): StringBuilder = {
+  private def bufRenderObj(fields: List[JField], buf: StringBuilder, settings: RenderSettings, indentLevel: Int): StringBuilder = {
     buf.append("{") //open bracket
-    if (!xs.isEmpty) {
-      xs.foreach {
+
+    if (! fields.isEmpty) {
+      fields.foreach {
         case JField(name, value) if value != JNothing =>
           bufQuote(name, buf)
           buf.append(":")
@@ -908,9 +916,11 @@ object JsonAST {
 
         case _ => // omit fields with value of JNothing
       }
+
       if (buf.last == ',')
         buf.deleteCharAt(buf.length - 1) //delete last comma
     }
+
     buf.append("}") //close bracket
     buf
   }
--- a/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
+++ b/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
@@ -889,9 +889,10 @@ object JsonAST {
   }
 
   private def bufRenderArr(values: List[JValue], buf: StringBuilder, settings: RenderSettings, indentLevel: Int): StringBuilder = {
-    buf.append("[") //open array
-
+    var firstEntry = true
     val currentIndent = indentLevel + settings.indent
+
+    buf.append("[") //open array
 
     if (! values.isEmpty) {
       if (settings.lineBreaks_?)
@@ -900,33 +901,38 @@ object JsonAST {
       values.foreach { elem =>
         Option(elem) match {
           case Some(elem) if elem != JNothing =>
+            if (firstEntry) {
+              firstEntry = false
+            } else {
+              buf.append(",")
+
+              if (settings.lineBreaks_?)
+                buf.append("\n")
+            }
+
             (0 until currentIndent).foreach(_ => buf.append(" "))
-
             bufRender(elem, buf, settings, currentIndent)
-            buf.append(",")
-
-            if (settings.lineBreaks_?)
-              buf.append("\n")
 
           case Some(_) =>
             // Ignore JNothing.
 
           case None =>
+            if (firstEntry) {
+              firstEntry = false
+            } else {
+              buf.append(",")
+
+              if (settings.lineBreaks_?)
+                buf.append("\n")
+            }
+
             (0 until currentIndent).foreach(_ => buf.append(" "))
-
-            buf.append("null,")
-
-            if (settings.lineBreaks_?)
-              buf.append("\n")
+            buf.append("null")
         }
       }
 
-      // Detect and eliminate dangling commas.
-      if (! settings.lineBreaks_? && buf.last == ',') {
-        buf.deleteCharAt(buf.length - 1)
-      } else if (settings.lineBreaks_? && buf.charAt(buf.length - 2) == ',') {
-        buf.deleteCharAt(buf.length - 2)
-      }
+      if (settings.lineBreaks_?)
+        buf.append("\n")
 
       (0 until indentLevel).foreach(_ => buf.append(" "))
     }
--- a/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
+++ b/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
@@ -854,7 +854,7 @@ object JsonAST {
   }
 
   def prettyRender(value: JValue): String = {
-    bufRender(value, new StringBuilder, RenderSettings.pretty).toString()
+    render(value, RenderSettings.pretty)
   }
 
   /** Renders JSON directly to string in compact format.
@@ -862,7 +862,11 @@ object JsonAST {
     * when the intermediate Document is not needed.
     */
   def compactRender(value: JValue): String = {
-    bufRender(value, new StringBuilder, RenderSettings.compact).toString()
+    render(value, RenderSettings.compact)
+  }
+
+  def render(value: JValue, settings: RenderSettings): String = {
+    bufRender(value, new StringBuilder, settings).toString()
   }
 
   /**
--- a/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
+++ b/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
@@ -898,6 +898,9 @@ object JsonAST {
     bufRender(value, appendContainer, settings).toString()
   }
 
+  case class RenderIntermediaryDocument(value: JValue)
+  def render(value: JValue) = RenderIntermediaryDocument(value)
+
   /**
    *
    * @param value the JSON to render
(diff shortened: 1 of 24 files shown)


farmdawgnation:
### My Rudamentary benchmarking

I developed a [rudimentary benchmark](https://github.com/farmdawgnation/rudimentary-json-benchmarking) that takes a 1MB JSON file and parses it into a `JValue`. Then we compute the average amount of time over 100 runs it takes to serialize that `JValue` to a `String` in milliseconds.

**Lift 2.6.2**

```
Running test with compactRender...
29.28 ms average over 100 runs

Running test with pretty(render(...))...
47.56 ms average over 100 runs

Running test with compact(render(...))...
40.73 ms average over 100 runs
```

**Lift 3 (master)**

```
Running test with compactRender...
14.18 ms average over 100 runs

Running test with pretty(render(...))...
30.29 ms average over 100 runs

Running test with compact(render(...))...
27.16 ms average over 100 runs
```

**Lift 3 (this branch)**

```
Running test with compactRender...
15.63 ms average over 100 runs

Running test with prettyRender...
18.0 ms average over 100 runs
```

### Chris Webster's Performance Benchmark

After seeing solid results with the above benchmark I altered the [lift-jsonBenchmark](https://github.com/chriswebster/lift-jsonBenchmark) developed by Chris to run against master and then this branch to see what sort of results that yielded. Those are summarized below.

**Lift 3 (master)**

```
[info] Result "serializeCaseClass":
[info]   548707.039 ±(99.9%) 8479.064 ops/s [Average]
[info]   (min, avg, max) = (457788.399, 548707.039, 589114.966), stdev = 35900.888
[info]   CI (99.9%): [540227.975, 557186.103] (assumes normal distribution)
[info]
[info]
[info] # Run complete. Total time: 00:06:44
[info]
[info] Benchmark                              Mode  Cnt       Score      Error  Units
[info] LiftJsonBenchmark.serializeCaseClass  thrpt  200  548707.039 ± 8479.064  ops/s
```

**Lift 3 (this branch)**

```
[info] Result "serializeCaseClass":
[info]   511882.150 ±(99.9%) 11243.983 ops/s [Average]
[info]   (min, avg, max) = (400837.177, 511882.150, 558733.741), stdev = 47607.729
[info]   CI (99.9%): [500638.167, 523126.133] (assumes normal distribution)
[info]
[info]
[info] # Run complete. Total time: 00:06:44
[info]
[info] Benchmark                              Mode  Cnt       Score       Error  Units
[info] LiftJsonBenchmark.serializeCaseClass  thrpt  200  511882.150 ± 11243.983  ops/s
```

farmdawgnation:
So somehow my changes caused this to start happening:

```
[error] /Users/matt/Sites/framework/core/json/src/test/scala/net/liftweb/json/Examples.scala:217: value ~ is not a member of (String, Null)
[error]   val nulls = (("f1" -> null) ~ ("f2" -> List(null, "s")))
[error]                               ^
[error] one error found
[error] (lift-json/test:compileIncremental) Compilation failed
[error] Total time: 6 s, completed Jul 24, 2015 11:31:17 AM
```

And I'm not really sure how...

farmdawgnation:
Particularly somehow b1f386c is the culprit.

I don't even..

farmdawgnation:
So somehow that `null` had a type other than `Null` in the previous code. O_o

Either way, appears to be working now. So yay.

farmdawgnation:
So, I don't know why `TimeHelpersSpec` refuses to pass... but it works locally...

Shadowfiend:
Ugh. The TimeHelpers thing is a recurring weirdness with the Iran timezone. One of these days I guess I'm going to have to step through various system times and find out what's going on.

Shadowfiend:
Wait, why was the ultimate fix an explicit cast to `String`, when just dropping the wrapping parens seems like it would also have fixed the issue without any casting?

farmdawgnation:
> Wait, why was the ultimate fix an explicit cast to String, when just dropping the wrapping parens seems like it would also have fixed the issue without any casting?

What wrapping parens?

Shadowfiend:
The commit you pointed out has this change: https://github.com/lift/framework/commit/b1f386c5393462bbd280409fe0dcf03ab3bc1973#diff-40306c0c73e4860cdff0c3735f497919R217 .

Presumably that changed the way the line is typed, leading to the issue you saw.

farmdawgnation:
Ah no, the parens were added *in response to* the issue. They shouldn't have ended up in a commit, but removing them didn't fix it for me.

Shadowfiend:
Further investigation indicates that it was a seemingly unrelated invocation of `render(personDSL)` elsewhere in the file that was causing things to be typed correctly before. Additionally, annotating all of the `val`s in `Examples` as `JValue` makes things compile, but causes bizarre extraction failures.

farmdawgnation:
Alright, well, it looks like we've got no objections to these changes on the ML after giving it a pretty good amount of time for feedback. So, I'm removing the [WIP] designation from this branch. Let's :shipit:.

farmdawgnation:
@Shadowfiend I'm not hearing many objections to this work. Do you (or someone else) want to do a line review with me?

core/json/src/test/scala/net/liftweb/json/JsonPrintingSpec.scala
```diff
@@ -29,14 +29,12 @@ import org.scalacheck.Prop.forAll
...
   "rendering does not change semantics" in {
-    val rendering = (json: Document) => parse(Printer.pretty(json)) == parse(Printer.compact(json))
+    val rendering = (json: JValue) => parse(JsonAST.prettyRender(json)) == parse(JsonAST.compactRender(json))
     check(forAll(rendering))
   }
 
   private def parse(json: String) = scala.util.parsing.json.JSON.parseRaw(json)
 
-  implicit def arbDoc: Arbitrary[Document] = Arbitrary(genJValue.map(render(_)))
+  implicit def arbDoc: Arbitrary[JValue] = Arbitrary(genJValue)
```
Shadowfiend: Is this still needed?

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -224,14 +225,19 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
...
-  val jsonOutputConverter = new FactoryMaker[scala.text.Document => String]({
-    import json.Printer
-    if (Props.devMode) Printer.pretty _ else Printer.compact _}){}
+  val jsonOutputConverter = new FactoryMaker[JsonAST.JValue => String]({
+    import json.{prettyRender, compactRender}
+
+    if (Props.devMode)
+      prettyRender _
+    else
+      compactRender _
```
Shadowfiend: Some if curlies missing here.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -896,52 +915,93 @@ object JsonAST {
...
+
+      values.foreach { elem =>
+        if (elem != JNothing) {
+          if (firstEntry) {
+            firstEntry = false
+          } else {
+            buf.append(',')
+
+            if (settings.lineBreaks_?)
+              buf.append('\n')
```
Shadowfiend: if curlies

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -896,52 +915,93 @@ object JsonAST {
...
-      if (buf.last == ',')
-        buf.deleteCharAt(buf.length - 1) //delete last comma
+
+          (0 until currentIndent).foreach(_ => buf.append(' '))
+          bufRender(elem, buf, settings, currentIndent)
+        }
+      }
+
+      if (settings.lineBreaks_?)
+        buf.append('\n')
```
Shadowfiend: if curlies

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -896,52 +915,93 @@ object JsonAST {
...
+
+      fields.foreach {
         case JField(name, value) if value != JNothing =>
+          if (firstEntry) {
+            firstEntry = false
+          } else {
+            buf.append(',')
+
+            if (settings.lineBreaks_?)
+              buf.append('\n')
```
Shadowfiend: if curlies

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -896,52 +915,93 @@ object JsonAST {
...
+
+          (0 until currentIndent).foreach(_ => buf.append(' '))
+
           bufQuote(name, buf)
-          buf.append(":")
-          bufRender(value, buf)
-          buf.append(",")
+          buf.append(':')
+          if (settings.spaceAfterFieldName)
+            buf.append(' ')
```
Shadowfiend: if curlies

Shadowfiend: Few more throughout.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -874,20 +843,70 @@ object JsonAST {
     }
   }
 
+  sealed trait AppendContainer[T] {
```
Shadowfiend: Not sure I like this name. Why not just name it `StringAppendable` or `UnifiedAppendable`? Either way, a scaladoc indicating why it's here would be handy.

Also wondering if it wouldn't make sense to make these value classes… Not sure we'd gain an advantage here since we treat them as their trait, though.

Actually, now that I think about it, is there a strong reason why we're using Scala's `StringBuilder` instead of Java's which implements `Appendable`?

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -874,20 +843,70 @@ object JsonAST {
...
+  case class AppendableAppendContainer(appendable: Appendable) extends AppendContainer[Appendable] {
+    override def append(thing: Char) = AppendableAppendContainer(appendable.append(thing))
+    override def append(thing: String) = AppendableAppendContainer(appendable.append(thing))
+    override def toString = appendable.toString
+  }
+
+  object RenderSettings {
+    val pretty = RenderSettings(2)
+    val compact = RenderSettings(0)
+  }
```
Shadowfiend: Personal preference: I'd put the singleton after the definition of `RenderSettings`.

Shadowfiend: Also not really something for this PR, but wondering if this whole shebang should mostly be moved to `Serialization` and referenced from here for convenience… Just a random thought stealing through my brain though.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
farmdawgnation: Yeah, performance on Java's was abysmal for some reason. :/

Shadowfiend: Just to make sure, did you try `StringBuilder` or `StringBuffer`? One of the two is synchronized and therefore slower. In particular, a quick peek at Scala's `StringBuilder` indicates it uses the [Java one](https://github.com/scala/scala/blob/v2.10.4/src/library/scala/collection/mutable/StringBuilder.scala#L28) [under the covers](https://github.com/scala/scala/blob/v2.10.4/src/library/scala/collection/mutable/StringBuilder.scala#L196-L199) anyway.

core/json/src/test/scala/net/liftweb/json/JsonPrintingSpec.scala
farmdawgnation: I think so, but admittedly the changes I made here required so much refactoring I didn't scrutinize them too much. Refactoring these tests would be A Good Thing™ - but given the breadth of this change I'd like to avoid doing so now. It somewhat invalidates the usefulness of the tests toward checking the branch if we go changing them in any meaningful way in the branch itself.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
farmdawgnation: I'm about 99% sure it was `StringBuilder` because the name is the same.

core/json/src/test/scala/net/liftweb/json/JsonPrintingSpec.scala
Shadowfiend: Oh, this isn't so much a question of refactoring but rather whether your changes made it so we're not using the `implicit` at all.

I dig the “enough changes for now” tack, though!

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
farmdawgnation: Eh, singleton / case class ordering is six one way / half dozen the other as long as they're next to each other.

And yeah, probably a good candidate for moving soon.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
Shadowfiend: I'm going to try and make some time to try again myself this week and see if I observe the same differences (unless you beat me to it)—since Scala's `StringBuilder` uses Java's behind the scenes, it seems more likely that the difference in performance was a fluke, and this has added enough abstraction weirdness that it'd be ideal if we could avoid it.

core/json/src/test/scala/net/liftweb/json/JsonPrintingSpec.scala
farmdawgnation: Ah, no I do think it's still in use. I think the `forAll` construct from scalacheck is using it.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
farmdawgnation: Make sense.

core/json/src/test/scala/net/liftweb/json/JsonPrintingSpec.scala
Shadowfiend: Got it. :thumbsup:

farmdawgnation:
I think all the concerns have been addressed.

farmdawgnation:
@Shadowfiend Have you had an opportunity to check out the StringBuilder performance?

Shadowfiend:
Just ran the tests with a hacked up version using `java.lang.StringBuilder`. Performance was marginally better:

 - `master`: [info] LiftJsonBenchmark.serializeCaseClass  thrpt  200  **463112.998 ± 5765.477  ops/s**
 - current: [info] LiftJsonBenchmark.serializeCaseClass  thrpt  200  **454022.779 ± 4450.973  ops/s**
 - `java.lang.StringBuilder`: [info] LiftJsonBenchmark.serializeCaseClass  thrpt  200  **521409.948 ± 3920.004  ops/s**

I think it's worth making the change and dropping the intermediate classes/traits.

farmdawgnation:
Alright, I'll get it done.

farmdawgnation:
Try that on for size.

Shadowfiend:
26-if-it-fits-it-ships-meme

andreak:
Just curious; Why not Use Scala's StringBuilder directly (StringBuilder.newBilder)?

Shadowfiend:
Scala's `StringBuilder` isn't `Appendable`, which means we'd only support writing to a string, not to any arbitrary `Appendable`. The current implementation lets you, for example, serialize directly into a `Writer`, and through it into an `OutputStream`, without having to go to `String` first. Prior to @farmdawgnation's latest commit, we supported both, but to do that we had to set up a contorted class that bridged between Scala's types and Java's `Appendable`. This seems cleaner, since Java's `Appendable` is more broadly compatible with other APIs, and seems to also be ~10-15% faster performance-wise.

Shadowfiend:
(Oh, notably, I believe it's also the most compatible with the previous version of lift-json's rendering, which handled only Java `Writer`s but is now generalized to all `Appendable`s.)

andreak:
Thanks

Shadowfiend:
Leet's run it! @farmdawgnation whenever you get a chance I guess let the list know we're in!

PR #1711, opened 2015-07-13
High Wire: Fix appendJs when invoked from inside a toJsCmd def

### LINKED ISSUE #1708: Wiring-contents not updated correctly if response is generated from a JsCmd where toJsCmd is a def
Ref: https://groups.google.com/d/msg/liftweb/4pPhUMqQ1vw/rGnFHXUuGNkJ
A minimal project is here which exposes the issue: https://github.com/andreak/wiring-bug

Shadowfiend:
This fixes #1708 specifically.

`appendJs` should append JavaScript to be sent down with the
current request. However, if `appendJs` is called in a `toJsCmd`
method, and that method is defined as a `def`, not a `val`, then
`toJsCmd` will be called *after* the `jsToAppend` is read. This
means that this appended JS won't be sent down.

One place where this manifests is when a `JsCmd`'s `toJsCmd`
method sends down a rendered snippet that contains a Wiring
component (because Wiring uses `appendJs`). In this case, the
`appendJs` that Wiring does is evaluated during the `toJsCmd`
invocation, which means the JsCmd that is sent down misses it.

When we send down JS in bulk for an AJAX or comet response,
we wrap it in a `JsCommands`, which is a collector of `JsCmd`.
We then call its `toResponse` to get a `JavaScriptResponse`.

Before, we passed the `jsToAppend` to `JsCommands`, and then
let `JsCommands` run `toJsCmd` on everyone while constructing
the response. In this PR, let the `toResponse` method call `toJsCmd`
on all the commands that are explicitly being sent down, and then
we read `jsToAppend` and concatenate that on the end of the other
commands inside `toResponse`. In the process, we remove external
invocations of `jsToAppend`.

So as to avoid accidental duplicate send-downs of `jsToAppend`, we
give `jsToAppend` a `clearAfterReading` boolean that, when set to
`true` (it defaults to `false`), will clear the list of `JsCmd`s that need
to be appended. `JsCommands` and the `jsToAppend` reading in
`CometActor`, which sends down additional JS as a partial update,
both set it to `true`. Amongst other things, this ensures the two won't
both try to send down the same JS.

Still pending here: I (or someone else) need to test some of the
scenarios of `appendJs`, like calling it from inside a comet, from
a regular AJAX request, and from a regular screen render.

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/CometActor.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/CometActor.scala
@@ -719,7 +719,7 @@ trait BaseCometActor extends LiftActor with LiftCometActor with CssBindImplicits
                   case e: Exception => reportError("Message dispatch for " + in, e)
                 }
 
-                val updatedJs = S.jsToAppend
+                val updatedJs = S.jsToAppend(clearAfterReading = true)
                 if (updatedJs.nonEmpty) {
                   partialUpdate(updatedJs)
                 }
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftMerge.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftMerge.scala
@@ -74,7 +74,7 @@ private[http] trait LiftMerge {
       LiftRules.javaScriptSettings.vend().map { settingsFn =>
         LiftJavaScript.initCmd(settingsFn(this))
       }.toList ++
-      S.jsToAppend ++
+      S.jsToAppend() ++
       eventJs
 
     allJs.foldLeft(js.JsCmds.Noop)(_ & _)
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftServlet.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftServlet.scala
@@ -632,15 +632,14 @@ class LiftServlet extends Loggable {
                 (JsCommands(S.noticesToJsCmd :: Nil) &
                   (js :: (xs.collect {
                     case js: JsCmd => js
-                  }).reverse) &
-                  S.jsToAppend
+                  }).reverse)
                 ).toResponse
               }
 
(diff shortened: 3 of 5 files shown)


andreak:
I have a rather large app I can test with. We use JsCmds in all possible ways, I think.

andreak:
I did some (quick) testing of our app now and it seems all usages of ```S.appendJS``` and ```def toJsCmd``` work as expected. The Wiring-UI issue certainly works.

I found an inssue where ```contextPath``` is not being prepended properly to attributes (href and src) when using ```val``` instead of ```def``` in ```toJsCmd``` but that's a separate issue.

andreak:
It seems the contextPath-issue only manifests it self when the JS-content is generated in another thread (job on thread-pool), not having access to Req of Session. Forget about this for now...

Shadowfiend:
Sweet! I also need to tweak up the Scaladocs (or add them in the case of `JsCommands`'s `toResponse`).

andreak:
Ping:-)

Any progress on merging this?

Shadowfiend:
Haha I was just looking at it today. Want to do that scaladoc update and then merge… Tomorrow night, probably, if that doesn't seem too long.

andreak:
Rock on!

Shadowfiend:
Looks like both failures were the typical failures with futures…

@farmdawgnation you interested in :shipit: ?

farmdawgnation:
:eyes:

Shadowfiend:
@farmdawgnation Thinkin' 'bout merging this tomorrow come hell or high water; lemme know if you spotted anything problematic :)

farmdawgnation:
I didn't. I intended to build a sample project with it and try to break it but got distracted by other things. :(

Shadowfiend:
No worries. Let's ship it since Andreas confirmed it works on their existing codebase.

andreak:
Yahoo!

PR #1714, opened 2015-07-24
Loc loco: Fix behavior of MatchWithoutCurrentValue when stars aren't present in path.

### LINKED ISSUE #1669: MatchWithoutCurrentValue doesn't work for Menu paths that don't have a * in them.
For some reason, if you don't include a * in your SiteMap entry, then MatchWithoutCurrentValue just won't work. So, for example, the following won't currently work as expected:

```scala
    // /product/1 produces product info
    // /product/2 should take us to the home page
    lazy val product = Menu.param[ProductInfo]("product", "Product Detail Page",
      id   => lookup(id),
      info => info.id.toString
      ) / "product" >> MatchWithoutCurrentValue >> FallbackToSearchPage
```

There's probably another codepath somewhere that doesn't check for MatchWithoutCurrentValue before bailing if you don't have a * in the path.

Relevant mailing list post [here](https://groups.google.com/d/msg/liftweb/524VF6BD2fs/vkpY4JG-Kh0J).

farmdawgnation:
This PR closes #1669 as it resolves the issue whereby given a menu declared like so:

```scala
    lazy val product = Menu.param[ProductInfo]("product", "Product Detail Page",
      id   => lookup(id),
      info => info.id.toString
      ) / "product" >> MatchWithoutCurrentValue >> FallbackToSearchPage
```

The `MatchWithoutCurrentValue` would not behave as expected because we hit a different processing pipeline due the absence of the star. Turns out that when we're working with `Menu`'s that lack a star they create a `RewriteResponse` internally. However, the code that creates that `RewriteResponse` is explicitly running the lookup function and failing to issue the rewrite if the function returns anything other than `Full`.

With these changes, we've resolved that issue. Now, `LocRewrite`s can be issued even when `T` can't be resolved, meaning the above code will work as expected.

Hat tip to @d6y for finding the original bug.

Some TODOs:

- [x] Figure out a way to test this thing.
- [x] Figure out what (if anything) we want to do for Lift 2.6. (though I suppose that doesn't block this getting merged into master)

### DIFF
--- a/persistence/proto/src/main/scala/net/liftweb/proto/Crudify.scala
+++ b/persistence/proto/src/main/scala/net/liftweb/proto/Crudify.scala
@@ -239,8 +239,7 @@ trait Crudify {
         override val rewrite: LocRewrite =
         Full(NamedPF(name) {
             case RewriteRequest(pp , _, _) if hasParamFor(pp, viewPath) =>
-              (RewriteResponse(viewPath),
-               findForParam(pp.wholePath.last).openOrThrowException("legacy code, it was open_!"))
+              (RewriteResponse(viewPath), findForParam(pp.wholePath.last))
           })
 
         override def calcTemplate = Full(viewTemplate)
@@ -286,8 +285,7 @@ trait Crudify {
           override val rewrite: LocRewrite =
           Full(NamedPF(name) {
               case RewriteRequest(pp , _, _) if hasParamFor(pp, editPath) =>
-                (RewriteResponse(editPath),
-                 findForParam(pp.wholePath.last).openOrThrowException("legacy code, it was open_!"))
+                (RewriteResponse(editPath), findForParam(pp.wholePath.last))
             })
 
           override def calcTemplate = Full(editTemplate)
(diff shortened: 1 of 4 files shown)


farmdawgnation:
Per the ML thread we're punting on backporting this to 2.6.

farmdawgnation:
Alright, unfortunately I had to touch a few things in lift-proto to make everything happy – but now everything compiles. Let's see what TravisCI says, but I'm removing the "[WIP]" prefix from this branch.

Shadowfiend:
So in this new code path, if `MatchWithoutCurrentValue` isn't included, do we properly 404? Particularly thinking about the `lift-proto` stuff since it seems like it may be using it to set up default CRUD pages.

farmdawgnation:
Yes. We should properly 404 in the event that isn't tacked on.

farmdawgnation:
Tests appear to be capable of passing in spite of what the build status indicates. I'm going to leave this up for another day or so and then turbo merge it.

PR #1717, opened 2015-08-29
Special case Option intermediary type in extraction.

### LINKED ISSUE #1071: lift-json flattens option types unexpectedly
When parsing case classes with a List of Option[String], the option 
classes get flatten out. 
For case class Library(name:String, docs:List[Option[String]]) 
when I parse and extract it, i get: 

<pre><code>Library(test,List(1, 2))</code></pre>
but I expect: 
<pre><code>Library(test,List(Some(1), Some(2))) </code></pre>

The test unit is at [[url:https://gist.github.com/1077437|https://gist.github.com/1077437]] 
Is this the expected behavior or a bug in lift-json 2.4M1? 

Joni Freeman states: " extracted type must be the same type as given to  the extract function"
http://groups.google.com/group/liftweb/browse_thread/thread/fb40cea9ebecbf49#

farmdawgnation:
This PR resolves a long-standing bug where trying to execute extraction when `Option` was an intermediary type on a `List` would cause fireworks and explosions. So, for example if you tried something like the following:

```scala
val json = JsonParser.parse("""["one", "two", null]""")
json.extract[List[Option[String]]]
```

You would get explosions and fireworks because lift-json would try to match up what you've provided to the constructor for `Option`. We now special case `Option` as a type of collection in the generation of the `Meta` ADT, and handle that case when constructing the concrete instance.

I'm not quite familiar enough with this code to know if this is a kludge or a clever hack. Would love some :eyes: from @jonifreeman (if you have time). Either way, I humbly submit this for review.

closes #1071

### DIFF
--- a/core/json/src/test/scala/net/liftweb/json/ExtractionBugs.scala
+++ b/core/json/src/test/scala/net/liftweb/json/ExtractionBugs.scala
@@ -59,7 +59,7 @@ object ExtractionBugs extends Specification {
     json.extract[Response] mustEqual Response(List(Map("one" -> 1, "two" -> 2)))
   }
 
-  "Extracting List[Option[String]]" in {
+  "Extraction should handle List[Option[String]]" in {
     val json = JsonParser.parse("""["one", "two", null]""")
     json.extract[List[Option[String]]] mustEqual List(Some("one"), Some("two"), None)
   }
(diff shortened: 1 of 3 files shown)


core/json/src/test/scala/net/liftweb/json/ExtractionBugs.scala
```diff
@@ -59,6 +59,11 @@ object ExtractionBugs extends Specification {
     json.extract[Response] mustEqual Response(List(Map("one" -> 1, "two" -> 2)))
   }
 
+  "Extracting List[Option[String]]" in {
```
Shadowfiend: Phrasing here seems wrong (“Extraction should handle…” seems to be the trend above).

Shadowfiend:
Looks cool!

core/json/src/test/scala/net/liftweb/json/ExtractionBugs.scala
farmdawgnation: Ah.

PR #1718, opened 2015-08-29
Handle incorrect types wrapped in the Option.

### LINKED ISSUE #1425: Improved checking when deserializing an Option[T]
Details https://groups.google.com/d/msg/liftweb/5nMZ9hHv9Uk/Y0odnMGMJUsJ

farmdawgnation:
~~This PR depends on #1717.~~

As described in #1425, previously if we were expecting an `Option[String]` but you provided an integer value for that key in an object, for example, we would just give you a `None` in that field. This isn't exactly desirable, since you probably generated that data somewhere else (maybe not Scala) expecting something to be in that key when you pipe it into Lift and try to deserialize it. So when we give you a `None` there we're not telling you that something has gone wrong.

So, this fixes that issue by checking to see whether or not what we're looking at is a `JNothing` or `JNull`. If it is, then hey - we didn't find anything and you get a `None`. If it's anything other than that you're a silly bear and gave us the wrong thing for what you're trying to instantiate. You get a `MappingException`!

closes #1425

### DIFF
--- a/core/json/src/main/scala/net/liftweb/json/Extraction.scala
+++ b/core/json/src/main/scala/net/liftweb/json/Extraction.scala
@@ -197,11 +197,7 @@ object Extraction {
       }
     }
 
-    if (clazz == classOf[Option[_]]) {
-      json.toOpt.map(extract0(_, mkMapping(typeArgs.head, typeArgs.tail)))
-    } else {
-      extract0(json, mkMapping(clazz, typeArgs))
-    }
+    extract0(json, mkMapping(clazz, typeArgs))
   }
 
   def extract(json: JValue, target: TypeInfo)(implicit formats: Formats): Any = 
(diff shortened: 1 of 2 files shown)


farmdawgnation:
Probably going to ship this soon unless someone wants to take a closer look at it.

joescii:
Not to quibble over white space and curlies, but we should at least not bother with explicit `null` checks when creating an `Option` since that's exactly what `Option.apply()` does: https://github.com/scala/scala/blob/v2.11.7/src/library/scala/Option.scala#L25

joescii:
Oh, but :+1:

farmdawgnation:
Good call. Though in the future do you mind letting me make those kinds of changes when it's my branch unless I'm MIA on responding to the comments for a bit? :angel:

That said, great change. :100:

farmdawgnation:
I vote we ship this unless someone else has remarks.

joescii:
Sure thing. I didn't think any harm in committing an idea since it was a branch, so no problem. :)

PR #1721, opened 2015-09-07
Squashing the logError JS stackoverflow

### LINKED ISSUE #1720: SOE in lift.logError client side js.
```
>window.lift.logError("hello");

lift.js:1 Uncaught RangeError: Maximum call stack size exceeded
    at Object.window.lift.logError (http://localhost:8081/classpath/lift.js:1:8238)
    at Object.lift_settings.logError (http://localhost:8081/lift/page/F276291420657F5GK33.js:2:264)
    at Object.window.lift.logError (http://localhost:8081/classpath/lift.js:1:8259)
    at Object.lift_settings.logError (http://localhost:8081/lift/page/F276291420657F5GK33.js:2:264)
    at Object.window.lift.logError (http://localhost:8081/classpath/lift.js:1:8259)
    at Object.lift_settings.logError (http://localhost:8081/lift/page/F276291420657F5GK33.js:2:264)
    at Object.window.lift.logError (http://localhost:8081/classpath/lift.js:1:8259)
    at Object.lift_settings.logError (http://localhost:8081/lift/page/F276291420657F5GK33.js:2:264)
    at Object.window.lift.logError (http://localhost:8081/classpath/lift.js:1:8259)
    at Object.lift_settings.logError (http://localhost:8081/lift/page/F276291420657F5GK33.js:2:264)

```

https://groups.google.com/forum/#!topic/liftweb/FKBlmGxMJZE

joescii:
Resolves #1720.

One other thought I had while under the hood of this one.  [Several other settings-based JS functions log an error that they are to be defined in settings](https://github.com/lift/framework/blob/09fe944fd314ba52447a5e85df25f12d2b130121/web/webkit/src/main/resources/toserve/lift.js#L61-L73).  We could consider doing the same for `logError` to be consistent.  I wasn't convinced this would be the right move, so I didn't bother touching it.

[Review on Reviewable](https://reviewable.io/reviews/lift/framework/1721)

### DIFF
--- a/web/webkit/src/main/resources/toserve/lift.js
+++ b/web/webkit/src/main/resources/toserve/lift.js
@@ -581,6 +581,7 @@
           doCycleIn200();
         });
       },
+      defaultLogError: function(msg) { consoleOrAlert(msg) },
       logError: function() { settings.logError.apply(this, arguments) },
       onEvent: function() { settings.onEvent.apply(this, arguments) },
       ajax: appendToQueue,
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
@@ -898,7 +898,7 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
    *   LiftRules.jsLogFunc = Full(v => JE.Call("alert",v).cmd)
    */
   @volatile var jsLogFunc: Box[JsVar => JsCmd] =
-    if (Props.devMode) Full(v => JE.Call("lift.logError", v))
+    if (Props.devMode) Full(v => JE.Call("lift.defaultLogError", v))
     else Empty
 
   /**


farmdawgnation:
I'm not so sure about this solution. This just removes the ability for the user to customize their error logging doesn't it?

joescii:
No, not at all.  I'm just giving a non-broken default.  You can still override this in boot:

```scala
LiftRules.jsLogFunc = Full(v => JE.Call("window.alert", v).cmd)
```

farmdawgnation:
Wasn't part of the point of this code on the front end that you could override it in the client too though? It looks like that's intended to be possible.

joescii:
Well it _is_ JS, so of course it can be overridden there too. ;)

But because it is a `var Box` I assumed that was the intended path to configure it. Maybe we need someone more familiar with the design of this bit like perhaps @shadowfiend. 

Joe

Shadowfiend:
Right now it looks like the customization point is `LiftRules` for these (because the initial settings for the Lift JS are generated via `LiftRules`), and I think that's okay.

joescii:
I have the hood up in this area so we can make our js more configurable. I think I'll have a better solution for this particular problem when I'm done.

joescii:
Yeah, I think my [js-config branch](https://github.com/lift/framework/tree/js-config) is much better. 

Instead of simply passing the `settings:JsObj` in `LiftJavaScript`, I am [creating it in a function](https://github.com/lift/framework/blob/1fa8d06eb7e7aa93f70f83c953c35b6d2c456118/web/webkit/src/main/scala/net/liftweb/http/js/LiftJavaScript.scala#L83-L84) which expects an object containing the defaults.  Then in `lift.js`, I am [passing it the original `settings` as the defaults](https://github.com/lift/framework/blob/1fa8d06eb7e7aa93f70f83c953c35b6d2c456118/web/webkit/src/main/resources/toserve/lift.js#L551).

This way it is possible to [use the default settings in `LiftRules`](https://github.com/lift/framework/blob/1fa8d06eb7e7aa93f70f83c953c35b6d2c456118/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala#L901).

What do y'all think? Close this PR and move forward with the new approach?

Shadowfiend:
Pop open another PR and let's have a look—I'm not 100% clear just from your message what's going on :)

joescii:
What's so hard to understand about my vague hand waving??

https://github.com/lift/framework/pull/1727

joescii:
Ok, I don't like the new idea that much after digging into it.  I see now I had a misunderstanding about how the settings were supposed to work in the first place.  Seems to me that the `LiftRules` JS functions are supposed to point at public JS functions on the `lift` object.  I suspect this particular problem with `logError` threw me off the right course a bit.

Shadowfiend:
So do we want to toss 1727 and put this in?

Shadowfiend:
I think I'm going to go ahead and merge this and we can decide on 1727 separately… It seems like this fixes the immediate issue, and we can go from there.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/lift/framework/1721#-:-K4nQVZjUt5xDhKifZ7D:815501225)*

Shadowfiend:
Sorry for the Reviewable noise there… Was trying something new out, didn't realize it would drop extra links and such.

Shadowfiend:
@joescii let me know if you'd prefer I not merge this as it currently stands; I'm rerunning the specs in the meantime to hopefully get a green light. I'll probably move on it tomorrow or the day after if I don't hear anything :)

joescii:
@Shadowfiend Sorry I've been dead silent lately. I had a trip out of town for work last week, and I've not been able to catch up.  I don't think this PR hurts anything, even though I feel it doesn't open up all of the hooks that I feel we should leave available to the end user.

Shadowfiend:
Perfect. That's pretty much where I thought it was as well, so it sounds like a good incremental step. Merging!

PR #1722, opened 2015-10-02
Props enhanced

joescii:
I haven't heard any feedback on the [ML thread] (https://groups.google.com/forum/#!searchin/liftweb/properties%7Csort:date/liftweb/5T_4xm1amr8/XaruS3VMBwAJ) regarding this enhancement, so I thought I'd take the next step to make a PR to have some more eyes on it.

Briefly, this PR allows Lift to be configured via `LiftRules` to (1) look up properties from more sources than just the `*.props` files such as JVM `System` properties, OS environment properties, or arbitrary `Map[String, String]` and (2) to reference other properties with unix-like `${propName}` syntax.  The extra property sources' precedence is well-defined and dictated by the developer via calling `prepend` vs `append`.  For interpolators we have `appendInterpolator` only since there is not a fixed resource to work around.

Note that interpolation is implemented such that infinite recursion is not possible. We do not check the values found for more keys to look up. 

The only missing feature I feel we have here is the ability to interpolate from the `*.props` files themselves. Currently we can only look up from maps provided to `appendInterpolator`.

See `net.liftweb.util.PropsSpec` for working usage examples.  Looking forward to the feedback!

### DIFF
--- a/core/util/src/main/scala/net/liftweb/util/Props.scala
+++ b/core/util/src/main/scala/net/liftweb/util/Props.scala
@@ -49,16 +49,18 @@ import common._
  * The standard Lift properties file extension is "props".
  */
 object Props extends Logger {
+  type PropProvider = scala.collection.Map[String, String]
+
   /**
    * Get the configuration property value for the specified key.
    * @param name key for the property to get
    * @return the value of the property if defined
    */
   def get(name: String): Box[String] = {
-    val tries:List[Box[Object]] =
-      prepended.map(m => Box.legacyNullTest(m get name)) ++
-      (props.get(name).toBox ::
-      appended.map(m => Box.legacyNullTest(m get name)))
+    val tries:List[Box[String]] =
+      prepended.map(m => Box(m.get(name))) ++
+      (Box(props.get(name)) ::
+      appended.map(m => Box(m.get(name))))
 
     tries.find(_.isDefined).getOrElse(Empty).map(interpolate)
   }
@@ -66,7 +68,7 @@ object Props extends Logger {
   private val interpolateRegex = """(.*?)\Q${\E(.*?)\Q}\E([^$]*)""".r
   private def interpolate(value:Object):String = {
     def lookup(key:String) = {
-      val tries = interpolators.map(m => Box.legacyNullTest(m get key))
+      val tries = interpolators.map(m => Box(m.get(key)))
       tries.find(_.isDefined).getOrElse(Empty)
     }
     val interpolated = for {
@@ -115,24 +117,21 @@ object Props extends Logger {
 
   /**
    * Updates Props to find property values in the argument BEFORE looking in the standard Lift prop files.
-   * @param props Arbitrary map of property key -> property value
+   * @param provider Arbitrary map of property key -> property value
    */
-  def prependSource(props:Map[String, String]):Unit = prependSource(props.asJava.asInstanceOf[java.util.Map[Object, Object]])
-  def prependSource(props:java.util.Map[Object, Object]):Unit = prepended = props :: prepended
+  def prependSource(provider:PropProvider):Unit = prepended = provider :: prepended
 
   /**
    * Updates Props to find property values in the argument AFTER first looking in the standard Lift prop files.
-   * @param props Arbitrary map of property key -> property value
+   * @param provider Arbitrary map of property key -> property value
    */
-  def appendSource(props:Map[String, String]):Unit = appendSource(props.asJava.asInstanceOf[java.util.Map[Object, Object]])
-  def appendSource(props:java.util.Map[Object, Object]):Unit = appended = appended :+ props
+  def appendSource(provider:PropProvider):Unit = appended = appended :+ provider
 
   /**
    * Updates Props to find interpolated values in the argument.
-   * @param props
+   * @param provider Arbitrary map of property key -> property value
    */
-  def appendInterpolator(props:Map[String, String]):Unit = appendInterpolator(props.asJava.asInstanceOf[java.util.Map[Object, Object]])
-  def appendInterpolator(props:java.util.Map[Object, Object]):Unit = interpolators = interpolators :+ props
+  def appendInterpolator(provider:PropProvider):Unit = interpolators = interpolators :+ provider
 
   /**
    * Enumeration of available run modes.
@@ -379,9 +378,9 @@ object Props extends Logger {
     }
   }
 
-  private var prepended:List[java.util.Map[Object, Object]] = Nil
-  private var appended:List[java.util.Map[Object, Object]] = Nil
-  private var interpolators:List[java.util.Map[Object, Object]] = Nil
+  private var prepended:List[PropProvider] = Nil
+  private var appended:List[PropProvider] = Nil
+  private var interpolators:List[PropProvider] = Nil
 
   /**
    * Resets the Props stacks to allow for isolated unit tests
--- a/core/util/src/main/scala/net/liftweb/util/Props.scala
+++ b/core/util/src/main/scala/net/liftweb/util/Props.scala
@@ -89,18 +89,21 @@ object Props extends Logger {
    * @param props Arbitrary map of property key -> property value
    */
   def prepend(props:Map[String, String]):Unit = {}
+  def prepend(props:java.util.Map[Object, Object]):Unit = {}
 
   /**
    * Updates Props to find property values in the argument AFTER first looking in the standard Lift prop files.
    * @param props Arbitrary map of property key -> property value
    */
   def append(props:Map[String, String]):Unit = {}
+  def append(props:java.util.Map[Object, Object]):Unit = {}
 
   /**
    * Updates Props to find interpolated values in the argument.
    * @param props
    */
   def appendInterpolator(props:Map[String, String]):Unit = {}
+  def appendInterpolator(props:java.util.Map[Object, Object]):Unit = {}
 
   /**
    * Enumeration of available run modes.
--- a/core/util/src/main/scala/net/liftweb/util/Props.scala
+++ b/core/util/src/main/scala/net/liftweb/util/Props.scala
@@ -26,37 +26,43 @@ import common._
 
 private[util] trait Props extends Logger {
   type PropProvider = scala.collection.Map[String, String]
+  type InterpolationValues = scala.collection.Map[String, String]
 
   /**
    * Get the configuration property value for the specified key.
+   *
    * @param name key for the property to get
    * @return the value of the property if defined
    */
-  def get(name: String): Box[String] =
-    sources.map(src => Box(src.get(name)))
-      .find(_.isDefined)
-      .getOrElse(Empty)
+  def get(name: String): Box[String] = {
+    lockedProviders
+      .flatMap(_.get(name))
+      .headOption
       .map(interpolate)
+  }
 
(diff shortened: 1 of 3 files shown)


core/util/src/main/scala/net/liftweb/util/Props.scala
```diff
@@ -53,17 +54,45 @@ object Props extends Logger {
...
+    val tries:List[Box[Object]] =
+      prepended.map(m => Box.legacyNullTest(m get name)) ++
+      (props.get(name).toBox ::
+      appended.map(m => Box.legacyNullTest(m get name)))
+
+    tries.find(_.isDefined).getOrElse(Empty).map(interpolate)
+  }
+
+  private val INTERPOLATE = """(.*?)\Q${\E(.*?)\Q}\E([^$]*)""".r
+  private def interpolate(v:Object):String = {
```
farmdawgnation: :-1: on naming these two the same thing. Let's name the all-caps `INTERPOLATE` to `interpolateRegex` or something like that.

farmdawgnation: Also, what's the reasoning behind using the single letter variable name `v`?

core/util/src/main/scala/net/liftweb/util/Props.scala
```diff
@@ -328,5 +378,18 @@ object Props extends Logger {
         Map()
     }
   }
+
+  private var prepended:List[java.util.Map[Object, Object]] = Nil
+  private var appended:List[java.util.Map[Object, Object]] = Nil
+  private var interpolators:List[java.util.Map[Object, Object]] = Nil
```
farmdawgnation: It seems strange to me as a Scala framework to use an instance of `java.util.Map` for being our Record of Truth™. What's the reasoning there?

core/util/src/main/scala/net/liftweb/util/Props.scala
```diff
@@ -328,5 +378,18 @@ object Props extends Logger {
...
+
+  /**
+   * Resets the Props stacks to allow for isolated unit tests
+   */
+  private [util] def testReset() = {
+    prepended = Nil
+    appended = Nil
+    interpolators = Nil
+  }
```
farmdawgnation: Is there something else we could do here aside from adding code to the mainline branch that's only ever going to be used in tests?

core/util/src/test/scala/net/liftweb/util/PropsSpec.scala
```diff
@@ -17,16 +17,18 @@
...
-/**
- * Systems under specification for Lift Mailer.
- */
-object PropsSpec extends Specification {
+object PropsSpec extends Specification with AfterEach {
   "Props Specification".title
   sequential
 
+  // TODO: Why doesn't this work??
+  def after = Props.testReset()
```
farmdawgnation: o.O

core/util/src/main/scala/net/liftweb/util/Props.scala
joescii: Good call on `INTERPOLATE`. I think originally I needed it to start with a capital to be a stable identifier for a pattern match.

As for `v`, I thought `interpolate` with a param `v` was sufficient to communicate that it is the value to interpolate. Since it is a private method, I didn't see any reason to err on the side of expounding further.  However, I'm accustomed to being told I'm wrong with my short variable names, so I've already updated it.

core/util/src/main/scala/net/liftweb/util/Props.scala
joescii: I expect `System.env` and `System.properties` to be used here, and both of those are mutable maps.  If we go with Scala maps, then we will take an snapshot at the time it is passed to us.  While I hope that not many people are mutating their system properties, I don't feel good about having it passed to us then we no longer call through to the map.

core/util/src/main/scala/net/liftweb/util/Props.scala
joescii: I'll look into using `RulesSeq[T]`. Perhaps there I can get the append/prepend that I want without having to finagle a testing hook.

core/util/src/test/scala/net/liftweb/util/PropsSpec.scala
joescii: Yeah, never figured out why `after` didn't do what I expected. Maybe it won't matter if I get what we want from using `RulesSeq[T]`. :)

core/util/src/main/scala/net/liftweb/util/Props.scala
eltimn: Why wouldn't sys.env and sys.props be used instead? They both return immutable.Map.

joescii: I got really excited about this idea, but [the scaladoc for `scala.sys`](http://scala-lang.org/files/archive/api/current/#scala.sys.package) shows that `props` is a `SystemProperties` which is a `mutable.Map`. Regardless, I feel a lot better about using a scala collection here, but still not too keen on making them immutable. I would be afraid that someone would be confused if the java props got changed and not reflected in the Lift `Props`.  What do you think, @farmdawgnation?

Shadowfiend: Fwiw I'm not sure he was suggesting we use an immutable map to begin with, just that we use a Scala one. I'm down for that (`mutable.Map` rather than `java.util.Map`).

joescii: Ah yes, that's a good point. I've never used a `mutable.Map`. I'll look into how I can write our stuff to not care so much about the mutability since we're doing read-only.

farmdawgnation:
Worth noting, Codacy had some [interesting suggestions](https://www.codacy.com/app/Lift/framework/pullRequest?prid=97701)

joescii:
Based on everyone's feedback: (1) I want to change the interface for this feature to use `RuleSeq` for the `append` and `prepend` configuration functions (and hence move the configuration from `Props` itself to `LiftRules` where everything else is configured anyway). (2) I will remove the use of `java.util.Map` as the container type (replace with either a Scala `Map` or structural type `{ def:get(String):String }` depending on how things shake out).

That being stated, I feel that it's a re-write at this point, so I'm closing this PR.  Thanks everyone for your valuable critical feedback.

joescii:
Bah! Can't use `RuleSeq` because that's over in `web` whereas `Props` is in `core`.

joescii:
Given that I couldn't use `RuleSeq` and I was able to just change from `java.util.Map` to `scala.collection.Map` which allows either mutable or immutable maps, it made much more sense to just update the existing branch.

core/util/src/main/scala/net/liftweb/util/Props.scala
```diff
@@ -48,22 +49,49 @@ import common._
...
 
   // def apply(name: String): String = props(name)
 
   def getInt(name: String): Box[Int] = get(name).map(toInt) // toInt(props.get(name))
   def getInt(name: String, defVal: Int): Int = getInt(name) openOr defVal // props.get(name).map(toInt(_)) getOrElse defVal
-  def getLong(name: String): Box[Long] = props.get(name).flatMap(asLong)
+  def getLong(name: String): Box[Long] = get(name).flatMap(asLong)
   def getLong(name: String, defVal: Long): Long = getLong(name) openOr defVal // props.get(name).map(toLong(_)) getOrElse defVal
-  def getBool(name: String): Box[Boolean] = props.get(name).map(toBoolean) // (props.get(name))
+  def getBool(name: String): Box[Boolean] = get(name).map(toBoolean) // (props.get(name))
```
Shadowfiend: This comment seems as innaccurate now as it was before :)

core/util/src/main/scala/net/liftweb/util/Props.scala
```diff
@@ -85,6 +113,38 @@ object Props extends Logger {
...
+   * @param provider Arbitrary map of property key -> property value
+   */
+  def appendInterpolator(provider:PropProvider):Unit = interpolators = interpolators :+ provider
+
+  /**
+   * Removes any PropProvider interpolators which match the predicate
+   * @param removeIf predicate for selecting PropProviders to remove
+   */
+  def removeInterpolators(removeIf: PropProvider=>Boolean):Unit =
+    interpolators = interpolators.filterNot(removeIf)
```
Shadowfiend: Why not expose everything except for prepend/append as `updateInterpolators(updater: Seq[PropProvider])`? That way the user can do whatever they want.

core/util/src/main/scala/net/liftweb/util/Props.scala
```diff
@@ -328,5 +388,8 @@ object Props extends Logger {
         Map()
     }
   }
+
+  private var sources:List[PropProvider] = List(props)
+  private var interpolators:List[PropProvider] = Nil
```
Shadowfiend: These should either both be volatile, or we should lock down source/interpolator management at some point.

core/util/src/main/scala/net/liftweb/util/Props.scala
```diff
@@ -48,22 +49,49 @@ import common._
...
+    val interpolated = for {
+      m <- interpolateRegex findAllMatchIn value.toString
+    } yield {
+      val before   = m group 1
+      val key      = m group 2
+      val after    = m group 3
+      val lookedUp = lookup(key).openOr(value.toString)
+
+      before + lookedUp + after
+    }
```
Shadowfiend: Wondering if this would be clearer using the regex as an extractor…

joescii: I definitely believe it will, but I'm not sure how to do that with `findAllMatchIn`

Shadowfiend: Turns out `unapplySeq` on `Regex` takes a `Match` as well, so you can:

```scala
    val interpolated = for {
      interpolateRegex(before, key, after) <- interpolateRegex findAllMatchIn value.toString
    } yield {
      val lookedUp = lookup(key).openOr(value.toString)

      before + lookedUp + after
    }
```

It also seems like `openOr`-ing the *value* is potentially dangerous—perhaps we should `openOr` the key instead, or the full thing we removed from the string (that is, `${<key>}`).

Last but not least, I'm a little wary of the `toString` conversions. How about we only run `interpolate` on something we know is a `String` already (e.g., using `Box.asA` in `get`)?

Shadowfiend:
Let me know if you're short on time here and want me to see if I can try some stuff, @joescii—I'm happy to help!

joescii:
Yeah, @Shadowfiend knock it out if you have time.  I am in agreement with your ideas :+1:

Shadowfiend:
Rebased against master and made the changes. Also did some naming tweaking in order to align the naming of the types and the functions that modify them, made `Props` a private trait that is extended by the singleton so that the tests can be properly isolated, and separated the types for the props providers and the interpolator values.

Let me know what you think!

joescii:
It looks awesome!  I had not thought about the private trait approach for
making these testable.  That's much cleaner than the mess I was trying to
make.  I also like how you locked the providers with lazy vals.  Both are
great ideas!

Joe

Shadowfiend:
Looks like some good feedback from the OG author ;)

Going to wait another day or so for some more feedback and then :shipit: .

Shadowfiend:
Alllll righty, let's play!

PR #1724, opened 2015-10-23
LAFuture Spec Reliability: Execute map/flatMap LAFuture tests on same thread for reliability.

farmdawgnation:
We were running into issues with test realibility in TravisCI with these
specs. So, let's make them more reliable by using an LAScheduler that
doesn't start a separate thread.


Shadowfiend:
👌

PR #1730, opened 2015-11-04
Future with session

joescii:
Our code base at work is getting littered with `S.session.map { session => S.initIfUninitted(session) { /** stuff */ } }` because we interleave threads from akka actors and futures.  Now that Lift supports only 2.11.x I think it would be a nice addition to have `LiftSession`-aware `scala.concurrent.Future[T]`'s that I've been working on for our project.

Just to take an excerpt from the spec code, this will now work:

``` scala
object TestVar1 extends SessionVar[String]("test1")
object TestVar2 extends SessionVar[String]("test2")

"A FutureWithSession" should {
  "have access to session variables in map() chains" in {
    val session = new LiftSession("Test Session", "", Empty)

    val future:Future[String] = S.initIfUninitted(session) {
      TestVar1("map1")
      TestVar2("map2")
      Future("something").withCurrentSession
    }
    val actual:Future[String] = future
      .map(_+"-"+TestVar1.get)
      .map(_+"-"+TestVar2.get)
    val expected:Option[Try[String]] = Some(Success("something-map1-map2"))

    actual.value must eventually(beEqualTo(expected))
  }
}
```

Because I converted the `Future("something")` into a `FutureWIthSession[T]` via the implicit `withCurrentSession` val, I am able to access the `SessionVar`s in the map lambdas.  

How do y'all feel about this addition? I've implemented everything we need at work, but I'll gladly finish this out if everyone likes it for Lift.

### DIFF
--- a/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
+++ b/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
@@ -262,8 +262,8 @@ object LAFuture {
    * @tparam T the type
    * @return an LAFuture that will yield its value when the value has been computed
    */
-  def apply[T](f: () => T, scheduler: LAScheduler = LAScheduler): LAFuture[T] = {
-    val ret = new LAFuture[T](scheduler)
+  def apply[T](f: () => T, scheduler: LAScheduler = LAScheduler, context:ContextFn = None): LAFuture[T] = {
+    val ret = new LAFuture[T](scheduler, context)
     scheduler.execute(() => {
       try {
       ret.satisfy(f())
@@ -280,8 +280,8 @@ object LAFuture {
    * @tparam T the type that
    * @return
    */
-  def build[T](f: => T, scheduler: LAScheduler = LAScheduler): LAFuture[T] = {
-    this.apply(() => f, scheduler)
+  def build[T](f: => T, scheduler: LAScheduler = LAScheduler, context:ContextFn = None): LAFuture[T] = {
+    this.apply(() => f, scheduler, context)
   }
 
   private val threadInfo = new ThreadLocal[List[LAFuture[_] => Unit]]
--- a/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
+++ b/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
@@ -24,7 +24,7 @@ import common._
  * A container that contains a calculated value
  * or may contain one in the future
  */
-class LAFuture[T](val scheduler: LAScheduler) {
+class LAFuture[T](val scheduler: LAScheduler = LAScheduler) {
   private var item: T = _
   private var failure: Box[Nothing] = Empty
   private var satisfied = false
@@ -32,10 +32,6 @@ class LAFuture[T](val scheduler: LAScheduler) {
   private var toDo: List[T => Unit] = Nil
   private var onFailure: List[Box[Nothing] => Unit] = Nil
   private var onComplete: List[Box[T] => Unit] = Nil
-
-  def this() {
-    this(LAScheduler)
-  }
 
   LAFuture.notifyObservers(this)
 
--- a/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
+++ b/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
@@ -79,7 +79,7 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler, context:LAFuture.Con
    * Private tail-recursive implementation of get
    */
   @scala.annotation.tailrec
-  private def get_rec: T = synchronized {
+  private [this] def get_rec: T = synchronized {
     if (satisfied) item
     else if (aborted) throw new AbortedFutureException(failure)
     else {
--- a/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
+++ b/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
@@ -77,15 +77,20 @@ class LAFuture[T](val scheduler: LAScheduler) {
   /**
    * Get the future value
    */
+  def get: T = get_rec
+
+  /**
+   * Private tail-recursive implementation of get
+   */
   @scala.annotation.tailrec
-  final def get: T = synchronized {
+  private def get_rec: T = synchronized {
     if (satisfied) item
     else if (aborted) throw new AbortedFutureException(failure)
     else {
       this.wait()
       if (satisfied) item
       else if (aborted) throw new AbortedFutureException(failure)
-      else get
+      else get_rec
     }
   }
 
--- a/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
+++ b/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
@@ -24,7 +24,7 @@ import common._
  * A container that contains a calculated value
  * or may contain one in the future
  */
-class LAFuture[T](val scheduler: LAScheduler = LAScheduler) {
+class LAFuture[T](val scheduler: LAScheduler = LAScheduler, wrapper:CommonLoanWrapper = LAFuture.defaultWrapper) {
   private var item: T = _
   private var failure: Box[Nothing] = Empty
   private var satisfied = false
@@ -48,7 +48,7 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler) {
           val ret = toDo
           toDo = Nil
           onFailure = Nil
-          onComplete.foreach(f => LAFuture.executeWithObservers(scheduler, () => f(Full(value))))
+          onComplete.foreach(f => LAFuture.executeWithObservers(scheduler, () => wrapper(f(Full(value)))))
           onComplete = Nil
           ret
         } else Nil
@@ -56,7 +56,7 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler) {
         notifyAll()
       }
     }
-    funcs.foreach(f => LAFuture.executeWithObservers(scheduler, () => f(value)))
+    funcs.foreach(f => LAFuture.executeWithObservers(scheduler, () => wrapper(f(value))))
   }
 
   /**
@@ -106,13 +106,13 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler) {
    * @return a Future that represents the function applied to the value of the future
    */
   def map[A](f: T => A): LAFuture[A] = {
-    val ret = new LAFuture[A](scheduler)
+    val ret = new LAFuture[A](scheduler, wrapper)
     onComplete(v => ret.complete(v.flatMap(n => Box.tryo(f(n)))))
     ret
   }
 
   def flatMap[A](f: T => LAFuture[A]): LAFuture[A] = {
-    val ret = new LAFuture[A](scheduler)
+    val ret = new LAFuture[A](scheduler, wrapper)
     onComplete(v => v match {
       case Full(v) =>
         Box.tryo(f(v)) match {
@@ -125,7 +125,7 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler) {
   }
 
   def filter(f: T => Boolean): LAFuture[T] = {
-    val ret = new LAFuture[T](scheduler)
+    val ret = new LAFuture[T](scheduler, wrapper)
     onComplete(v => ret.complete(v.filter(f)))
     ret
   }
@@ -176,7 +176,7 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler) {
    */
   def onSuccess(f: T => Unit) {
     synchronized {
-      if (satisfied) {LAFuture.executeWithObservers(scheduler, () => f(item))} else
+      if (satisfied) {LAFuture.executeWithObservers(scheduler, () => wrapper(f(item)))} else
       if (!aborted) {
         toDo ::= f
       }
@@ -190,7 +190,7 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler) {
    */
   def onFail(f: Box[Nothing] => Unit) {
     synchronized {
-      if (aborted) LAFuture.executeWithObservers(scheduler, () => f(failure)) else
+      if (aborted) LAFuture.executeWithObservers(scheduler, () => wrapper(f(failure))) else
       if (!satisfied) {
         onFailure ::= f
       }
@@ -204,8 +204,8 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler) {
    */
   def onComplete(f: Box[T] => Unit) {
     synchronized {
-      if (satisfied) {LAFuture.executeWithObservers(scheduler, () => f(Full(item)))} else
-      if (aborted) {LAFuture.executeWithObservers(scheduler, () => f(failure))} else
+      if (satisfied) {LAFuture.executeWithObservers(scheduler, () => wrapper(f(Full(item))))} else
+      if (aborted) {LAFuture.executeWithObservers(scheduler, () => wrapper(f(failure)))} else
       onComplete ::= f
     }
   }
@@ -227,8 +227,8 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler) {
       if (!satisfied && !aborted) {
         aborted = true
         failure = e
-        onFailure.foreach(f => LAFuture.executeWithObservers(scheduler, () => f(e)))
-        onComplete.foreach(f => LAFuture.executeWithObservers(scheduler, () => f(e)))
+        onFailure.foreach(f => LAFuture.executeWithObservers(scheduler, () => wrapper(f(e))))
+        onComplete.foreach(f => LAFuture.executeWithObservers(scheduler, () => wrapper(f(e))))
         onComplete = Nil
         onFailure = Nil
         toDo = Nil
@@ -250,6 +250,10 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler) {
 final class AbortedFutureException(why: Box[Nothing]) extends Exception("Aborted Future")
 
 object LAFuture {
+  val defaultWrapper:CommonLoanWrapper = new CommonLoanWrapper {
+    override def apply[T](f: => T): T = f
+  }
+
   /**
    * Create an LAFuture from a function that
    * will be applied on a separate thread. The LAFuture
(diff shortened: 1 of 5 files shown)


Shadowfiend:
Keep in mind we usually discuss additions on the list before putting them in—not mandatory for committers, but usually good form unless you urgently need something in the snapshot build :)

joescii:
Right, thanks for that reminder. I remember thinking to myself "don't forget to post this PR on the mailing list so everyone sees it." Like most of my after hours Lift work, always be suspicious of omissions and other midnight pilot errors. :)

Shadowfiend:
Do you still consider this WIP? Any chance we could get some scaladocs with a quick example?

web/webkit/src/test/scala/net/liftweb/http/FutureWithSessionSpec.scala
```diff
@@ -0,0 +1,272 @@
...
+  "A FutureWithSession" should {
+    "fail if session is not available" in {
+      val actual:Future[String] = Future("something").withCurrentSession
+
+      // TODO: Why the hell does this not work??
+//      val expected:Option[Try[String]] = Some(Failure(new Exception))
+//      actual.value must eventually(beEqualTo(expected))
+
+      actual.value.isDefined must eventually(beTrue)
+      actual.value.get must beFailedTry.withThrowable[Exception]("LiftSession not available in this thread context")
```
Shadowfiend: I think you can avoid `get`s and such here (and clear Codacy's PR annoyances) by using `beLike`:

```
actual.value must eventually(beLike {
  case Some(Failure(exception)) =>
    exception must beInstanceOf[exception]
})
```

joescii:
I've actually been using this code at work and it's rock solid.  I really wanted to get this same pattern to work for `LAFuture` before we tried shipping it because it feels odd to support Scala's `Future` but not our own.  Unfortunately, I got wrapped around the axle trying to get it to work.  I'm about to try another shot at it.

How do you feel about supporting only Scala `Future` this way?

joescii:
Well, good news!  Now that I'm trying it again, I have `LAFuture` working as expected.  It's just not complete yet.  I'll try to finish it today.

farmdawgnation:
woohoooo

joescii:
Removed with [WIP] label!

Shadowfiend:
I'm not going to lie, it makes me slightly twitchy that we have a class that contains an `LAFuture` and also extends `LAFuture`.

I'm wondering if at least for `LAFuture` we should instead add some functionality to `LAFuture` itself; in particular, a `wrappingFn` constructor parameter or some-such, that would always run around the `onComplete`/`onFail`/`onSuccess` stuff and passed on through maps/flatMaps/etc. Then the implicit conversion can just pre-populate the session stuff for that wrapping function.

I also have an itching suspicion we have some similar uses with `LoanWrapper` lying around for Lift, so maybe that's something that could see reuse. Not sure. Let me ponder this one a little bit, and if I can come up with something better maybe we sneak it in after M8.

joescii:
I didn't feel great about it either, TBH.  The pattern made sense when I was doctoring the Scala `Future`.  I really like your suggestion.  It's actually a bit more flexible, although I cannot offhand think of more use cases than `S.initIfUninitted`.  

I suppose then we wouldn't need the implicit conversion at all.  They could just be functions on `LAFuture`.  Which way do you think we should go with that?

joescii:
Just opened the hood.  We have to go the implicit conversion route.  I had forgotten that `LAFuture` is in the `core` project and `LiftSesson` is in the `http` project, so we can't have the reference to `LiftSession` back in `LAFuture` itself.

joescii:
Hmm... I thought it would _just work_, but only two of the interesting test cases work. Will have to check back when not late at night. :+1:

joescii:
After sleeping on it, the problem was very obvious.  As is usually the case when @Shadowfiend makes a suggestion, the code's design/implementation is light years a head of the first idea.  Thanks!

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
```diff
@@ -295,13 +300,13 @@ object LAFuture {
...
 
-  private def executeWithObservers(scheduler: LAScheduler, f: () => Unit) {
+  private def executeWithObservers(scheduler: LAScheduler, f: () => Unit, wrapper:CommonLoanWrapper) {
     val cur = threadInfo.get()
     scheduler.execute(() => {
       val old = threadInfo.get()
       threadInfo.set(cur)
       try {
-        f()
+        wrapper(f())
```
Shadowfiend: Wondering if we can make `wrapper` an `Option` instead and go from there?

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
```diff
@@ -0,0 +1,70 @@
...
+    override def apply[T](f: => T): T = S.initIfUninitted(session) { f }
+  }
+
+  private def withSession[T](f:LAFuture[T], s:LiftSession):LAFuture[T] = {
+    val sf = new LAFuture[T](f.scheduler, sessionWrapper(s))
+    f.onComplete(sf.complete)
+    sf
+  }
+
+  implicit class LAFutureDecorator[T](future:LAFuture[T]) {
```
Shadowfiend: This should be able to `extends AnyVal` I think.

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
```diff
@@ -0,0 +1,70 @@
...
+    val withCurrentSession:LAFuture[T] = S.session.map( s => withSession(future, s) )
+      .openOr {
+        val f = new LAFuture[T](future.scheduler)
+        f.fail(Failure("LiftSession not available in this thread context", Empty, Empty))
```
Shadowfiend: Seems like it would make sense for the chain on this `Failure` to be whatever we got from `S.session`?

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
joescii: I'm a fan of no-op base cases like this over conditional logic including `getOrElse`. Certainly we can. What do you like better about `Option`?

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
joescii: Sure... but what does extending `AnyVal` do for an object that's really just a namespace?

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
joescii: Good call.

joescii: Well.... I looked at the code for `S.session` and it's either a `Full` or `Empty`.  No `Failure` to propagate!

Shadowfiend: Well, not today, but it would be a better reflection of the intent of an API that returns a `Box` IMO.

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
Shadowfiend: Avoids an object allocation, I believe. With this code, when you do that invocation, you'll be instantiating a throwaway `LAFutureDecorator` to invoke that method on it. By making it a value class, you avoid that instantiation.

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
Shadowfiend: `Option` means we're doing a conditional check vs a function invocation, in what is likely to be the most common case for most users of the API.

Shadowfiend: (This is more guess than anything else—I'm not sure how good the JVM is at inlining this kind of nested function invocation in cases where it's a no-op vs how good it is at improving branching in cases of a simple conditional check.)

joescii: Right, I fault on the side of the code being clear more than guessing at how it may perform when it comes to something as sophisticated as the JVM.  I can really go either way here.  Do you still think the Option is a better route?

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
joescii: Cool, I didn't know about that optimization!

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
joescii: Makes sense.  I'll pass the `Failure` along, and use what I have today in the case of `Empty`. Sound about right?

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
Shadowfiend: I can go either way… `Option` seems like the most correct encoding of what we're doing, from a type perspective, I think, which is why I prefer it.

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
Shadowfiend: I think you can just:

```scala
S.session match {
  case Full(session) =>
    withSession(future, session)
  case emptyBox =>
    emptyBox ?~! "LiftSession not available in this thread context"
```

That should create a `Failure` if it's `Empty`, or chain to the existing
one if it's a `Failure`.

joescii: Make that TWO things I have learned from you on this PR.  Thanks!

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
joescii: Scalac said "no"

```
value class needs to have exactly one val parameter
object LAFutureWithSession extends AnyVal {
       ^
```

A quick googling didn't turn up anything that makes me feel we're trying to do the right thing here.

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
joescii: It turns out the types don't work correctly.  `?~!` returns a `Box[LiftSession]`.

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
```diff
@@ -0,0 +1,78 @@
...
+    sf
+  }
+
+  private def withFailure[T](f:LAFuture[T], failure:Failure):LAFuture[T] = {
+    val ff = new LAFuture[T](f.scheduler)
+    ff.complete(failure)
+    ff
+  }
+
+  implicit class LAFutureDecorator[T](future:LAFuture[T]) {
```
Shadowfiend: Re: your remark about the compiler, this needs to be `implicit class LAFutureDecorator[T](val future: LAFuture[T]) extends AnyVal {`. Note the `val` in front of `future`, which I have a tendency to forget when trying to do a value class. If you added it and it still complained at you, then never mind :)

If it works, you can probably apply the same thing to the Scala `Future` one as well.

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
```diff
@@ -0,0 +1,78 @@
...
+    val withCurrentSession:LAFuture[T] = S.session match {
+      case Full(s) =>   withSession(future, s)
+      case f:Failure => withFailure(future, f)
+      case Empty =>     withFailure(future, Failure("LiftSession not available in this thread context", Empty, Empty))
```
Shadowfiend: Re: the types, I think if you do:

```scala
S.session match {
  case Full(session) =>
    withSession(future, session)
  case empty: EmptyBox =>
    withFailure(future, empty ?~! "LiftSession not available in this thread context."
} 
```

Right? `EmptyBox`.`?~!` [should always return a `Failure`](https://github.com/lift/framework/blob/master/core/common/src/main/scala/net/liftweb/common/Box.scala#L870).

joescii: Yes! That works because `EmptyBox` extends `Box[Nothing]`.  I didn't know about `EmptyBox` :)

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
joescii: Ah ha... I was confused.  If you go back to my compiler error, I was trying to make the object extend AnyVal.  I didn't realize you meant the implicit class.

joescii:
BTW, as my last two commits show, I combed through the codacy issues.  The remaining are either on purpose or incorrectly flagged.

Shadowfiend:
Yeah, Codacy's had some oddities lately. I asked them about it, and they said they're aware of it but right this second they're focused on some other stuff. If it keeps up for another couple of months it may be worth dropping it altogether.

dpp:
Complaining about if/else on a boolean? Sheesh. Codacy seems more noise than signal.

farmdawgnation:
Yeah. I actually surfaced a lot of these complaints with one of their founders because of some issues we've had at Domino with it. He's saying that come February/March there should be a lot of improvements coming down the pipe. I think I'm going to keep an eye on what the suggestions look like moving forward on lift/framework before making a call on turning it back on for Domino. :/

dpp:
Does Code Climate do Scala yet? May be worth a bake-off.

farmdawgnation:
Doesn't look like it. Though it seems they might have opened up part of their platform for extension.

farmdawgnation:
I just applied to join their developer program. Maybe they'll let me wire in some simple scalachecking? < shrug>

Shadowfiend:
It's worth noting that codacy lets us turn off certain warnings, though I don't recall how fine-grained it is.

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
```diff
@@ -295,13 +298,13 @@ object LAFuture {
     }
   }
 
-  private def executeWithObservers(scheduler: LAScheduler, f: () => Unit) {
+  private def executeWithObservers(scheduler: LAScheduler, f: () => Unit, context:ContextFn) {
```
Shadowfiend: Could this be `private[LAFuture]`? Not sure if that locks out the companion class or not.

Shadowfiend:
I've nuked the imports-at-the-beginning-of-file issue at a project level, I think, and ignored some of the more specific ones that are still potentially valuable (like the if vs match one).

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
joescii: No, I tried that per the Codacy recommendation. The companion class can't see it.

Shadowfiend: Got it, cool. Will mark it accordingly :)

Bhashit:
Hello people, any chance this is going to be merged anytime soon, or a similar solution.

farmdawgnation:
We're targeting this for 3.1 so it'll likely be merged after 3.0 goes final.

Shadowfiend:
Incidentally @pdyraga just ran into this full-speed at @elemica heh.

Bhashit:
@Shadowfiend By the way, I've just had to create something similar to this, called `FutureWithContext` that uses `LiftSession#buildDeferredFunction` so that the code can access RequestVars etc.

Shadowfiend:
Yes, saw that. Definitely a recurring pain point.

Bhashit:
@joescii Just an idea: I have been working on something similar, but I needed RequestVars too (using `LiftSession#buildDeferredFunction`). I'm not completely sure about this, but wouldn't it be nicer if the Futures were aware of both sessions and request context? RequestVars are commonly used for a lot of stuff (in Lift internals as well).

pdyraga:
I understand that PR is on a very advanced level but as @Shadowfiend said, I've ran into this issue full speed recently so please pardon me and let me share some thoughts.

The problem with `Future` and `LAFuture` is that they start evaluating as soon as there's a worker thread available. It means that even if we do `.withCurrentSession`, inner task may fail because of session unavailability.

We often use `Future`s to lazy-load some part of a page, e.g. retrieve list of users from a database and then apply Lift bindings on them (for 2.6 it's possible with http://www.ontheserverside.com/blog/2014/11/27/binding-futures-in-lift-2/ and for 3 it's out of the box thanks to @Shadowfiend).

So far, I've found two solutions for this problem: custom `LAScheduler` / `ExecutionContext` and kind of a factory method for `Future`s / `LAFuture`s.

First possibility: 
```
object SnippetScheduler extends LAScheduler {
  def execute(f: () => Unit): Unit = {
    val session = S.session openOrThrowException "LiftSession is not available"
    LAScheduler.execute(session.buildDeferredFunction(f))
  }
}

object SnippetExecutionContext {
  object Implicits {
    implicit lazy val instance = new SnippetExecutionContext()
  }
}

private[helpers] class SnippetExecutionContext extends ExecutionContext with Loggable {

  def execute(runnable: Runnable): Unit = {
    SnippetScheduler.execute(() => runnable.run())
  }

  def reportFailure(cause: Throwable): Unit = {
    logger.error("Could not execute Runnable", cause)
  }
}
```

and later use it as:
```
def work() {
  // must be done in session scope (e.g. calls S.?)
}

// With LAFuture:
LAFuture(() => work(), SnippetScheduler)

// With Scala Future:
import helpers.SnippetScheduler
Future {
  work()
}
```

Another possibility is a factory method that will produce `Future` with task already wrapped in session context:
```
object FutureHelpers {
  def withCurrentSession[T](task: () => T)(implicit executionContext: ExecutionContext) = {
    val session = S.session openOrThrowException "LiftSession is not available"
    val taskInSession = session.buildDeferredFunction(task)
    Future(taskInSession())
  }
}
```
and later:
```
FutureHelpers.withCurrentSession(work)
```

Also, I tend do agree with @Bhashit about access to `RequestVar`s. I think that "session-aware Futures" should offer the same access scope as `LazyLoad` that we already have in Lift.

Shadowfiend:
@joescii what do you think of closing this in favor of #1813 ?

joescii:
I hope to soon do the due diligence of checking out that PR, but what
little bit I've seen about it, I suspect it's a much better path than what
I attempted.

farmdawgnation:
Superseded by #1813.

PR #1733, opened 2015-11-24
Fix BaseParsers ws parser to not match newlines.

Shadowfiend:
In lift-markdown, BaseParsers was using a regex containing `\v` to indicate a
vertical tab. As of Java 8, `\v` refers to a character class that contains all
vertical whitespace, including newlines. We now explicitly use the unicode
value for vertical tab in the regex, fixing the failing test in Java 8 and
generally making things work as expected.

See [initial report from @farmdawgnation ](https://groups.google.com/forum/#!topic/liftweb/10aF9sk-jO0).


farmdawgnation:
:+1:

PR #1734, opened 2015-11-25
CSS URL prefixer comma fix, % fix, and specs

### LINKED ISSUE #1732: fixCSS fails when URL contains a comma
(As discussed in https://groups.google.com/d/msg/liftweb/SUxP84MYk7c/K7lRbL_GCwAJ ) Minimal example:

    net.liftweb.util.CSSParser("").fixCSS("body { background: url(','); }")

Expected outcome:

    Full("body { background: url(','); }")

Outcome:

    Failure(Parser did not consume all input.  Parser error?,Empty,Empty)

The case is relevant because base64 encoded inline graphics contain a comma:

    data:image/svg+xml;base64,PD94...

A quick fix for this scenario might be to simply add the comma to the list of characters in https://github.com/lift/framework/blob/a3075e0676d60861425281427aa5f57c02c3b0bc/core/util/src/main/scala/net/liftweb/util/CSSHelpers.scala#L96-L103

Shadowfiend:
While fixing #1732, did some renaming (with deprecations), added a spec
for `CssUrlPrefixer`, and fixed the originally-reported comma issue, the same
issue with supporting `%` in CSS URLs, as well as fixing the parser's insistence
on using single quotes to wrap a URL even when the original CSS either had no
quotes or double quotes.

Biggest thing I think needs review here beyond the fixes is that any old classes
or methods are still accessible at their old names with deprecation.

Closes #1732 .

### DIFF
--- a/core/util/src/main/scala/net/liftweb/util/CSSHelpers.scala
+++ b/core/util/src/main/scala/net/liftweb/util/CSSHelpers.scala
@@ -102,26 +102,68 @@ case class CssUrlPrefixer(prefix: String) extends Parsers  {
 
 
   lazy val spaces = (elem(' ') | elem('\t') | elem('\n') | elem('\r')).*
+
+  def pathWith(additionalCharacters: Char*) = {
+    elem("path",
+      c => c.isLetterOrDigit ||
+           c == '?' || c == '/' ||
+           c == '&' || c == '@' ||
+           c == ';' || c == '.' ||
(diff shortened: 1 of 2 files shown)


Shadowfiend:
I'll be merging this around the end of the week assuming no feedback :)

Shadowfiend:
Seeing no feedback, in we go!

PR #1736, opened 2015-11-26
JSON JS escaping

### LINKED ISSUE #1676: Lift-json and the dreaded U+2028 character: to escape or not?
From this [thread](https://groups.google.com/d/topic/liftweb/X_u8FoJ2j88/discussion)

Bug (matt): 

> We ran into an interesting issue on our servers tonight, involving the dreaded u\2028 character. After some research, I learned something new: There are some characters that are valid javascript but not valid json, including U+2028.

> Essentially, someone inserted some data to our db with one of these characters, and once we send this down as json (a JObject), the client throws an 'Unexpected Token ILLEGAL' error. As the above link describes, this is character is not valid json.

> ow, I could easily escape these characters on my end, but I wonder if it isn't a better idea to have lift-json handle this case. It would avoid a lot of potential complications, methinks. Antonio recommended a potential solution being to have a toggle on the implicit formats value, but I would go so far as to say, since 'lift-json' handles json, it should *always* escape these characters since they are invalid json.

Solutions (antonio):

> (1) make the user do their own escaping
> (2) bake it in so u+202[89] are always escaped by lift-json
> (3) insert some sort of toggle that allows us to enable this escaping, so that it is disabled by default when lift-json is just dealing in plain JSON.

> I'm okay with the formats solution, or, because this is a fairly specific case, could we just have an extra parameter to render in an overload to avoid API changes and roll this in now rather than later? Mostly I'm keeping in mind the fact that for Lift, as a web framework, there are two core/critical consumers of JSON: other JSON libraries, and JavaScript. Thus, I think we should make sure that JS is a well-considered consumer as soon as possible, though I'm fine with avoiding inefficiencies (i.e., longer encodings) for non-JS consumers.

> I should mention, according to this StackOverflow question, it looks like these are the characters that need JS escaping:

> \u0000\u00ad\u0600-\u0604\u070f\u17b4\u17b5\u200c-\u200f\u2028-\u202f\u2060-\u206f\ufeff\ufff0-\uffff

Shadowfiend:
I added a few scaladocs beyond the core work. The main commit:

Until now, we've had a hardcoded set of characters that would always be escaped
in JSON; these were never modifiable. We now plug in to the new `RenderSettings`
case class to add the ability to specify sets of characters that should be
escaped.

Additionally, we add a default set of such characters that corresponds to the
characters that must be escaped when JSON is evaluated directly by a JavaScript
engine rather than by a valid JSON parser (as when using JSON-P).

Lastly, we provide two default `RenderSettings` instances, `prettyJs` and
`compactJs`, that encapsulate the appropriate escaping with pretty or compact
rendering.

Closes #1676.

### DIFF
--- a/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
+++ b/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
@@ -846,7 +846,13 @@ object JsonAST {
   }
 
   object RenderSettings {
+    /**
+     * Pretty-print JSON with 2-space indentation.
+     */
     val pretty = RenderSettings(2)
+    /**
+     * Compact print JSON on one line.
+     */
     val compact = RenderSettings(0)
 
     /**
@@ -878,6 +884,13 @@ object JsonAST {
      */
     val compactJs = RenderSettings(0, jsEscapeChars)
   }
+  /**
+   * RenderSettings allows for customizing how JSON is rendered to a String.
+   * At the moment, you can customize the indentation (if 0, all the JSON is
+   * printed on one line), the characters that should be escaped (in addition
+   * to a base set that will always be escaped for valid JSON), and whether or
+   * not a space should be included after a field name.
+   */
   case class RenderSettings(
     indent: Int,
     escapeChars: Set[Char] = Set(),
@@ -886,10 +899,16 @@ object JsonAST {
     val lineBreaks_? = indent > 0
   }
 
+  /**
+   * Render `value` using `RenderSettings.pretty`.
+   */
   def prettyRender(value: JValue): String = {
     render(value, RenderSettings.pretty)
   }
 
+  /**
+   * Render `value` to the given `appendable` using `RenderSettings.pretty`.
+   */
   def prettyRender(value: JValue, appendable: Appendable): String = {
     render(value, RenderSettings.pretty, appendable)
   }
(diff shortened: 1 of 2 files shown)


core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -845,20 +846,69 @@ object JsonAST {
...
   case class RenderSettings(
     indent: Int,
+    escapeChars: Set[Char] = Set(),
     spaceAfterFieldName: Boolean = false
   ) {
     val lineBreaks_? = indent > 0
   }
 
+  /**
+   * Render `value` using `RenderSettings.pretty`.
```
farmdawgnation: Maybe this should mention it's only for JSON instead of expecting people to click through to the render settings docs? Also, could we make the reference to those settings a link?

farmdawgnation:
I like it.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
Shadowfiend: Not sure what you mean regarding “only for JSON”—it's in `JsonAST` and `lift-json`, right?

I'll definitely make them a link, good catch.

Shadowfiend:
We probably also need to change Lift's built-in stuff to use the right render settings :/

Shadowfiend:
Actually just kidding, this is only for stuff that parses JSON as JS, and the Lift JSON stuff generally is parsed as JSON on the browser, so we're good.

Shadowfiend:
@farmdawgnation dropped some links on the scaladocs :tada:

farmdawgnation:
:+1:

PR #1737, opened 2015-11-29
Making the Msgs fade and effect scripts append to lift's page script

karma4u101:
Instead of (as in lift 2.6) using the in-lining tailScript function Msgs fade and effect scripts will now instead append to lifts page script. The change makes it better align with Lifts CSP settings.

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/builtin/snippet/Msg.scala
+++ b/web/webkit/src/main/scala/net/liftweb/builtin/snippet/Msg.scala
@@ -116,13 +116,13 @@ object Msg extends DispatchSnippet {
   }
 
   /**
-   * This method renders a <script/> element that renders effects for
+   * This method appends a script element to lift's page script that renders effects for
    * the given id.
    *
    * @see net.liftweb.builtin.snippet.Msgs#effects[T](Box[NoticeType.Value],String,T,Box[JsCmd => T])
    */
   def effects(id: String): NodeSeq =
-    Msgs.effects(Empty, id, NodeSeq.Empty, Msgs.tailScript)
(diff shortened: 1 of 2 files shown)


Shadowfiend:
Should we also nuke the [only other invocation of `tailScript`](https://github.com/lift/framework/blob/a3075e0676d60861425281427aa5f57c02c3b0bc/web/webkit/src/main/scala/net/liftweb/builtin/snippet/Msg.scala#L125) and then get rid of that method altogether?

karma4u101:
Sounds good to me, sry I somehow missed that one ;)

karma4u101:
PR is updated the tailScript function is removed as it is no longer referenced.

web/webkit/src/main/scala/net/liftweb/builtin/snippet/Msg.scala
```diff
@@ -116,13 +116,13 @@ object Msg extends DispatchSnippet {
   }
 
   /**
-   * This method renders a <script/> element that renders effects for
+   * This method appends a script element to lift's page script that renders effects for
```
Shadowfiend: Maybe “appends JavaScript to the current page's page script that…” rather than “appends a script element”? Definitely it's not appending an element.

Shadowfiend:
Lookin' good! Quick scaladoc comment, but other than that I'm diggin' it!

karma4u101:
@Shadowfiend scaladoc is updated lmkwut

Shadowfiend:
Looks good! :+1: and merging so I can check that lift/lift_30_sbt#2 works as expected!

PR #1741, opened 2015-12-06
Well Rested Futures: Implement Scala Future support in RestHelper

### LINKED ISSUE #1735: Help RestHelper understand Scala Futures.
As of the earliest work on Lift 3 by David, RestHelper understands `LAFuture`
and properly wraps `LAFuture` responses in a continuation. Let's add that for
standard Scala `Future`s as well and unify our support for futures across the
codebase (since we already support them in snippet bindings).

Shadowfiend:
`RestHelper` has long supported returning an `LAFuture` as the result of
a `RestHelper` block and auto-wrapped those in `RestHelper.async` so
they're dealt with via continuation, much like comet responses. This PR
extends `RestHelper` to support the same functionality for Scala `Future`s,
thus rounding out our dual support for Scala and Lift futures both here and
in snippet bindings.

We do this by moving the `CanResolveAsync` abstraction introduced for
snippet async future binding into `lift-util` and then reusing that abstraction
in `RestHelper`. We also add some specs for `CanResolveAsync` and a
spec to prove that `LAFuture` is still working as expected.

Closes #1735.

### DIFF
--- a/core/util/src/main/scala/net/liftweb/util/CanResolveAsync.scala
+++ b/core/util/src/main/scala/net/liftweb/util/CanResolveAsync.scala
@@ -40,7 +40,10 @@ trait CanResolveAsync[ResolvableType, ResolvedType] {
    */
   def resolveAsync(resolvable: ResolvableType, onResolved: (ResolvedType)=>Unit): Unit
 }
-object CanResolveAsync {
+
+trait LowPriorityCanResolveAsyncImplicits {
+  self: CanResolveAsync.type =>
+
   // Low priority implicit for resolving Scala Futures.
   implicit def resolveFuture[T](implicit executionContext: ExecutionContext) = {
     new CanResolveAsync[Future[T],T] {
@@ -59,3 +62,4 @@ object CanResolveAsync {
     }
   }
 }
+object CanResolveAsync extends LowPriorityCanResolveAsyncImplicits
--- a/core/util/src/main/scala/net/liftweb/util/CanResolveAsync.scala
+++ b/core/util/src/main/scala/net/liftweb/util/CanResolveAsync.scala
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2015 WorldWide Conferencing, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.liftweb
+package util
+
+import scala.concurrent.{ExecutionContext,Future}
+
+import actor.LAFuture
+
+/**
+ * Represents a unifying class that can resolve an asynchronous container.  For
+ * example, a `Future[String]` can be resolved by a
+ * `CanResolveAsync[Future[String], String]`.
+ *
+ * Provides one method, `[[resolveAsync]]`, that takes the async container and
+ * a function to run when the container resolves.
+ */
+trait CanResolveAsync[ResolvableType, ResolvedType] {
+  /**
+   * Should return a function that, when given the resolvable and a function
+   * that takes the resolved value, attaches the function to the resolvable
+   * so that it will asynchronously execute it when its value is resolved.
+   *
+   * See `CanResolveFuture` and `CanResolveLAFuture` in `lift-webkit` for
+   * example usages.
+   */
+  def resolveAsync(resolvable: ResolvableType, onResolved: (ResolvedType)=>Unit): Unit
+}
+object CanResolveAsync {
+  // Low priority implicit for resolving Scala Futures.
+  implicit def resolveFuture[T](implicit executionContext: ExecutionContext) = {
+    new CanResolveAsync[Future[T],T] {
+      def resolveAsync(future: Future[T], onResolved: (T)=>Unit) = {
+        future.foreach(onResolved)
+      }
+    }
+  }
+
+  // Low priority implicit for resolving Lift LAFutures.
+  implicit def resolveLaFuture[T] = {
+    new CanResolveAsync[LAFuture[T],T] {
+      def resolveAsync(future: LAFuture[T], onResolved: (T)=>Unit) = {
+        future.onSuccess(onResolved)
+      }
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


core/util/src/main/scala/net/liftweb/util/CanResolveAsync.scala
```diff
@@ -0,0 +1,61 @@
...
+   * that takes the resolved value, attaches the function to the resolvable
+   * so that it will asynchronously execute it when its value is resolved.
+   *
+   * See `CanResolveFuture` and `CanResolveLAFuture` in `lift-webkit` for
+   * example usages.
+   */
+  def resolveAsync(resolvable: ResolvableType, onResolved: (ResolvedType)=>Unit): Unit
+}
+object CanResolveAsync {
+  // Low priority implicit for resolving Scala Futures.
```
farmdawgnation: What makes an implicit low or high priority?

core/util/src/test/scala/net/liftweb/util/CanResolveAsyncSpec.scala
```diff
@@ -0,0 +1,56 @@
...
+import org.specs2.mutable.Specification
+
+import actor.LAFuture
+
+object CanResolveAsyncSpec extends Specification {
+  "CanResolveAsync" should {
+    "resolve Scala Futures" in {
+      val myPromise = Promise[String]()
+
+      val resolver = implicitly[CanResolveAsync[Future[String], String]]
```
farmdawgnation: What does `implicitly` do again? The scaladoc is particularly unhelpful.

web/webkit/src/main/scala/net/liftweb/http/rest/RestHelper.scala
```diff
@@ -543,10 +542,17 @@ trait RestHelper extends LiftRules.DispatchPF {
...
+    asyncContainer: AsyncResolvableType
+  )(
+    implicit asyncResolveProvider: CanResolveAsync[AsyncResolvableType, T],
+             responseCreator: T => LiftResponse
+  ): () => Box[LiftResponse] = () => {
     RestContinuation.async(reply => {
-      in.foreach(t => reply.apply(c(t)))
+      asyncResolveProvider.resolveAsync(
+        asyncContainer,
+        { resolved => reply(responseCreator(resolved)) }
```
farmdawgnation: Couldn't this also be `reply(responseCreator(_))`?

web/webkit/src/test/scala/net/liftweb/http/rest/RestHelperSpec.scala
```diff
@@ -17,20 +20,46 @@ class RestHelperSpec extends WebSpec(RestHelperSpecBoot.boot _) {
...
+
+    "respond async with LAFutures" withReqFor testFutureReq in { req =>
+      val helper = FutureRestSpecHelper()
+
+      try {
+        helper(req)()
+
+        failure("Failed to respond asynchronously.")
+      } catch {
+        case ContinuationException(_, _, resolverFunction) =>
```
farmdawgnation: Lulwut. That's how continuations work?!

web/webkit/src/test/scala/net/liftweb/http/rest/RestHelperSpec.scala
```diff
@@ -17,20 +20,46 @@ class RestHelperSpec extends WebSpec(RestHelperSpecBoot.boot _) {
...
     "give the correct response" withReqFor testOptionsReq in { req =>
-      RestHelperSpecRest(req)() match {
-        case Full(OkResponse()) => success
-        case other =>              failure("Invalid response : " + other)
+      RestHelperSpecRest(req)() must beLike {
+        case Full(OkResponse()) => ok
+      }
+    }
+
+    "respond async with LAFutures" withReqFor testFutureReq in { req =>
```
farmdawgnation: Do we also want to assert they respond async with Scala futures?

core/util/src/main/scala/net/liftweb/util/CanResolveAsync.scala
Shadowfiend: Where it is in the resolution order; these, however, are misnomers. They'd be low priority if they were in a trait mixed into the singleton. Need to fix that.

core/util/src/test/scala/net/liftweb/util/CanResolveAsyncSpec.scala
Shadowfiend: Best way to phrase it is that it fills in what the compiler would give you if you wanted to pass this to an implicit parameter.

web/webkit/src/test/scala/net/liftweb/http/rest/RestHelperSpec.scala
Shadowfiend: I need to change this to say “respond async with something that CanResolveAsync”. No need to test whether there are `CanRespondAsync`s for both `LAFuture` and `Future`, since that's verified elsewhere.

web/webkit/src/test/scala/net/liftweb/http/rest/RestHelperSpec.scala
Shadowfiend: Yep. We [catch the exception](https://github.com/lift/framework/blob/a3075e0676d60861425281427aa5f57c02c3b0bc/web/webkit/src/main/scala/net/liftweb/http/LiftServlet.scala#L173-L174) and then make a call as to whether we'll spin-wait or let the container's internal request continuation handling do its thing, depending on what's available.

farmdawgnation: spin wait

web/webkit/src/test/scala/net/liftweb/http/rest/RestHelperSpec.scala
farmdawgnation: Yep

core/util/src/test/scala/net/liftweb/util/CanResolveAsyncSpec.scala
farmdawgnation: O.o

web/webkit/src/main/scala/net/liftweb/http/rest/RestHelper.scala
Shadowfiend: Negative, the compiler barfs.

core/util/src/main/scala/net/liftweb/util/CanResolveAsync.scala
Shadowfiend: Made them actually low priority.

web/webkit/src/test/scala/net/liftweb/http/rest/RestHelperSpec.scala
Shadowfiend: Fix't 'er.

Shadowfiend:
This should be good for another :eyes:.

farmdawgnation:
Code looks good but it looks like your spec fails in Travis.

https://travis-ci.org/lift/framework/builds/96416723#L2972-L2977

Shadowfiend:
I'm thinking probably the `satisfy` call satisfies async rather than sync, so I need to drop the `.complete_?` check in the specs.

Shadowfiend:
Looks like the remaining codacy issues are related to duplication and “complexity”, but I don't see how we can improve on those, so I think this should be good to merge.

farmdawgnation:
Lehgo!

PR #1742, opened 2015-12-08
The FocusOnLoad object and avoiding inlining javascript 

karma4u101:
The FocusOnLoad object is now avoiding in-lining javascript by now instead utilizing lifts page script file.

While working on the lift_30_sbt lift_basic template I noticed that the ProtoUser controlled login page added a **in-line** element focus script via ProtoUser's use of the FocusOnLoad object. To avoid relaxing Lift 3's CSP rules by setting UnsafeInline for the script sources in lift_basic a change to the FocusOnLoad object was needed.

### DIFF
--- a/persistence/proto/src/main/scala/net/liftweb/proto/ProtoUser.scala
+++ b/persistence/proto/src/main/scala/net/liftweb/proto/ProtoUser.scala
@@ -796,7 +796,7 @@ trait ProtoUser {
    * default: S.?("email.address.not.found")
    */
   def userNameNotFoundString: String = S.?("email.address.not.found")
-
+ 
   def loginXhtml = {
     (<form method="post" action={S.uri}><table><tr><td
               colspan="2">{S.?("log.in")}</td></tr>
@@ -867,8 +867,9 @@ trait ProtoUser {
     val emailElemId = nextFuncName
     S.appendJs(Focus(emailElemId))
     val bind =
-      ".email" #> <input type="text" name="username" id={emailElemId}/> &
-      ".password" #> <input type="password" name="password"/> &
+      ".email [id]" #> emailElemId &
+      ".email [name]" #> "username" &
+      ".password [name]" #> "password" &
       "type=submit" #> loginSubmitButton(S.?("log.in"))
 
     bind(loginXhtml)
--- a/persistence/proto/src/main/scala/net/liftweb/proto/ProtoUser.scala
+++ b/persistence/proto/src/main/scala/net/liftweb/proto/ProtoUser.scala
@@ -864,8 +864,10 @@ trait ProtoUser {
       }
     }
 
+    val (emailElem, id) = findOrAddId(<input type="text" name="username"/>)
+    FocusOnLoadAppendJs(id)
     val bind =
-      ".email" #> FocusOnLoad(<input type="text" name="username"/>) &
+      ".email" #> emailElem &
       ".password" #> <input type="password" name="password"/> &
       "type=submit" #> loginSubmitButton(S.?("log.in"))
 
--- a/persistence/proto/src/main/scala/net/liftweb/proto/ProtoUser.scala
+++ b/persistence/proto/src/main/scala/net/liftweb/proto/ProtoUser.scala
@@ -864,10 +864,10 @@ trait ProtoUser {
       }
     }
 
-    val (emailElem, id) = findOrAddId(<input type="text" name="username"/>)
-    FocusOnLoadAppendJs(id)
+    val emailElemId = nextFuncName
+    S.appendJs(Focus(emailElemId))
     val bind =
-      ".email" #> emailElem &
+      ".email" #> <input type="text" name="username" id={emailElemId}/> &
       ".password" #> <input type="password" name="password"/> &
       "type=submit" #> loginSubmitButton(S.?("log.in"))
 
--- a/web/webkit/src/main/scala/net/liftweb/http/js/JsCommands.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/js/JsCommands.scala
@@ -703,7 +703,7 @@ object JsCmds {
    *
    * @return the element and a script that will give the element focus
    */
-  @deprecated("Use S.appendJs(Focus(id))","Since v3.0")
+  @deprecated("Use S.appendJs(Focus(id))","3.0.0")
   object FocusOnLoad {
     def apply(in: Elem): NodeSeq = {
       val (elem, id) = findOrAddId(in)
--- a/web/webkit/src/main/scala/net/liftweb/http/js/JsCommands.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/js/JsCommands.scala
@@ -697,34 +697,18 @@ object JsCmds {
   }
 
   /**
-   * Makes the parameter the selected HTML element on load of the page by appending 
-   * a script snippet to lifts page script that will give the element focus.
+   * Makes the parameter the selected HTML element on load of the page
    *
    * @param in the element that should have focus
    *
-   * @return the element 
+   * @return the element and a script that will give the element focus
    */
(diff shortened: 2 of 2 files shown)


Shadowfiend:
Sweet! Thanks for doing this! Going to run it through some tests this evening if I get a chance.

karma4u101:
Well this is a bit hackish the function (object) is going from not having side effects to having a side affect. 
Maybe a better solution would be to change the ProtoUser call to FocusOnLoad to some new calls, something like a call to something similar to findOrAddId (cant be done directly as it is protected) that just add a Id if not present and then have a new function using that id returning a JsCmd (with the focus code) and then wrap that return value in a S.appendJs call then just pass the id added transformed "email input node" to the css selector. I am not sure how that would play out and look like in the css selector in protouser login and where to put the new JsCmd function or what function to use instead of findOrAddId (something that just would add the id to the element if not present and return the element, maybe there already is one). 
This alternative "cleaner" solution dose however not "fix" the FocusOnLoad usage already out there in the wild.  
Any thoughts on this ?

karma4u101:
I did go ahead with some more updates.

If the new updates is in the overall right direction a question still remains on what to do about the FocusOnLoad object, should we (I) revert it to its original in-lining js functionality prior to this PR or should it stay as is possible with a "see" annotation pointing to FocusOnLoadAppendJs or would it be appropriate to add a deprecation notice on it? 
Is FocusOnLoadAppendJs a acceptable name for the new object.

@Shadowfiend or anyone else thoughts on this ?

eltimn:
I think we should probably just deprecate FocusOnLoad and I'm not sure FocusOnLoadAppendJs is necessary. Wouldn't it be simple enough to just do this:

    val (emailElem, id) = findOrAddId(<input type="text" name="username"/>)
    S.appendJs(Focus(id))
    val bind = ...

That's my 2 cents.

karma4u101:
@eltimn Thanks for your reply I have totally missed the Focus case class so in light of a focus function already existsing I totally agree about FocusOnLoadAppendJs not being needed. I will go ahead and update the PR in line with what you suggests.

eltimn:
After thinking about this some more, I would probably write it this way:

    val emailElemId = nextFuncName
    S.appendJs(Focus(emailElemId))
    val bind =
      ".email" #> <input type="text" name="username" id={emailElemId} />

No sense in calling `findOrAddId` with an element we know doesn't have an id.

karma4u101:
I think this is shaping up 
* Updated ProtoUser to use the the code Tim suggested above. 
* Reverting back FocusOnLoad (to its pre PR state) but added a deprecation annotation. 
* Removing newly (this PR) added FocusOnLoadAppendJs in favour of using the already existing Focus case class. 
Is everyone happy with this @Shadowfiend @eltimn

persistence/proto/src/main/scala/net/liftweb/proto/ProtoUser.scala
```diff
@@ -864,8 +864,10 @@ trait ProtoUser {
       }
     }
 
+    val emailElemId = nextFuncName
+    S.appendJs(Focus(emailElemId))
     val bind =
-      ".email" #> FocusOnLoad(<input type="text" name="username"/>) &
+      ".email" #> <input type="text" name="username" id={emailElemId}/> &
```
Shadowfiend: You could also just attach an `onclick` handler and let the attribute extraction stuff take care of it?

web/webkit/src/main/scala/net/liftweb/http/js/JsCommands.scala
```diff
@@ -703,13 +703,14 @@ object JsCmds {
    *
    * @return the element and a script that will give the element focus
    */
+  @deprecated("Use S.appendJs(Focus(id))","3.0.0")
```
Shadowfiend: 👌

persistence/proto/src/main/scala/net/liftweb/proto/ProtoUser.scala
karma4u101: I have tried some non working stuff combining onload with SHtml.onEvent and also with SHtml.ajaxInvoke calls but haven't got anything working that makes the email input get focus on page load. 
There is a very simple solution that works in html5 browsers (i.e autofocus) but I guess the point is to support older browsers also.
Could you elaborate a bit more on what you had in mind  ?

Shadowfiend: Yeah, I was thinking just making this:

```scala
      ".email" #> <input type="text" name="username" id={emailElemId} onclick="this.focus()" /> &
```

Or somesuch? Since this will go through the LiftMerge extractor, that
will take care of auto-assigning an `id` if needed, etc, etc.

Though it's also worth mentioning that it feels like we shouldn't be
replacing the whole `.email` element at all, but rather just binding
its `name` and possibly its `onclick`:

```scala
".email [name]" #> "username" &
".email [onclick]" #> "this.focus()" &
".password [name]" #> "password"
```

That's rather independent of this PR, though :)

eltimn: I don't think this is the same as calling `document.getElementById("#someid").focus()`

We want the focus to happen on page load, not when a user clicks on the element.

I do agree about the element, though, we should just be adding the id to the existing element.

Shadowfiend: Wow… Yeah, you're incredibly right, I don't know what I was thinking O.O

Well with that in mind, I think this is fine to go in as is and we can always
do another PR to improve how we actually bind the fields, yeah?

eltimn: Fine with me.

karma4u101: yea good points, as I am poking around in this code I will take a look at minimizing the node replacements.
 
Lets take a look at what I am (we are) trying to accomplish in this PR
**Goal:**
  1.  Avoid Inline JS, hens replace or modify the FocusOnLoad call used in ProtoUser.
  2.  Somehow preserve the original behaviour although we modify or remove the use of FocusOnLoad.

**The original behaviour** that we try to preserve is in short  
User clicks the "Login" menu and the page loads and sets focus on the email input element so that **without any future user interaction** (click or key strokes) the user can start typing his/here email address in the email input field. 

**The road so far:**
I decided that my original approacher to modify the FocusOnLoad object was a bad idea case it introduced a unnecessary side effect and Tim showed a solution that I agreed was a less intrusive solution than the one I had come up with. 
Using the solution Tim suggested dose satisfied the goals (1,2) stated above but the suggestions you made to "let the attribute extraction stuff take care of it" (lets call this goal 3) is appealing so while we are at it we should take a stab on it. 
The solutions I have been trying has been unsuccessful and unfortunately the suggestion to use onclick "this.focus()" dose not satisfy this goal (3) either.  

@Shadowfiend  (or anyone else) any ideas ?

Shadowfiend: I don't actually think using attribute extraction is an option here, as there's no event attribute we can attach this behavior to. I mean, we can always update the attribute extraction code to see `autofocus` and turn it into JS, but that's a counter-goal IMO.

In short, I think the current solution you have is correct :)

karma4u101:
@Shadowfiend okey cool 
As I was at it anyway I did go ahead and adjusted the bindings, I have tested against the lift 3.0 lift_basic template app and everything works as expected.

Shadowfiend:
Wooorks for me! I'll give it a whirl against `lift_basic` as well and then merge this, probably tomorrow.

karma4u101:
Great  :thumbsup:

Shadowfiend:
Confirmed, let's do this!

PR #1744, opened 2015-12-12
Has Pwn Property: Fix hasOwnProperty use in IE.

### LINKED ISSUE #1743: exception in IE8 in lift-3.0, `hasOwnProperty` is undefined
An error is being thrown when I try to use lift-3.0 with IE8.
Details as shown by IE8 browser:

```
lift.js  line:1 char:4618 code:0  `hasOwnProperty` is undefined.
```

screenshot (time to live 1 year): http://pointsgame.net/vn971/temp/2015.12.09_19:50:15_6803739.png

Relevant ML topic name: "[lift 3.0-M7 regression](https://groups.google.com/d/msg/liftweb/0XhyizAdJIQ/O2_0SvRPBwAJ)"

Shadowfiend:
There was code that referenced a nonexistent hasOwnProperty variable, which we
now introduce, and we change the one other usage of hasOwnProperty to also use
this variable. This should generally be IE-safe---the core issue is that IE 8
does not have hasOwnProperty on native objects like Array, which can cause
things to go boom boom.

Fixes #1743.

### DIFF
--- a/web/webkit/src/main/resources/toserve/lift.js
+++ b/web/webkit/src/main/resources/toserve/lift.js
@@ -619,7 +619,7 @@
       },
       extend: function(obj1, obj2) {
         for (var item in obj2) {
-          if (hasOwnProperty.call(obj2, item) {
+          if (hasOwnProperty.call(obj2, item)) {
             obj1[item] = obj2[item];
           }
         }
--- a/web/webkit/src/main/resources/toserve/lift.js
+++ b/web/webkit/src/main/resources/toserve/lift.js
@@ -1,6 +1,8 @@
 (function(window, undefined) {
   "use strict";
 
+  var hasOwnProperty = Object.prototype.hasOwnProperty;
+
   window.lift = (function() {
     // "private" vars
     var ajaxPath = function() { return settings.liftPath + '/ajax' },
@@ -616,8 +618,8 @@
         }
       },
       extend: function(obj1, obj2) {
-        for(var item in obj2) {
-          if (obj2.hasOwnProperty(item)) {
+        for (var item in obj2) {
+          if (hasOwnProperty.call(obj2, item) {
             obj1[item] = obj2[item];
           }
         }


Shadowfiend:
@vn971 any chance you could give this branch a whirl and make sure it fixes your issue?

vn971:
@Shadowfiend Sure, I think I'll try today.

vn971:
Thanks!

vn971:
It seems rather fixed than not. It throws a different Error now:

`window.lift is null or not an object
line: 3  char: 1 code: 0  URI: ***/lift/page/F65*******.js
`

The line in question seems to be:
`window.lift.extend(lift_settings,window.liftJQuery);`


Should I raise a separate ticket?

vn971:
Ouch, it's not actually in IE8 only, it's in firefox+Linux, too. I try to understand whether I misconfigured something, but I get:

```
16:19:59.778 GET 
http://webai.localhost:4141/classpath/lift.js [HTTP/1.1 404 Not Found 3ms]
16:19:59.965 GET 
http://webai.localhost:4141/classpath/lift.js [HTTP/1.1 404 Not Found 3ms]
16:20:00.150 TypeError: window.lift is undefined F652205397165VVZVK4.js:3:1
```

What's surprising is that lift.js is not found. Did I mess something up?..

vn971:
updated^

vn971:
I can raise an ML topic, of course, if that would be preferable.

vn971:
Note that I cannot test this fix (`has-pwn-property`) because I can not rebase above to 3.0-M6.

There is no 3.0-M6 git tag in the git repository.

Shadowfiend:
There is: https://github.com/lift/framework/tree/3.0-M6-release :)

I'll give this a quick whirl in the browsers I have today, probably got
a tad overconfident at 2am :)

eltimn:
I'm wondering if we'd be better off using `jQuery.extend` when it's available and only use the custom extend function with liftVanilla. I'm still trying to figure out how we could do that.

Using `liftVanilla` is really only practical with newer browsers, so there shouldn't be a problem there.

vn971:
@Shadowfiend sorry, you're right. I forgot to add `--tags` to `git fetch`.

Shadowfiend:
< shrug > I don't think it's much of a big deal, tbh, @eltimn . The particular issue here was I forgot a paren >_>

Shadowfiend:
Give that a shot when you get a chance @vn971… It should *actually* work this time <_< >_>

eltimn:
:ok_hand:

vn971:
@Shadowfiend awesome, it fully works now!

Shadowfiend:
:tada:

Will give a day or two to see if we get some feedback, but then going to merge this sucker in.

vn971:
@Shadowfiend is it good to leave it as-is? I mean, M7 is published already, some people (like me) might discover it using `sbt-updates` plugin.
Maybe delete the published M7 as an artifact?

Shadowfiend:
Please bring that up on the list and we can discuss there.

Shadowfiend:
Seeendin' 'er in!

PR #1746, opened 2015-12-15
Add 'X-Requested-With' header to liftVanilla

### LINKED ISSUE #1745: Add 'X-Requested-With' header to liftVanilla
Lift checks for that header in `Req.ajax_?` to determine if the request is ajax.

eltimn:
fixes #1745 

I added a way to run JavaScript tests, but they need to be run in a browser manually by loading `web/webkit/SpecRunner.html`

Here's the output:

screenshot from 2015-12-15 15 42 11


eltimn:
I went ahead and added an npm script to run the tests and jshint since we're using travis. You can still run the tests manually with a browser so you don't need to install node to run them.

@Shadowfiend @farmdawgnation is there a specific reason there's no `language: scala` entry in travis.yml? Without it travis assumes it's a ruby environment and installs ruby. It doesn't seem to hurt anything, but seems wasteful.

Shadowfiend:
Re: node, is there any reason we can't just use Nashorn given that we've imposed build-using-Java-8-minimum semantics? I realize we're currently building with Java 7 until we release M7, but it seems worth it to keep the build depending solely on Java and Java package resolution (via webjars?).

web/webkit/src/main/resources/toserve/lift.js
```diff
@@ -463,15 +463,15 @@
...
-        if (evt.done != null) {
+        if (evt.done !== null) {
           doneMsg();
         }
-        else if (evt.success != null) {
+        else if (evt.success !== null) {
           successMsg(evt.success);
         }
-        else if (evt.failure != null) {
+        else if (evt.failure !== null) {
```
Shadowfiend: We sure these three don't break anything? `undefined != null` is false while `undefined !== null` is true…

web/webkit/src/main/resources/toserve/lift.js
```diff
@@ -745,7 +745,9 @@
           if (xhr.status === 200) {
             if (dataType === "script") {
               try {
+                /* jshint ignore:start */
                 eval(xhr.responseText);
+                /* jshint ignore:end */
```
Shadowfiend: I don't suppose there's a way to indicate this without polluting the file itself?

web/webkit/src/main/resources/toserve/lift.js
```diff
@@ -837,6 +844,7 @@
       xhr.open("GET", url, true);
       xhr.timeout = settings.cometGetTimeout;
       xhr.setRequestHeader("Accept", "text/javascript, application/javascript, application/ecmascript, application/x-ecmascript, */*; q=0.01");
+      xhr.setRequestHeader("X-Requested-With", "XMLHttpRequest"); // So Req.ajax_? will return true
```
Shadowfiend: IMHO, we can leave off the comment about `Req.ajax_?` in both places it exists. Can keep it if you feel strongly otherwise, though.

Shadowfiend:
Also, using webjars would probably allow us to not import jasmine directly into the repo, etc etc.

Oh, and re: language in the travis file: I don't think there's a reason, but not sure.

web/webkit/src/main/resources/toserve/lift.js
eltimn: No. I'll revert it. I changed a couple of other ones that I'll revert as well.

web/webkit/src/main/resources/toserve/lift.js
eltimn: I think the only way is to turn off eval checking globally, which is probably fine. It's likely whoever will be editing this knows not to use eval unless it's necessary.

eltimn:
Part of the difficultly with testing is the way we set lift, liftJQuery, and liftVanilla as global variables. That's why I went the browser and phantomjs route. I couldn't figure out a way to get the file loaded to be available for the tests.

What I'd like to do is rewrite it using ES6 modules and transpile them into ES5 and requirejs modules. That would make it easier to do testing.

While we could use Nashorn directly, it's probably easier to just use sbt-web and plugins. We'd need that for the webjars anyway.

So, my thinking was with the node stuff at least we have some way to write tests for the JavaScript code.

I could add sbt-web to get the dependencies, but without rewriting the js code, I don't see any other way to have tests for it.

Shadowfiend:
Hmmm… I see. That definitely sounds like a bigger hunk of work, and I think this is progress, so I'm game!

eltimn:
Rhino in yuicompressor conflicts with the Rhino version in sbt-web. So running any sbt-web tasks doesn't work. It's installed only to download the webjar dependencies.

I tried to include yuicompressor intransitively in sbt-yiucompressor and use a newer version of rhino, but it wouldn't compile that way.

It would be nice if we could get rid of all of those old JavaScript libraries we have in toserve.

I decided to leave the jshint comments in the js source as that's really the best way. I did change the eval ones to a single line, but had to add some for the `!= null` ones.

If this looks good, I'll squash my commits and issue a new pull request.

README.md
```diff
@@ -145,4 +147,4 @@ changes to Lift. We do, however, accept some small changes and bugfixes into Lif
 
 ## Continuous Integration
 
-SNAPSHOTs are built at CloudBees: https://lift.ci.cloudbees.com/
+SNAPSHOTs are built by [Travis CI](https://travis-ci.org/lift/framework)
```
Shadowfiend: Oh yeah good catch!

README.md
```diff
@@ -1,5 +1,7 @@
 # The Lift Web Framework
 
+[![Build Status](https://travis-ci.org/lift/framework.svg?branch=master)](https://travis-ci.org/lift/framework)
```
Shadowfiend: We [threw around this idea a while ago](https://groups.google.com/d/msg/liftweb/hHzK9lgXyKs/6ouEvww0LfwJ) and weren't toooo thrilled at the potential for a misfailed build to be sitting on master.

That said, I think the work @farmdawgnation did has made a passing build an overwhelming likelihood as long as everything is working as it should, so I'm down for adding this now.

.travis.yml
```diff
@@ -1,7 +1,7 @@
-sudo: false
+language: scala
 cache:
   directories:
-    - '$HOME/.m2/repository'
+    - '$HOME/node_modules'
```
Shadowfiend: Should we add `$HOME/.ivy`?

travis.sh
```diff
@@ -4,6 +4,11 @@ set -ev
 
 ./liftsh test
 
+cd web/webkit
+npm run lint
+npm run test
+cd -
```
Shadowfiend: We've had this issue before at Elemica—if `./liftsh test` fails but `npm run test` passes, will this properly mark the build as failed?

web/webkit/SpecRunner.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <title>Jasmine Spec Runner v2.4.0</title>
```
Shadowfiend: Do we still need this given the work to run things via `npm`?

.travis.yml
eltimn: I tried that and it didn't seem to help, but there's probably something we can do to cache the dependencies. Although, I'm not sure it matters all that much as this build doesn't run all that often.

travis.sh
eltimn: Since `set -e` is used in the script it should stop and exit on the first non-zero return code.

web/webkit/SpecRunner.html
eltimn: I figured if a developer doesn't want to install node locally they can still run the tests through the browser. But, if you don't think that's necessary, I can remove it.

travis.sh
Shadowfiend: Got it, got it.

.travis.yml
Shadowfiend: Yeah, no big in the short term agreed. Cool.

web/webkit/SpecRunner.html
Shadowfiend: I think it's fair to say that node is needed to run JS tests. Ideally we'll get to a place where we can make that no longer true, but until then that's fine IMO. Your call, ultimately.

web/webkit/src/main/resources/toserve/lift.js
```diff
@@ -680,7 +685,7 @@
         success: onSuccess,
         error: function(_, status) {
           if (status != 'abort')
-            return onFailure.apply(this, arguments)
+            return onFailure.apply(this, arguments);
```
Shadowfiend: Codacy flagged this as being better with curlies around the if body, I tend to agree.

Shadowfiend:
Yeah, this is looking good. I'm down for a squash+merge when you're comfortable with it :)

eltimn:
New PR at #1748

PR #1748, opened 2015-12-26
Add 'X-Requested-With' header to liftVanilla

### LINKED ISSUE #1745: Add 'X-Requested-With' header to liftVanilla

eltimn:
Fixes #1745

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,6 @@ project/sbt-launch*
 # Doc generation
 .dexy
 core/documentation-helpers/src/test/**Test.scala
+
+# JavaScript testing
+node_modules
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
-sudo: false
+language: scala
 cache:
   directories:
-    - '$HOME/.m2/repository'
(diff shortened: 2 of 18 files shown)


Shadowfiend:
This all looks pretty swell to me. :+1: though probably the one issue that Codacy identified (used `!=` instead of `!==`) could be addressed.

eltimn:
There are a bunch of places in lift.js that codacy is still reporting issues. Although, it only seems to complain when you touch the code where the problem is.

I will add a `.jshintrc` file so jshint will pick them up better. I figured it would be better to do that as a separate commit, so it's just the formatting changes.

Shadowfiend:
This looks 100% good to me. If you want to add `jshintrc` that's fine, but if not I'm down with that as well and happy to merge as-is.

farmdawgnation:
:+1:

Shadowfiend:
There's a merge conflict here, but other than that this is looking good to me.

Shadowfiend:
Let's do eeet!

PR #1749, opened 2015-12-26
Upgrade mongo-java-driver to v2.14.0

### LINKED ISSUE #1747: Investigate upgrading MogoDB dependencies

eltimn:
Fixes #1747 

We can't upgrade to the 3 series until we remove a few functions, which I've deprecated in this commit. This version is compatible with the latest server versions, so it should be sufficient for a while longer.

### DIFF
--- a/persistence/mongodb-record/src/main/scala/net/liftweb/mongodb/record/field/DBRefField.scala
+++ b/persistence/mongodb-record/src/main/scala/net/liftweb/mongodb/record/field/DBRefField.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010-2011 WorldWide Conferencing, LLC
+ * Copyright 2010-2015 WorldWide Conferencing, LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -64,7 +64,7 @@ class DBRefField[OwnerType <: BsonRecord[OwnerType], RefType <: MongoRecord[RefT
 
   def asXHtml = <div></div>
 
-  def defaultValue = new DBRef(null, null, null)
+  def defaultValue = new DBRef("", null)
 
   def setFromAny(in: Any): Box[DBRef] = in match {
     case ref: DBRef => Full(set(ref))
@@ -81,13 +81,11 @@ class DBRefField[OwnerType <: BsonRecord[OwnerType], RefType <: MongoRecord[RefT
   // assume string is json
   def setFromString(in: String): Box[DBRef] = {
     val dbo = JSON.parse(in).asInstanceOf[BasicDBObject]
-    MongoDB.use(ref.meta.connectionIdentifier) ( db => {
-      val id = dbo.get("$id").toString
-      ObjectId.isValid(id) match {
-        case true => Full(set(new DBRef(db, dbo.get("$ref").toString, new ObjectId(id))))
-        case false => Full(set(new DBRef(db, dbo.get("$ref").toString, id)))
-      }
-    })
+    val id = dbo.get("$id").toString
+    ObjectId.isValid(id) match {
+      case true => Full(set(new DBRef(dbo.get("$ref").toString, new ObjectId(id))))
+      case false => Full(set(new DBRef(dbo.get("$ref").toString, id)))
+    }
   }
 
   def toForm: Box[NodeSeq] = Empty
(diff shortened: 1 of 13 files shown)


persistence/mongodb-record/src/main/scala/net/liftweb/mongodb/record/field/DBRefField.scala
```diff
@@ -81,13 +81,11 @@ class DBRefField[OwnerType <: BsonRecord[OwnerType], RefType <: MongoRecord[RefT
...
-      ObjectId.isValid(id) match {
-        case true => Full(set(new DBRef(db, dbo.get("$ref").toString, new ObjectId(id))))
-        case false => Full(set(new DBRef(db, dbo.get("$ref").toString, id)))
-      }
-    })
+    val id = dbo.get("$id").toString
+    ObjectId.isValid(id) match {
+      case true => Full(set(new DBRef(dbo.get("$ref").toString, new ObjectId(id))))
+      case false => Full(set(new DBRef(dbo.get("$ref").toString, id)))
+    }
```
Shadowfiend: So can this no longer fail? I see we're going to a `Full` box always, but we weren't before I guess?

persistence/mongodb-record/src/main/scala/net/liftweb/mongodb/record/field/MongoCaseClassField.scala
```diff
@@ -107,6 +107,7 @@ class MongoCaseClassListField[OwnerType <: Record[OwnerType],CaseType](rec: Owne
     asJValue match {
       case JArray(list) =>
         list.foreach(v => dbl.add(JObjectParser.parse(v.asInstanceOf[JObject])))
+      case _ =>
```
Shadowfiend: :tada:

persistence/mongodb-record/src/test/scala/net/liftweb/mongodb/record/MongoClientSaveSpec.scala
```diff
@@ -64,7 +64,7 @@ class MongoClientSaveSpec extends Specification with MongoTestKit {
     sd2.save(false) // no exception thrown
     sd2.save(true) must throwA[MongoException]
     sd2.saveBox() must beLike {
-      case Failure(msg, _, _) => msg must contain("E11000 duplicate key error index") // exception thrown
+      case Failure(msg, _, _) => msg must contain("E11000 duplicate key error") // exception thrown
```
Shadowfiend: Should we just check for the error code since it's the bit that's likely to stay stable?

persistence/mongodb/src/main/scala/net/liftweb/mongodb/Mongo.scala
```diff
@@ -126,6 +127,7 @@ object MongoDB {
     * and the use of getLastError.
     * See: http://docs.mongodb.org/ecosystem/drivers/java-concurrency/
     */
+  @deprecated("No longer relevant. See mongo-java-drivers's JavaDocs for details", "3.0")
```
Shadowfiend: When we say it's not “relevant”, do we mean you don't need to wrap your code in this? If so, perhaps we can say “No longer needed” instead?

persistence/mongodb-record/src/main/scala/net/liftweb/mongodb/record/field/DBRefField.scala
eltimn: We used to have to fetch the DB, but DBRef no longer supports that. So, yeah, it will always be`Full` now. Not really sure how else to handle this.

persistence/mongodb/src/main/scala/net/liftweb/mongodb/Mongo.scala
eltimn: Correct. `Acknowledged` writes are now the preferred way to do this and so Mongo is deprecating this functionality.

"No longer needed" is probably better. I'll update it.

persistence/mongodb-record/src/test/scala/net/liftweb/mongodb/record/MongoClientSaveSpec.scala
eltimn: That's probably a good idea.

persistence/mongodb-record/src/main/scala/net/liftweb/mongodb/record/field/DBRefField.scala
Shadowfiend: Yeah, I think this is fine, just wanted to double-check.

Shadowfiend:
Looks good to me!

PR #1750, opened 2015-12-26
Use S.originalRequest in Menu snippet

### LINKED ISSUE #1695: Menu.item evaluation in AJAX requests
https://github.com/lift/framework/blob/master/web/webkit/src/main/scala/net/liftweb/builtin/snippet/Menu.scala#L443

For AJAX requests, there is no such Loc. S.originalRequest.map(_.location) meaning that Menu.item will never be correctly evaluated, S.originalRequest.flatMap(_.location) would probably be better in the snippet. Tim Nelson tested this and it does indeed fix the problem.

Mailing list thread: https://groups.google.com/forum/#!topic/liftweb/z9L8p7gKO64

eltimn:
Fixes #1695

### DIFF
--- a/web/webkit/src/test/scala/net/liftweb/builtin/snippet/MenuSpec.scala
+++ b/web/webkit/src/test/scala/net/liftweb/builtin/snippet/MenuSpec.scala
@@ -17,8 +17,7 @@
 package net.liftweb
 package builtin.snippet
 
-import xml._
-import org.specs2.matcher.XmlMatchers
+import scala.xml._
 import org.specs2.mutable.Specification
 
 import common._
@@ -27,12 +26,9 @@ import mockweb._
 import MockWeb._
 import mocks._
 import sitemap._
-import util.Helpers.{randomString, secureXML}
+import util.Helpers.randomString
 
-/**
- * System under specification for Menu.
- */
-object MenuSpec extends Specification with XmlMatchers {
+object MenuSpec extends Specification {
   "Menu Specification".title
 
   case class Param(s: String)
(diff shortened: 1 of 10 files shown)


web/webkit/src/test/scala/net/liftweb/builtin/snippet/MenuSpec.scala
```diff
@@ -0,0 +1,122 @@
...
+import http._
+import mockweb._
+import MockWeb._
+import mocks._
+import sitemap._
+import util.Helpers.{randomString, secureXML}
+
+/**
+ * System under specification for Menu.
+ */
```
Shadowfiend: This seems like a pretty redundant comment? Might be auto-generated though, dunno.

eltimn: I agree. Those are in a lot if not all of our tests. Not sure where they came from.

Shadowfiend:
Took a quick look at trying to figure out how to properly mock `S.originalRequest`, was also unsuccessful. Probably worth keeping in the back of our minds to see if we can make that clearer in the future, but for now this looks good.

Looks like the travis build blew up in a variety of ways, going to rerun it and see if it's branch-related or not.

eltimn:
@Shadowfiend the tests fail due to setting SiteMap being set. I'm almost done with a solution.

Shadowfiend:
Got it!

eltimn:
I had to duplicate some of the mongo changes to get some tests passing.

Shadowfiend:
This looks good to me; can probably ask for feedback from the original reporter as well once we merge!

PR #1751, opened 2015-12-29
Ajax Event Extraction: Allow event extraction in AJAX responses

### LINKED ISSUE #1700: Inline JS (in onclick) is not replaced in AJAX-response
Ref this thread
https://groups.google.com/d/msg/liftweb/b6lUWIDOSiY/_6dxm2sDo64J

With the new mechanism in place to replace code like this:
 
```html
<span onclick="cranks()">
```
 
with
 
```html
<span id="lift-event-js-F8991755608893YMP2K">
```
 
it doesn't seem to apply for the same code being served as AJAX-response, meaning the same code will get the original onclick-markup when being served as AJAX.

Shadowfiend:
Abstract out extraction of HTML event attributes into JS via
a common `HtmlNormalizer` singleton that also handles URL
normalization (prefixing of context path, rewriting, etc). This
extracts some functionality that was mostly in `LiftMerge`, and
allows for extensible invocations that let `LiftMerge` do the
additional things it needs to do to properly render a full page
(like HTML head and tail merging).

The current code is decent but not final, and there are currently
no tests. I'll be going back and writing tests for `LiftMerge`'s
original behavior to confirm that there were no regressions, though
of course that's always hard to do after the fact.

Would appreciate a first pass at what's here now—I've improved
verbiage throughout and things should hopefully be easier to follow
than they were before.

This fixes #1700.

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/HtmlNormalizer.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/HtmlNormalizer.scala
@@ -257,7 +257,7 @@ private[http] final object HtmlNormalizer {
                     normalizeHtmlAndEventHandlers(element.child, contextPath, stripComments, nextState, additionalChanges)
 
                   (
-                    normalizedNodes.+:[Node,NodeSeq](element.copy(child = newChildren)),
+                    normalizedNodes.:+[Node,NodeSeq](element.copy(child = newChildren)),
                     extractedJs & elementJsCmds & additionalJsCmds & childJsCmds
                   )
 
@@ -277,13 +277,13 @@ private[http] final object HtmlNormalizer {
                   additionalChanges
                 )
 
-              (normalizedNodes.+:[Node,NodeSeq](Group(normalizedGroupNodes)), extractedJs & js)
+              (normalizedNodes.:+[Node,NodeSeq](Group(normalizedGroupNodes)), extractedJs & js)
 
             case c: Comment if stripComments =>
               (normalizedNodes, extractedJs)
 
             case _ =>
-              (normalizedNodes.+:[Node,NodeSeq](nodeToNormalize): NodeSeq, extractedJs)
+              (normalizedNodes.:+[Node,NodeSeq](nodeToNormalize): NodeSeq, extractedJs)
           }
       }): (NodeSeq, JsCmd)
     }
--- a/web/webkit/src/main/scala/net/liftweb/http/HtmlNormalizer.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/HtmlNormalizer.scala
@@ -1,0 +1,291 @@
+package net.liftweb
+package http
+
+import scala.xml._
+import scala.collection.immutable.Vector
+
+import js.JsCmd
+import js.JsCmds.Noop
+import js.JE.{AnonFunc, Call, JsRaw}
+
+import util.Helpers._
+
+/**
+ * Represents an HTML attribute for an event handler. Carries the event name and
+ * the JS that should run when that event is triggered as a String.
+ */
+private case class EventAttribute(eventName: String, jsString: String)
+private object EventAttribute {
+  /**
+   * Some elements allow a URL attribute to take a `javascript:(//)`-style
+   * URL instead of setting an `on*` event in order to invoke JS. For example,
+   * you can (and Lift does) set a form's `action` attribute to
+   * `javascript://(some JS)` instead of setting `onsubmit` to `(someJS);
+   * return false`.
+   *
+   * This is a map from those attribute names to the corresponding JS event
+   * that would be used to execute that JS when it isn't run in line.
+   */
+  val eventsByAttributeName =
+    Map(
+      "action" -> "submit",
+      "href" -> "click"
+    )
+
+  object EventForAttribute {
+    def unapply(attributeName: String): Option[String] = {
+      eventsByAttributeName.get(attributeName)
+    }
+  }
+}
+
+/**
+ * Helper class that performs certain Lift-specific normalizations of HTML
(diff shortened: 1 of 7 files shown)


web/webkit/src/main/scala/net/liftweb/http/HtmlNormalizer.scala
```diff
@@ -0,0 +1,291 @@
+package net.liftweb
+package http
```
farmdawgnation: I'm slowly becoming skeptical of this package organization. We're stuffing a bunch of things in a package named `http` that doesn't deal with the HTTP protocol, which isn't thrilling.

Food for thought.

web/webkit/src/main/scala/net/liftweb/http/HtmlNormalizer.scala
```diff
@@ -0,0 +1,291 @@
...
+import js.JsCmds.Noop
+import js.JE.{AnonFunc, Call, JsRaw}
+
+import util.Helpers._
+
+/**
+ * Represents an HTML attribute for an event handler. Carries the event name and
+ * the JS that should run when that event is triggered as a String.
+ */
+private case class EventAttribute(eventName: String, jsString: String)
```
farmdawgnation: Is this an appropriate point to have Javascript as a `String` instead of as a `JValue`?

Not saying it isn't, mind, I just don't know this code well enough to know myself and it always seems like a question worth asking. :-)

web/webkit/src/main/scala/net/liftweb/http/HtmlNormalizer.scala
```diff
@@ -0,0 +1,291 @@
...
+  //  - The optional id that was found in this set of attributes.
+  //  - The normalized metadata.
+  //  - A list of extracted `EventAttribute`s.
+  private def normalizeUrlAndExtractEvents(
+    attributeToNormalize: String,
+    attributes: MetaData,
+    contextPath: String,
+    shouldRewriteUrl: Boolean, // whether to apply URLRewrite.rewriteFunc
+    eventAttributes: List[EventAttribute] = Nil
+  ): (Option[String], MetaData, List[EventAttribute]) = {
```
farmdawgnation: If this a candidate for returning a `case class`?

web/webkit/src/main/scala/net/liftweb/http/HtmlNormalizer.scala
```diff
@@ -0,0 +1,291 @@
...
+          shouldRewriteUrl,
+          eventAttributes
+        )
+
+      attributes match {
+        case attribute @ UnprefixedAttribute(
+               EventAttribute.EventForAttribute(eventName),
+               attributeValue,
+               remainingAttributes
+             ) if attributeValue.text.startsWith("javascript:") =>
```
farmdawgnation: Do we want to limit the elements that we will handle `javascript://` on instead of handling it anywhere?

farmdawgnation:
Got a little under halfway through before my eyes started glazing over :P Will try to make another pass later. haha.

web/webkit/src/main/scala/net/liftweb/http/HtmlNormalizer.scala
Shadowfiend: The package might be better named `web`, but I think it's extremely nice from a framework usability perspective that you can import `net.liftweb.http._` and be good to go on all non-AJAX stuff (and some AJAX stuff as well).

Shadowfiend: (Which isn't to say you're wrong, just saying that's a nice byproduct and as we consider the issue you've brought up, it's worth keeping that aspect in mind.)

web/webkit/src/main/scala/net/liftweb/http/HtmlNormalizer.scala
Shadowfiend: Yeah, this is once it's extracted directly from HTML (I assume you meant `JsCmd` or `JsExp` rather than `JValue`), so there's no real need to reify it into a typed thing yet.

web/webkit/src/main/scala/net/liftweb/http/HtmlNormalizer.scala
Shadowfiend: It could be. I vacillated a bit on that front but basically couldn't come up with a proper name for the things it was doing. That's because it's currently doing 3 semi-independent things (extracting id, normalizing URLs, and extracting event attributes). On its face, this seems like a bad thing (separation of concerns and such), but because `MetaData` is a linked list, the quickest way to do this is recursively starting at the beginning—and because this is run as part of the core HTML rendering pipeline I didn't want to get too crazy on potentially sacrificing performance for clarity.

Anyway, since the method itself is private and is only used in one place (where the sub-parts are immediately unpacked on assignment), I settled on a tuple + a comment.

web/webkit/src/main/scala/net/liftweb/http/HtmlNormalizer.scala
Shadowfiend: For sure, and we do. The mechanism for doing that is the [`EventForAttribute` extractor](https://github.com/lift/framework/blob/ajax-event-extraction/web/webkit/src/main/scala/net/liftweb/http/HtmlNormalizer.scala#L35-L39), since in cases where we detect a `javascript:[//]` URL we *also* need to drop that attribute and replace it with the event attribute (e.g. `href="javascript:alert('hello');"` needs to lose the `href` and add an `onclick`).

Shadowfiend:
Sadly the current solution completely breaks `LiftMerge`. Fortunately, I now have specs for `LiftMerge`. Will open a separate PR for this.

PR #1752, opened 2016-01-10
85th Normal Form: Allow event extraction in AJAX responses.

### LINKED ISSUE #1700: Inline JS (in onclick) is not replaced in AJAX-response

Shadowfiend:
Abstract out extraction of HTML event attributes into JS via
a common `HtmlNormalizer` singleton that also handles URL
normalization (prefixing of context path, rewriting, etc). This
extracts some functionality that was mostly in `LiftMerge`, and
allows for extensible invocations that let `LiftMerge` do the
additional things it needs to do to properly render a full page
(like HTML head and tail merging).

The merging functionality in `LiftMerge` is now covered by a test, as
is the new `HtmlNormalizer` that acts as a common spot for HTML
normalization that `LiftMerge` and AJAX stuff both need to do. I'm
satisfied with this approach, so I'm down to hear any and all criticism
of it as it currently stands.

Of note: things are a little different than they were in the first
PR; in particular, `HtmlNormalizer` is now structured so you can
call its sub-parts and do stuff around them, which is what `LiftMerge`
does, instead of having a callback-based extensibility approach.

This fixes #1700.

### DIFF
--- a/core/util/src/main/scala/net/liftweb/util/ListHelpers.scala
+++ b/core/util/src/main/scala/net/liftweb/util/ListHelpers.scala
@@ -238,10 +238,7 @@ trait ListHelpers {
   }
 
   /** Add utility methods to Lists */
-  implicit def toSuperList[T](in: List[T]): SuperList[T] = new SuperList(in)
-
-  /** Add utility methods to Lists */
-  class SuperList[T](val what: List[T]) {
+  implicit class SuperList[T](what: List[T]) extends AnyRef {
     /** permute the elements of a list */
     def permute = permuteList(what)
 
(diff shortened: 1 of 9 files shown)


Shadowfiend:
Just added specs for `HtmlNormalizer`, so I think this is complete for me and ready for full review!

farmdawgnation:
`+1,422 −232`

/me faints

Mayyyyy have to save this for the weekend. :P

Shadowfiend:
900 lines of tests though 🎉 \>_>

Shadowfiend:
Going to give this another day or so and then merge 'er.

Shadowfiend:
Time for ze games.

PR #1754, opened 2016-01-12
Relax Max: Make Content-Security-Policy report-only in production

### LINKED ISSUE #1753: Default CSP to report-only for Lift 3.0.0
As a transitory step, Lift 3.0.0 should only report CSP violations,
not enforce them, even in production. A slight deviation from “secure
by default”, but it should make the transition to Lift 3 more straightforward.
Lift 3.1.0 will move us to a production default of enforced security policies.

Shadowfiend:
This is done so that transition to Lift 3.0.0 will be less
bumpy. See [the associated mailing list discussion](https://groups.google.com/forum/#!searchin/liftweb/CSP%7Csort:date/liftweb/2lDoWpT6ufk/166IsbPjBAAJ) for
more. I also added a spec for `SecurityRules`, although
it doesn't quite cover everything.

Fixes #1753 .

### DIFF
--- a/web/webkit/src/test/scala/net/liftweb/http/SecurityRulesSpec.scala
+++ b/web/webkit/src/test/scala/net/liftweb/http/SecurityRulesSpec.scala
@@ -1,0 +1,238 @@
+/*
+ * Copyright 2011 WorldWide Conferencing, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package net.liftweb
+package http
+
+import java.net.URI
+
+import scala.concurrent.duration.{Duration,SECONDS,DAYS}
+
+import org.specs2.mutable.Specification
+
+class HttpsRulesSpec extends Specification {
+  "HttpsRules" should {
+    "default to no required HTTPS" in {
+      HttpsRules().headers must be empty
+    }
+
+    "provide a secure variant with 1-year max age and including sub-domains" in {
+      HttpsRules.secure.requiredTime must_== Some(Duration(365, DAYS))
+      HttpsRules.secure.includeSubDomains must_== true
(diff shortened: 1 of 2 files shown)


web/webkit/src/test/scala/net/liftweb/http/SecurityRulesSpec.scala
```diff
@@ -0,0 +1,238 @@
+/*
+ * Copyright 2011 WorldWide Conferencing, LLC
```
eltimn: Should probably change this to 2016.

eltimn:
:+1 other than the copyright year

web/webkit/src/test/scala/net/liftweb/http/SecurityRulesSpec.scala
```diff
@@ -0,0 +1,238 @@
...
+      ContentSecurityPolicy().reportUri must_== Some(ContentSecurityPolicy.defaultReportUri)
+    }
+
+    "provide [X-]Content-Security-Policy if enforcement is enabled" in {
+      ContentSecurityPolicy()
+        .headers(enforce = true, logViolations = true)
+        .collect {
+          case (headerName, _) if headerName.contains("Content-Security-Policy") =>
+            headerName
+        } must_== List("Content-Security-Policy", "X-Content-Security-Policy")
```
farmdawgnation: It might be better to use `must contain` on the list this `collect` generates than asserting over the entire list at once. The errors in this case (when they occur) are a bit hard to parse out as a human and we don't _really_ care about the ordering for this to be valid.

farmdawgnation:
Overall, this looks good to me. :+1:

web/webkit/src/test/scala/net/liftweb/http/SecurityRulesSpec.scala
Shadowfiend: Because this is security stuff, I'm trying to be a little more specific about checking that it includes these and only these values and headers. Maybe there's a verified that checks that all contents are exactly the same but not necessarily in order (`must contain` doesn't fail if there are unexpected things, iirc). 

That said, the error messages aren't particularly hard to parse imo: you get both the result and the expectation printed out (plus that weird string diff, but that's not useful 99% of the time anyway), right?

web/webkit/src/test/scala/net/liftweb/http/SecurityRulesSpec.scala
Shadowfiend: I'll try and do that. This is why I strongly dislike these headers... so many of them are out of date anyway. < grump grump grump >

Shadowfiend:
Kk, fixed the copyright date so I think I'm just going to go ahead and send this in.

PR #1756, opened 2016-01-25
Server comm failure callbacks

joescii:
In this PR I added front-end hooks for applications to be notified of possible server communication problems.  There are 4 new public methods on `lift`:

```javascript
lift.addAjaxOnCommFailure()
lift.addAjaxOnCommSuccess()
lift.addCometOnCommFailure()
lift.addCometOnCommSuccess()
```

With these in place, it is possible for an app to show the user that server communication has potentially been lost after some threshold number of failure events for instance.

See relevant [ML thread](https://groups.google.com/forum/#!topic/liftweb/0z6SV0ucEW4).

### DIFF
--- a/web/webkit/src/main/resources/toserve/lift.js
+++ b/web/webkit/src/main/resources/toserve/lift.js
@@ -104,11 +104,12 @@
 
     // Notify all functions in arr, passing any extra args
     function notifyAll(arr) {
-      var args = [].shift.call(arguments);  // Every arg but the passed array
+      var args = [].slice.call(arguments, 1);  // Every arg but the passed arr
       for(var i = 0; i < arr.length; i++) try {
         arr[i].apply(this, args);
       } catch (e) {
-        // Do nothing, just keep stuff from crapping out
+        settings.logError("Server communication callback failed!");
+        settings.logError(e);
       }
     }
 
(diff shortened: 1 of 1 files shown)


joescii:
Ah, travis failure. Looks like I've been too loose with teh JavaScriptz

joescii:
Any ideas why we're failing to boot in the travis script? I wouldn't have expected a JS enhancement to cause that... :-/

eltimn:
"src/main/resources/toserve/lift.js: line 110, col 24, Possible strict violation"

It looks like it doesn't like the use of `this` on line 110. This [SO article](http://stackoverflow.com/questions/7688765/why-is-jshint-complaining-that-this-is-a-strict-violation) explains it.

If we just assign the function to a variable `var notifyAll = function` instead of `function notifyAll` then the error goes away. I'm not sure if that's the best solution or not, but, I updated your branch locally with that change. I'll push that now.

joescii:
Awesome, thanks for the help Tim.  I guess I missed the second jshint complaint when I saw the fail to boot stacktrace. 

We could make that function call less fancy now.  When I first started it, I thought I wouldn't know the number of parameters.  Now that it's worked out, we know it's either 1 or 2 args.

joescii:
Nullifying this PR per @Shadowfiend's recognition that any Lift user can already accomplish the same effect by configuring `settings.ajaxPost` and `settings.ajaxGet` to be notified of failure or success of ajax and comet communication respectively.

PR #1761, opened 2016-01-29
Add WebJars support

### LINKED ISSUE #1755: Browsers cache lift.js even if source has changed.
As pointed out in [this thread on the ML](https://groups.google.com/forum/#!topic/liftweb/VY10pnY0MCs), Lift 3's `lift.js` artifact will be cached by browsers.  I have verified that changes to `lift.js` do not in any way trigger a refresh of the asset in the browser.  This will cause browsers to use older version of `lift.js` as teams upgrade Lift versions.

eltimn:
This adds support for WebJars. Please see the commit message for more details.

I know we're planning on releasing M8 soon, but it would be nice to get this in Lift 3.0 if possible. I'm not sure if that means it must be in M8 or not. If it's too late for 3.0, then it's ready whenever.

I created a couple of example apps that I tested with. They are here: [lift-examples](https://github.com/eltimn/lift-examples)

This would also fix #1755

There is one small problem with the build configuration. If you run publish or publishLocal from lift-framework the package task in lift-webkit will not get triggered properly and the assets will not get included in the JAR. I tried adding a dependency to them, but that only helps when you're in the lift-webkit project. In order to publishLocal, you need to run:

    ./liftsh "project lift-webkit" package "project lift-framework" publishLocal

I added a similar command to travis.sh, so that should run okay there, but it'd be nice to get this figured out.


joescii:
In the ML thread, I had asked about thoughts regarding combining js files into one asset.  While I agree that many may choose to combine assets statically as part of the build process, it seems we should provide a means of combining for folks opting to utilize our webjar support. Have you thought more about this?

joescii:
I get a jshint problem when building locally:

`jlift.js:0: Bad option: 'esversion'`

[This commit](https://github.com/lift/framework/commit/443e38d81c7b6f973f0a9b21b7109315d073c182) on my branch resolves it for me.

joescii:
BTW, I notice that the code for my comments are not showing up in my view of this PR.  Is it my view, or am I doing something dorky here?

eltimn:
I think you clicked directly into the commit instead of clicking the "Files Changed" tab on the PR page.

joescii:
You're right.  I'll do that next time :+1:

eltimn:
@joescii regarding combining files; I see 3 problems that need to be solved:

1. How to tell Lift the dependencies to combine. This would take the place of having to put the script tags in your HTML, so wouldn't be a big deal.
2. How to minify and combine. I don't think just combining them is enough, so I would want to minify as well. The only Java based minifier that I know of is google-closure. We could use that but it would add a dependency that some may not use. We could roll our own. The combining part should be pretty easy, but the minifying, not so much.
3. Testing is another concern. That's even more of an obstacle than 2 as I don't know of any Java libraries that do this.

So, I'm still not sure if that's something I'd want to tackle. But, I do understand the desire to not have to rely on build tools. 

The main reasons I wanted to add WebJar support to Lift are:

1. To have lift.js available as a WebJar asset, so it's available to build tools.
2. To have an easy way to create starter and demo apps. I doubt I would use it for more than a few dependencies in a real project.

Personally, I'll stick to using gulp and grunt for now. I understand the hesitancy of having to install node, which is why I'm working on an sbt plugin that helps with that.

joescii:
My initial thoughts on combining files was pretty simple, but perhaps naively so.  I figured we could just provide multiple `src` params in the script tag for each asset.  Combining could just be a newline between each asset.  We're already depending on the webjars to provide the minified assets, and a newline between them should be no big deal.

Note however, I'm not talking about minifying the project's js sources.  Perhaps that's what you thought I meant.

eltimn:
I'm closing this for now. I have some build tool improvements that I'm working on that should help simplify the JavaScript build that I'll update this with once ready.

I created #1764 to track it.

PR #1762, opened 2016-01-29
Ajax retry configuration

joescii:
This PR is to give Lift applications more control over ajax retries.  With these changes an application can change the ajax queue's sorting logic and retry timeout calculation.  

See [this ML thread](https://groups.google.com/forum/#!topic/liftweb/0z6SV0ucEW4) for some discussion regarding this idea.

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
@@ -922,6 +922,11 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
     else Full(toSend => Noop)
   ) {}
 
+  val ajaxQueueSortFunc: FactoryMaker[Box[JsVar => JsCmd]] = new FactoryMaker[Box[JsVar => JsCmd]](() =>
+    if (ajaxRetryInOrderReceived.vend) Full(ajaxQueue => JE.Call("lift.ajaxQueueSortOrderedRetry", ajaxQueue))
+    else Full(ajaxQueue => JE.Call("lift.ajaxQueueSortDefault", ajaxQueue))
+  ) {}
+
   /**
    * An XML header is inserted at the very beginning of returned XHTML pages.
    * This function defines the cases in which such a header is inserted.  The
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
@@ -924,8 +924,10 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
 
   val ajaxQueueSortFunc: FactoryMaker[Box[JsVar => JsCmd]] = new FactoryMaker[Box[JsVar => JsCmd]](() =>
     if (ajaxRetryInOrderReceived.vend) Full(ajaxQueue => JE.Call("lift.ajaxQueueSortOrderedRetry", ajaxQueue))
-    else Full(ajaxQueue => JE.Call("lift.ajaxQueueSortDefault", ajaxQueue))
+    else Full(ajaxQueueSortDefaultFunc)
   ) {}
+  private [http] val ajaxQueueSortDefaultFunc:JsVar => JsCmd =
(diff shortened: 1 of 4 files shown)


web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -915,6 +915,118 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
...
+    *     case _ => LiftRules.ajaxAddMetaAndQueueSortDefaultFuncs
+    *   }
+    * }
+    * }}}
+    *
+    */
+  val ajaxAddMetaAndQueueSortFuncs: FactoryMaker[Box[(JsVar => JsCmd, JsVar => JsCmd)]] = new FactoryMaker[Box[(JsVar => JsCmd, JsVar => JsCmd)]]( () =>
+    Full((ajaxAddMetaDefaultFunc, ajaxQueueSortDefaultFunc))
+  ) {}
+
```
joescii: I'd love to have some feedback on the design of this setting.  To achieve ordered messages, you have to change both of these functions together.  I think this will be the most common use case.  However, I can imagine where someone may want to change the `ajaxAddMeta` function, but not `ajaxQueueSort`.  For instance, if they wanted to add some meta data related to `ajaxCalcRetryTime` since it has access to the meta.  

I went around in circles a bit and decided to just get it out here and have some feedback.

farmdawgnation:
:eyes:

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
farmdawgnation: Could we have this `FactoryMaker` vend some case class instead of a tuple? Would make the code a lot easier to reason about while scanning. And would also avoid client code doing a lot of `_1` and `_2`.

joescii: Absolutely could. I'm just not quite sold on making sure they're coupled together once I noticed that the `ajaxAddMeta` could conceivably be useful for `ajaxCalcRetryTime`.  Maybe I'm being too open-minded to the world of possibilities which stretches beyond what anybody would actually attempt.

joescii: Ultimately, I'm trying to get the flexibility intended here, while making really easy for a developer to do the right thing.

farmdawgnation: So the case here is we want to make it easy for the developer to do what they want while encouraging them to couple these things as well because it's important to consider both at the same time? Am I interpreting that right?

joescii: Right. The ajax queue contains all of the requests wrapped with meta data.  Hence `ajaxQueueSort()` works on those wrappers.  If you want to do something different in `ajaxQueueSort()`, then it's highly likely you'll need to add to that wrapper in `ajaxAddMeta()`.

Furthermore, the out-of-the-box configuration alternative we're offering here is the retry in order, which certainly needs both in order to function correctly.  Hence I made the pair a single setting so it's not easy to set it incorrectly.

Another alternative I considered was having these two as separate settings and provide a `LiftRules` function of type `Unit => Unit` which sets both correctly for retrying in order.  I'm not aware of a prior precedent for this approach in `LiftRules`, though.

joescii:
After discussion, we have decided the the same functionality of this PR can be achieve by overriding the ajax call in its entirety.  The complication of our JS settings don't seem worth having.

Shadowfiend:
Some more thoughts here so it's written down somewhere public :)

I think the goal of `lift.js` should be to provide a reasonable set of
core behavior that the framework is built around, but that customization
of that behavior should be possible but not necessarily easy. My thought
is just that Lift's core competency isn't really JS—we're trying to provide
just enough glue to make interaction between Lift and the client relatively
painless, but not try to accommodate the 10% case of needing to
potentially change that behavior easily.

So, if we find something you *can't* customize, we should probably fix
that. It's also probably worth publishing some documentation/a wiki page/
something of the sort regarding how to customize queueing behavior and
things like that. Might be something we can pull from the work you ultimately
do for lift-ng! :)

PR #1763, opened 2016-02-01
Bump our snapshot build back to Java 8.

Shadowfiend:
Now that M8 is out the door and we've announced the change in
supported Java version, we can bump the snapshot to Java 8.


farmdawgnation:
:+1:

PR #1766, opened 2016-02-06
Convert \\ to return a JObject

### LINKED ISSUE #1518: Inconsistency when querying over JSON lists
Consider:

```scala
import net.liftweb.json._
import net.liftweb.json.JsonDSL._

object JsonTest extends App {
  val json =
    ("people" -> List(
      ("person" ->
        ("name" -> "Joe") ~
        ("age" -> 35)
      ),
      ("person" ->
        ("name" -> "Marilyn") ~
        ("age" -> 33)
      ),
      ("person" ->
        ("name" -> "Erik") ~
        ("age" -> 28)
      )))

  val names = (json \\ "name").children map { case JField(_, JString(name)) => name }
  println(names) // List("Joe", "Marilyn", "Erik")   -- as expected

  val json2 =
    ("people" -> List(
      ("person" ->
        ("name" -> "Joe") ~
        ("age" -> 35)
      )))

  val names2 = (json2 \\ "name").children map { case JField(_, JString(name)) => name }
  println(names2)  // List()   -- expecting List("Joe")

  val json3 =
    ("people" -> List[(String, JObject)]())

  val names3 = (json3 \\ "name").children map { case JField(_, JString(name)) => name }
  println(names3)  // List()   -- as expected
}
```

...is it by design that the `json2` case doesn't work as expected?

farmdawgnation:
This PR closes #1518.

I tried to be clever in an earlier implementation of this by leaving the old version in place. Unfortunately, doing so proved tricker than I'd have liked with the compiler. So unless someone comes up with Clever Ideas™ or manages to make it work, this is my proposed solution. :)

Until now, \\ returned a JValue, and could vary between returning
the value inside a field or a JObject with multiple matching values
depending on what was matched. This led to inconsistencies in the
API and unpredictable inconsistencies for the consumer of the API
(typically, if there was only one field that matched, you would get the
field's value, while multiple fields would produce a JObject with the
matching fields).

We now always return a JObject, and never return a matching field's
value directly.

### DIFF
--- a/core/json/README.md
+++ b/core/json/README.md
@@ -1,7 +1,7 @@
 Parsing and formatting utilities for JSON.
 
 A central concept in lift-json library is Json AST which models the structure of
-a JSON document as a syntax tree. 
+a JSON document as a syntax tree.
 
     sealed abstract class JValue
     case object JNothing extends JValue // 'zero' for JValue
@@ -10,7 +10,7 @@ a JSON document as a syntax tree.
     case class JDouble(num: Double) extends JValue
     case class JInt(num: BigInt) extends JValue
     case class JBool(value: Boolean) extends JValue
-    case class JObject(obj: List[JField]) extends JValue 
+    case class JObject(obj: List[JField]) extends JValue
     case class JArray(arr: List[JValue]) extends JValue
 
     case class JField(String, JValue)
@@ -83,7 +83,7 @@ Migration from older versions
 
 JField is no longer a JValue. This means more type safety since it is no longer possible
 to create invalid JSON where JFields are added directly into JArrays for instance. Most
-noticeable consequence of this change is that map, transform, find and filter come in 
+noticeable consequence of this change is that map, transform, find and filter come in
 two versions:
 
     def map(f: JValue => JValue): JValue
@@ -100,7 +100,7 @@ in the name to traverse values in the JSON.
 2.2 ->
 ------
 
-Path expressions were changed after 2.2 version. Previous versions returned JField which 
+Path expressions were changed after 2.2 version. Previous versions returned JField which
 unnecessarily complicated the use of the expressions. If you have used path expressions
 with pattern matching like:
 
@@ -117,7 +117,7 @@ Any valid json can be parsed into internal AST format.
 
     scala> import net.liftweb.json._
     scala> parse(""" { "numbers" : [1, 2, 3, 4] } """)
-    res0: net.liftweb.json.JsonAST.JValue = 
+    res0: net.liftweb.json.JsonAST.JValue =
           JObject(List(JField(numbers,JArray(List(JInt(1), JInt(2), JInt(3), JInt(4))))))
 
 Producing JSON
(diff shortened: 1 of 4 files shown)


joescii:
I don't follow why the return type needs to change.  In issue #1518, the samples show the use of `.children` which is where the `JField` type comes into play.

farmdawgnation:
I suppose could wrap the result into a `JObject` instead of just returning a raw `List[JField]`, then we'd be able to use the `JValue` return type. Is that's preferable here?

farmdawgnation:
Hmm. Yeah, I can see how that would be better... I'll change it.

joescii:
Give that a shot and I'll take a look.  I'm still tinkering with the original example given in the issue to get it fully loaded into working memory. :)

farmdawgnation:
Oh, okay, so I remembered why I thought this would be a good idea.

This change significantly changes the behavior of the function. Changing the return type will actually break compiles instead of silently breaking matches. So, we can set the return type to `JObject` or we can set it to `List[JField]` - but if we leave it as `JValue` the compiler won't provide any help toward isolating places where the behavior will have changed.

joescii:
Ah, I see your point.  So... Apparently we've changed something already in the types because the code in the issue doesn't compile against master.  I couldn't quite figure out what was going on until I pasted it into a test to find out that things indeed didn't line up.

farmdawgnation:
Yeah JField is no longer a JValue. That's probably why. 

I think List[JField] is the right answer here because of how this method tends to be used. It tends to be used as a find instead of a filter. At least IME.

joescii:
Ah, ok.  And it makes a lot of sense that `JField` would NOT be a `JValue` anyway.  

While I understand the motivation to prefer a compilation-breaking change over a behavior-breaking change, why should we go with `JField` in particular rather than `List[JValue]`?  The `name` will always match `nameToFind` in `\\`.

farmdawgnation:
Hm. Good point. Can you raise that question on the list?

farmdawgnation:
I'm currently out and will probably forget.

joescii:
Thread updated :+1:

farmdawgnation:
Alright, per the thread I'm changing this to return a JObject. :tada:

joescii:
:+1:

Shadowfiend:
:+1: from me as well, let's get 'er in!

Shadowfiend:
Can we make sure the lift-json README is up-to-date as well? In particular, can we double-check that the examples involving `\\` are still correct?

farmdawgnation:
I laugh at your entirely fair recommendation. And then agree to do it.

Shadowfiend:
Lol. Yeah, I was going to merge this last night as the last thing I was going to do, and then I remembered :(

farmdawgnation:
README will need some updates. I'll try to get around to that today.

farmdawgnation:
Okay, so I was wrong. README and such are fine it seems. I did nuke some whitespace tho.

Shadowfiend:
Nailed it! I'm :+1: here if you consider everything ready to go then :)

farmdawgnation:
Let's do it.

PR #1767, opened 2016-02-06
[WIP] LiftRules.putAjaxFnsInContainerSession

joescii:
This PR is the first result of [this ML thread](https://groups.google.com/forum/#!searchin/liftweb/stateless$20story/liftweb/KHjbjev8A0E/ldKNmvFdVeEJ).  Here I have added `LiftRules.putAjaxFnsInContainerSession` which defaults to `false` to achieve Lift's current behavior.  When set to `true`, then Lift has the ability to handle an ajax call originally served by a different instance of a Lift server _provided both instances are/were members of the same container cluster_.

I also have created a sample application which can demonstrably exercise this feature [here](https://github.com/joescii/lift-jetty-cluster/tree/failover-testing).  Details for testing are on the README of that particular branch.  As a heads-up, it requires a MySQL installation to enable Jetty to form a cluster.

This PR is the first of (at least) two which will make Lift capable of full server failover and replacement.  The second will handle re-establishing comets (or _rehydrating_ to use the venerable @Shadowfiend's term).  I have already set the version for this branch to `3.1-SNAPSHOT` so consider it a low priority.  I initially planned to create a single PR with both ajax and comet support, but I'm quite sure they are entirely unrelated and will likely cause confusion in our discussion of implementation details.


web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -1713,6 +1713,8 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
 
   @volatile var cometGetTimeout = 140000
 
+  @volatile var putAjaxFnsInContainerSession = false
```
farmdawgnation: Should we do something like this...

```  def addToPackages(what: String) {
    if (doneBoot) throw new IllegalStateException("Cannot modify after boot.");
    otherPackages = what :: otherPackages
  }
```

to prevent modification after `Boot`? This doesn't seem like something that should change mid-application.

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
```diff
@@ -364,7 +364,26 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
...
 
-  private val nmessageCallback: ConcurrentHashMap[String, S.AFuncHolder] = new ConcurrentHashMap
+  // Returns a view of the functions stored in the http session
+  private def viewHttpSessionFns:java.util.Map[String, S.AFuncHolder] = java.util.Collections.unmodifiableMap(
+    if(LiftRules.putAjaxFnsInContainerSession) (for {
+        hs <- httpSession
+        map <- Box.legacyNullTest(hs.attribute("ajaxFns").asInstanceOf[ConcurrentHashMap[String, S.AFuncHolder]])
+      } yield map).openOr(new java.util.HashMap[String, S.AFuncHolder])
+    else new java.util.HashMap[String, S.AFuncHolder]
+  )
```
farmdawgnation: This is kind of difficult to parse. Can we do...

```scala
private def viewHttpSessionFns:java.util.Map[String, S.AFuncHolder] = {
  val backingMap = for {
    hs <- httpSession if LiftRules.putAjaxFnsInContainerSession
    map <- Box.legacyNullTest(hs.attribute("ajaxFns").asInstanceOf[ConcurrentHashMap[String, S.AFuncHolder]])
  } yield {
    map
  }

  java.util.Collections.unmodifiableMap(backingMap)
}
```

WDYT?

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
joescii: That is a fair point. I wish we had a _set at most once_ semantics thingy lying around. How have we handled this with other settings?

farmdawgnation: With the code snippet I used above. :grin:

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
```diff
@@ -364,7 +364,26 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
...
+
+  // Mutates the http session function map with the given argument and updates the container
+  private def updateHttpSessionFns(f:ConcurrentHashMap[String, S.AFuncHolder]=>Unit):Unit =
+    if(LiftRules.putAjaxFnsInContainerSession) httpSession.foreach { hs =>
+      val map = Box.legacyNullTest(hs.attribute("ajaxFns").asInstanceOf[ConcurrentHashMap[String, S.AFuncHolder]]).openOr(new ConcurrentHashMap[String, S.AFuncHolder])
+      f(map)
+      // Now that the map has been mutated, must write to the session so the container can replicate
+      hs.setAttribute("ajaxFns", map)
+    }
```
farmdawgnation: I'm all about dat for-comprehension today.

```scala
private def updateHttpSessionFns(f:ConcurrentHashMap[String, S.AFuncHolder]=>Unit):Unit = {
  for {
    httpSession <- httpSession if LiftRules.putAjaxFnsInContainerSession
    map <- Box.legacyNullTest(hs.attribute("ajaxFns").asInstanceOf[ConcurrentHashMap[String, S.AFuncHolder]])
  } {
    f(map)

    // Now that the map has been mutated, must write to the session so the container can replicate
    hs.setAttribute("ajaxFns", map)
  }
}
```

farmdawgnation:
1. Are there any tests we could add for this functionality?
2. How does this dovetail with things bound with things like `SHtml.text`? E.g. would an entire form work using this or only the ajax bits of it?

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
joescii: I'm liking the for-comps! :+1:

joescii:
1. Seems likely. I'm a huge advocate for TDD when you know what you're doing. In this case, I just inserted code and explored. :)  Once we get enough feedback to feel confident this is the right way to solve this problem, then I'm game for even going as far as to start fresh and write tests first if it makes sense.

2. In addition to a critique on my approach, I also need a lot of questions like this one.  `SHtml.text` is a Lift feature I've yet to use, so I'll take a closer look to understand how it's impacted.  Then we can decide if it should fall under this same PR (and perhaps rename the `LiftRules` setting appropriately) or place it in a separate PR for individual review.

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
```diff
@@ -579,12 +603,18 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
...
+      updateHttpSessionFns(copyFunctions(funcs, _))
+  }
+
+  private def copyFunctions(from: Map[String, S.AFuncHolder], to:java.util.Map[String, S.AFuncHolder]):Unit =
+    from.foreach {
       case (name, func) =>
-        nmessageCallback.put(name,
+        to.put(name,
           if (func.owner == Full(uniqueId)) func else func.duplicate(uniqueId))
     }
```
joescii: I want to point out that I originally wanted to put the entire `nmessageCallback` in the session, but unfortunately it contains massive amounts of non-serializable stuff. IIRC [this particular `put`](https://github.com/lift/framework/blob/joe-ajax-fns-in-session/web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala#L995-L997) was a notable culprit.  So there is more that we could theoretically support for failover if we meticulously groom the object tree that is creeping into this map.

joescii:
Follow up regarding (2): It appears that the function in `SHtml.text` eventually finds its way into [`S.__functionMap`](https://github.com/lift/framework/blob/joe-ajax-fns-in-session/web/webkit/src/main/scala/net/liftweb/http/S.scala#L274-L281) and hence is not covered by this PR at the moment.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
Shadowfiend: I've just been making the var be the feeder for a lazy val that gets computed and reused. See [`lockedSecurityRules`](https://github.com/lift/framework/blob/joe-ajax-fns-in-session/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala#L284).

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
Shadowfiend: Er… That `put` is what ensures that `RequestVar` context will be available in AJAX callbacks. So if you're not managing to capture it somehow, your AJAX callbacks will be able to run elsewhere, but only without part of their context.

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
Shadowfiend: You could just combine the cast in so that it's also safe: `map <- Box.asA[ConcurrentHashMap[String, S.AFuncHolder]](hs.attribute("ajaxFns")` `Box.asA` returns an `Empty` for a `null`.

Shadowfiend: Though notably that comprehension has different behavior, since the current code will create a map if it can't find it, while the comprehension will simply not run if there's no existing entry.

farmdawgnation: Ah. I didn't catch that we defaulted to creating a map here. That can also be done, too, though. hehe.

joescii: Yep, I agree with you @Shadowfiend.  I converted it to a nicer-looking for-comp, but I didn't change it exactly like @farmdawgnation suggested for that reason,

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
joescii: Oh, that's simple and perfect :+1:

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
joescii: I'm still capturing it. The snippet above doesn't show it, but what happened is I turned that bit of code into a `def copyFunctions` so I could use the logic for both the original `nmessageCallback` and the container-managed map.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
joescii: @farmdawgnation I'm not sure what happened in my head, but when I first saw your snippet I misunderstood how it would work for this setting. I get it now, lol. I opted for the approach @Shadowfiend used tho.  Which do y'all think is better?

farmdawgnation: I generally assume Antonio's code is always better than my own. :)

That said, his would fail silently if you tried to do a bad thing (change after boot) wouldn't it?

joescii: Yes it would. That's one thing that is preferable about the other approach.

joescii:
Ha... I just noticed the Codacy results on this PR.  I have a lot of work to do, lol

farmdawgnation:
Yo, how is this looking for 3.1?

Shadowfiend:
Going to drop the milestone here. We can restore it when timelines are a little clearer, I think.

farmdawgnation:
This still has a WIP designation. Is it ready for some review or nah?

joescii:
Nope, we don't need this PR, @farmdawgnation. The route I eventually decided to take is _much_ cleaner that what I found myself needing to do with this one. Apologies for the lack of clean up.

Issue #1757, opened 2016-01-26
lift-json issues with deeply nested tuples including Scala primitive types


Shadowfiend:
``` scala
import net.liftweb.common._
import net.liftweb.json.{Extraction, _}
import net.liftweb.util.ControlHelpers._
import org.specs2.mutable.

case class Holder(items: List[(String, Long)])

class ExtractionBugSpec extends Specification {
  "Extraction" should {
    "deserialize List of Tuple with Long" in {
      implicit val formats = DefaultFormats

      val holder = Holder(List(("string", 1L)))
      val serialized = compactRender(Extraction.decompose(holder))
      println(serialized)

      val deserialized = tryo(parse(serialized).extract[Holder])
      deserialized must beLike {
        case Full(holderObj) => ok
        case other => {
          println(other)
          ko
        }
      }
    }
  }
}
```

Result:

```
[info] Extraction should
[error] x deserialize List of Tuple with Long
[error]    'Failure(No usable value for items
[error]    No usable value for _2
[error]    No information known about type,Full(net.liftweb.json.MappingException: No usable value for items
[error]    No usable value for _2
[error]    No information known about type),Empty)' doesn't match the expected pattern ko (ExtractionBugSpec.scala:23)
```

ML thread: https://groups.google.com/d/msg/liftweb/8kTftPP7jyQ/jS6UYHinDAAJ

Shadowfiend:
Some deeper investigation revealed that paranamer or reflection (not sure which yet) is reporting that the `Long` parametrized type is instead a `java.lang.Object`: `scala.collection.immutable.List<scala.Tuple2<java.lang.String, java.lang.Object>>`.

andreak:
Last time i had this issue (but with (Long, String)), I came to the conclusion that a custom serializer is needed for this (after some googleing).

```scala
import net.liftweb.json.CustomSerializer
import net.liftweb.json.JsonAST.{JField, JObject, JString}

class LongStringTupleSerializer extends CustomSerializer[(Long, String)](format => ( {
	case JObject(List(JField(k, JString(v)))) => (k.toLong, v)
}, {
	case (l: Long, t: String) => JObject(List(JField(l.toString, JString(t))))
}))
```

Usage:
```scala
implicit val formats = DefaultFormats.lossless ++ JodaTimeSerializers.all + new JsonBoxSerializer + new LongStringTupleSerializer
```

Shadowfiend:
[...]

PR #1768, opened 2016-02-07
Work on improved Tuple support in lift-json

farmdawgnation:
# Introduction

Tuples are a wily beast in Scala land, but one that has its point and its place. Given that Tuples are a very good construct to understand JavaScript's heterogenous arrays, it makes sense that lift-json would interpret them in this fashion when serializing and deserializing JSON.

Ticket: #1757.

# Summary

Before this pull request, tuples will be serialized to this format:

```json
{"_1": "value1", "_2": "value2"}
```

After this pull request, tuples will be serialized to this format:

```json
["value1", "value2"]
```

However, we've also added support for understanding truly heterogeneous collections from the client. So, if you wanted to transmit data that looked something like this:

```json
[["vector1", 2], ["vector2", 2]]
```

You should now be able to ask lift-json to interpret this as a `List[(String, Int)]`. Likewise, you could decompose a `List[(String, Int)]` to that format. This is a fairly common format in JavaScript land, so this change will let developers using lift-json interact with libraries and APIs expecting these formats more fluidly.

# State of this PR

* All tests are currently passing.
* Support for scala primitives, like `Int`, seems to be a bit more spotty than for their Java object counterparts, like `Integer`.
* The changes in this PR are still able to understand the _old_ serialized format of tuples, so this shouldn't be a breaking change for deserialization. It will, however, change the wire format for anyone who is using tuples. Perhaps we should consider using a flag to turn on / off the new format?


thomasdziedzic:
Just encountered an issue related to deserializing tuples in lift, are there any updates with regards to this PR since the last push/comment on the related thread?

farmdawgnation:
Hey there,

No I've largely been waiting for 3.0 to get out the door before pushing too hard on this since this is significant enough to have to wait on 3.1. 

As I recall this is close - the biggest gap I remember is lack of support for scala primitives.

farmdawgnation:
Oh hey it's been almost a year since I started this. Should probably circle back to it at some point. :P

farmdawgnation:
Alright, I think this is officially ready for some consideration. @Shadowfiend tag you're it!

farmdawgnation:
Codacy is complaining about things that I'm not super motivated to fix. I want the "X" to go away but it's not going away. 😢

Shadowfiend:
I can try to clear those in the mornin' if you'd like.

farmdawgnation:
By all means [ shrug ]

core/json/src/main/scala/net/liftweb/json/Extraction.scala
```diff
@@ -319,14 +331,54 @@ object Extraction {
       }
     }
 
+    def newTuple(root: JValue, mappings: List[Mapping]): Any = {
```
Shadowfiend: Should this be `private[this]`?

core/json/src/main/scala/net/liftweb/json/Extraction.scala
```diff
@@ -319,14 +331,54 @@ object Extraction {
...
+      root match {
+        case JArray(items) if items.length >= 1 && items.length <= 22 =>
+          val builtItems: Seq[Object] = items.zip(mappings).map({
+            case (i,m) =>
+              build(i, m).asInstanceOf[Object]
```
Shadowfiend: Probably worth giving `i` and `m` longer names.

core/json/src/main/scala/net/liftweb/json/Extraction.scala
```diff
@@ -319,14 +331,54 @@ object Extraction {
...
+          })
+          val tupleClass = Class.forName("scala.Tuple" + items.length)
```
Shadowfiend: We should build this ahead of time in a top-level `final val`; `Class.forName` is really expensive. We might also use that to remove the `>=1 && <= 22` magic numbers above.

This could go hand-in-hand with our friendly neighborhood `tuples` set, since it's already got the right `Class` instance anyway.

core/json/src/main/scala/net/liftweb/json/Extraction.scala
```diff
@@ -319,14 +331,54 @@ object Extraction {
...
+
+          val typedTupleConstructor = tupleClass.getConstructors()(0)
+          typedTupleConstructor.newInstance(builtItems: _*)
+
+        case JObject(items) if items.forall(_.name.startsWith("_")) =>
+          newTuple(JArray(items.map(_.value)), mappings)
```
Shadowfiend: We probably need to sort these. For example, if we get a `{"_3": 5, "_1": "hello", "_2": "uh-oh"}`, that should deserialize to `Tuple3[String,String,Int]("hello", "uh-oh", 5)`, but I think with this code we'd get `Tuple3[Int,String,String](5, "hello", "uh-oh")`.

core/json/src/main/scala/net/liftweb/json/Extraction.scala
```diff
@@ -319,14 +331,54 @@ object Extraction {
...
+
+        case JArray(items) =>
+          throw new IllegalArgumentException("Cannot create a tuple of length " + items.length)
```
Shadowfiend: It would be nice to move this condition above the `JObject` one so the two array possibilities are together.

core/json/src/main/scala/net/liftweb/json/Meta.scala
```diff
@@ -108,55 +109,80 @@ private[json] object Meta {
...
           val typeArgs = typeConstructors(t, k)(valueTypeIndex)
           factory(fieldMapping(typeArgs)._1)
-        } else factory(fieldMapping(typeParameters(t, k, context)(valueTypeIndex))._1)
+        } else {
+          factory(fieldMapping(typeParameters(t, k, context)(valueTypeIndex))._1)
+        }
+      }
+
+      def mkHeteroContainer(baseType: Type): Mapping = {
+        val hereroContainerTypes = baseType match {
```
Shadowfiend: Should be `hetero` with a t.

core/json/src/main/scala/net/liftweb/json/Meta.scala
```diff
@@ -108,55 +109,80 @@ private[json] object Meta {
...
+          case ptype: ParameterizedType =>
+            ptype.getActualTypeArguments().map {
+              case c: Class[_] =>
+                c
+              case p: ParameterizedType =>
+                p.getRawType.asInstanceOf[Class[_]]
+              case x =>
+                fail("do not know how to get type parameter from " + x)
+            }
```
Shadowfiend: Do we just want a `MatchError` if `baseType` isn't a `ParameterizedType`, or do we want to `fail` with our own error message?

core/json/src/test/scala/net/liftweb/json/ExtractionBugs.scala
```diff
@@ -96,4 +96,30 @@ object ExtractionBugs extends Specification {
...
+    case class Holder(items: List[(String, String)])
+
+    val holder = Holder(List(("string", "string")))
+    val serialized = compactRender(Extraction.decompose(holder))
+
+    val deserialized = parse(serialized).extract[Holder]
+    deserialized must_== holder
+  }
+
+  "deserialize list of heterogenous tuples" in {
```
Shadowfiend: Either `deserialize a list` or `deserialize lists`?

core/json/src/test/scala/net/liftweb/json/ExtractionBugs.scala
```diff
@@ -96,4 +96,30 @@ object ExtractionBugs extends Specification {
...
+    implicit val formats = DefaultFormats
+
+    // MSF: This currently doesn't work with scala primitives?! The type arguments appear as
+    // java.lang.Object instead of scala.Int. :/
```
Shadowfiend: Is this still true? And if it is still true, is that something that was already missing in the standard object serialization of tuples, or not so much?

core/json/src/main/scala/net/liftweb/json/Extraction.scala
farmdawgnation: Potentially? I can't think of a reason this would be invoked elsewhere. :)

core/json/src/main/scala/net/liftweb/json/Extraction.scala
farmdawgnation: Indeed. I probably copied this from some other spot.

core/json/src/main/scala/net/liftweb/json/Extraction.scala
farmdawgnation: Good call. Let me look into that.

core/json/src/main/scala/net/liftweb/json/Extraction.scala
farmdawgnation: I have a back-compat test somewhere, I think, that suggests otherwise. If so, I'll point to it here. If not, I'll take a closer look at it.

core/json/src/test/scala/net/liftweb/json/ExtractionBugs.scala
farmdawgnation: > Is this still true?

I'm not sure. My memory seems to think "no" is the answer there, but that could be a 2.12 development which would still make it true for 2.11.

> And if it is still true, is that something that was already missing in the standard object serialization of tuples, or not so much?

If it is still true, it would be reflective of an issue that has always been latent in scalap, I think. Whether or not we triggered that error condition in the old implementation: I don't know.

Let me do some more research on the above.

core/json/src/main/scala/net/liftweb/json/Extraction.scala
farmdawgnation: Ah, sorry, no, the pattern here is that we nest the `newX` functions inside of the larger extraction handler. So, visibility modifiers aren't allowed here because we're not in the class scope (and nothing outside this function can access it anyway).

At some point I'd like to unwind that behavior, but now is not the right time to do that.

farmdawgnation:
Okay, following up on [the issue with Scala primitives](https://github.com/lift/framework/pull/1768#discussion_r105205381)...

The underlying problem here is that (apparently) to Java reflection the Scala primitives just looks like an instance of `java.lang.Object`. We special case this in a lot of different areas of the extraction code, and I have some ideas of some kinds of special casing here, too, but we've probably long ago crossed that threshold where this code was maintainable. :-(

I think we might want to consider a fuller rewrite of lift-json's parser at this point, unfortunately. This is something we should be able to support, but I'm fighting a _lot_ of opaque code and special cases to try and make it happen without very much quick progress. These are signs that the area of code I'm working in might be due for a wholesale teardown and rebuild.

The question here, then, is what do we want to do with this PR?

core/json/src/main/scala/net/liftweb/json/Extraction.scala
```diff
@@ -333,22 +333,25 @@ object Extraction {
...
           val builtItems: Seq[Object] = items.zip(mappings).map({
-            case (i,m) =>
-              build(i, m).asInstanceOf[Object]
+            case (item, mapping) =>
+              build(item, mapping).asInstanceOf[Object]
           })
-          val tupleClass = Class.forName("scala.Tuple" + items.length)
+          val tupleIndex = items.length - 1
+          val tupleClass = tuples.lift(tupleIndex).getOrElse {
+            throw new IllegalArgumentException(s"Cannot instantiate a tuple of length ${items.length} even though that should be a valid tuple length.")
```
Shadowfiend: Still think it may be worth turning `tuples` into a `Map` (only doing the `.getConstructors()(0)` bit once during the singleton intialization).

The error here also doesn't seem quite right (at this point we actually couldn't load the tuple class, so while technically we can't instantiate it there's a deeper reason). Additionally, if we hadn't been able to get this class, we would have blown up while initializing the whole singleton object. I'm wondering if we just do a `.get` here and point to the initialization of `tuples` to show that if we've gotten to this point in the code, we know that entry exists.

core/json/src/main/scala/net/liftweb/json/Extraction.scala
```diff
@@ -333,22 +333,25 @@ object Extraction {
...
           typedTupleConstructor.newInstance(builtItems: _*)
 
-        case JObject(items) if items.forall(_.name.startsWith("_")) =>
-          newTuple(JArray(items.map(_.value)), mappings)
-
         case JArray(items) =>
           throw new IllegalArgumentException("Cannot create a tuple of length " + items.length)
 
+        case JObject(items) if items.forall(_.name.startsWith("_")) =>
+          newTuple(JArray(items.map(_.value)), mappings)
```
Shadowfiend: Still wondering about this. Even if we have a backward-compat deal that shows that we used to serialize `_1` and friends in order, the old way of dealing with tuples (which is constructor-based) would have dealt with a user furnishing an out-of-order set of _*s (not necessarily serialized from here). If we have a test that shows that you can pass `{"_3": "5", "_1": "1", "_2": "8"}` (e.g.) to a `(String, String, String)` and it deserializes the same way with this extraction strategy, I'm game. But I don't see how that would happen, tbh.

farmdawgnation:
> We may, however, want to make the serialization behavior (old-style vs array-style) togglable somehow, otherwise we lose backward-compatibility.

That's reasonable. I don't think we have a pattern for this kind of toggle in lift-json already, however. Did you have any concrete ideas about how that might look?

> The other option is saying "screw it, we won't support tuple arrays in this version of the extractor ever, and if you want that look to the new one if/whenever it gets off the ground”.

So I'm still struggling with this. There's some variation of the phrase like "the lack of a feature is better than the presence of a feature implemented poorly." That said, for all cases except Scala primitives nested inside a type argument my understanding is that this would work. And that's quite a lot of cases.

core/json/src/main/scala/net/liftweb/json/Extraction.scala
farmdawgnation: I think I'm better understanding what these comments about array ordering were getting at and I think you're right. If so, then fixing this should be as simple massaging the key to a numeral and doing a sort before reducing it to a `JArray`.

core/json/src/main/scala/net/liftweb/json/Extraction.scala
farmdawgnation: > I'm wondering if we just do a .get here and point to the initialization of tuples to show that if we've gotten to this point in the code, we know that entry exists.

No, I think having a `getOrElse` with an exception here is a good way of refactor-proofing the code. Having worked in a project that made way-too-ample use of `Option.get` last year I think I've landed on always justifying a retrieval that has no alternative much like `Box`'s `openOrThrowException` requires us to do. The alternative is making a bet that this code can't be fouled by an accidental fat finger and then surfacing an opaque error message to a user of the Framework. Certainly, surfacing this error at all would be bad, but making it opaque is just insult to injury. :)

> Still think it may be worth turning tuples into a Map (only doing the .getConstructors()(0) bit once during the singleton intialization).

I'm not quite following the entirety of this.

Shadowfiend: Sorry, basically doing something like this:

```scala
val tupleClasses =
  Seq(classOf[Tuple1[_]]...)
val tupleConstructors =
  tupleClasses.map(clazz => (clazz -> clazz.getConstructors()(0))).toMap
```

And then looking the constructor up in the map every time.

Shadowfiend:
> Did you have any concrete ideas about how that might look?

Not per se, but it could just be a boolean flag to `extract` for now, perhaps.

farmdawgnation:
Alright, I'm going to figure out something for turning this off/on then and we'll ship it as is for now. Which is to say, with caveats.

core/json/src/main/scala/net/liftweb/json/Meta.scala
farmdawgnation: Yeah we can add a fail condition of our own here.

farmdawgnation: Oh wait we already do that. I think that's the right call. :)

farmdawgnation:
Okay I think this PR is ready for further review and (hopefully) a merge.

I opted to store the flag for the experimental tuple support in the `Formats`. That object is well suited both semantically and in terms of availability in all the places where I need to make decisions on how to handle tuples. I've added a guard everywhere that I (think) I was introducing new behavior but **please please please** double check me so we don't accidentally introduce a regression.

For now, we'll accept the tradeoff of not supporting Scala primitives in favor of having an exit hatch for folks who want to use heterogenous collections. I'll prepare a blog post about this for my personal blog so there's _something_ out there that I can point folks to when they inevitably ask questions about our support for this new style of Tuple.

If it looks good to you give it the sign-off and I'll merge once the blog post is ready. :)

core/json/src/main/scala/net/liftweb/json/Formats.scala
```diff
@@ -32,6 +32,14 @@ trait Formats { self: Formats =>
   val fieldSerializers: List[(Class[_], FieldSerializer[_])] = Nil
 
   /**
+   * Support for the experimental tuple decomposition/extraction that represents tuples as JSON
+   * arrays. This provides better support for heterogenous arrays in JSON, but enable it at your
+   * own risk as it does change the behavior of serialization/deserialization and comes
+   * with some caveats (such as Scala primitives not being recognized reliably during extraction).
```
Shadowfiend: Also, should we really label this “experimental”? We're not planning on iterating on it really, it's just a different serialization approach? Perhaps `serializeTuplesAsArrays`?

core/json/src/test/scala/net/liftweb/json/ExtractionBugs.scala
```diff
@@ -122,4 +122,50 @@ object ExtractionBugs extends Specification {
...
+    case class Holder2(items: List[(String, Integer)])
+
+    val holder = Holder2(List(("string", 10)))
+    val serialized = compactRender(Extraction.decompose(holder))
+
+    val deserialized = parse(serialized).extract[Holder2]
+    deserialized must_== holder
+  }
+
+  "deserialize an out of order old-style tuple w/ experimental tuples enabled" in {
```
Shadowfiend: We have existing tests for these working with experimental tuples disabled as well, right? I assume that was the old default, just want to make sure we didn't delete the tests in question at some point.

core/json/src/test/scala/net/liftweb/json/ExtractionBugs.scala
```diff
@@ -122,4 +122,50 @@ object ExtractionBugs extends Specification {
     val deserialized = parse(serialized).extract[Holder2]
     deserialized must_== holder
   }
+
+  "deserialize list of homogonous tuples w/ experimental tuples enabled" in {
```
Shadowfiend: Given the enabled/disabled tests are identical here, I wouldn't be opposed to collapsing them into a single test---but don't care too much.

core/json/src/main/scala/net/liftweb/json/Formats.scala
farmdawgnation: I think there should be a yellow flag of sorts that alerts someone who sees nothing except this flag being used that they should tread carefully. If you'd like to create an alias value named "serializeTuplesAsArrays" that reads from this and reference that directly in the extractor, that seems sensible to me. But at least until we're getting some positive reports from the field on this feature I'd like to have a warning inherent in the name. And calling it experimental fits the bill, imo.

core/json/src/test/scala/net/liftweb/json/ExtractionBugs.scala
farmdawgnation: I prefer the way this comes out in the test report.

core/json/src/test/scala/net/liftweb/json/ExtractionBugs.scala
farmdawgnation: The old behavior probably isn't covered explicitly, but would be covered in Lift-Json's handling of constructor arguments in ExtractionExamplesSpec.

farmdawgnation:
> Just to clarify, you now have two options:
>
> * Serialize tuples as JSON objects (old behavior). Works with primitives.
> * Serialize tuples as arrays (new behavior). Doesn't work with primitives.

Correct.

> Additionally:

> * Deserialization of tuples as objects is supported regardless of flags. This supports primitives.

Correct.

> * Deserialization of tuples as arrays is supported regardless of flags. This does not support primitives.

Incorrect. To deserialize an array into a tuple the flag must be enabled in your formats object. Unfortunately, the code paths diverge here in the "planning" stage of extraction, before we have the actual data available. So, I either create a planning instruction that handles a Heterogenous Collection (`HCol`) or I create one that asks for a constructor.

If I were to try to handle this I would have to break our abstraction a bit by specifically checking constructor's class names and munging based on that. Since that risks introducing edge cases where we break the existing, stable serialization, I decided not to go that route.

> Lastly, blog posts as documentation for Lift seems a bit ~. Can we put at least a blurb about this in the lift-json README? Definitely don't want to stop you from writing a blog post---just want to make sure the basics are easily accessible from Lift sources as well.

Yeah, you're right. We need to document the current state of the world in the README and API docs. The purpose of the blog post is to answer the question "How is the new thing different from the old thing?" which is a question I prefer to answer in a medium that has a datestamp firmly attached to it.

I'll look to make code changes after your responses to all of my comments today.

core/json/src/main/scala/net/liftweb/json/Formats.scala
Shadowfiend: From my perspective, the implementation is tested. It may have bugs, which we will fix because we support the feature. But we don't expect the API to change, and we do intend on supporting it, at least until this whole extraction paradigm is ripped out in some potential future. That doesn't merit calling it experimental IMO.

Additionally, I'm opposed to having this as a temporary name, because then we create headaches when we want to make it the “correct” name.

I think the new work you're doing on a TypeTags-based extractor merits being called experimental, because the support horizon on it differs from that of other features (by virtue of being in a separate codebase hehe). But if we're introducing this I assume we intend to support it, and if we intend to support it I think it's not experimental.

If you agree, cool. If not, maybe let's see if someone else wants to chime in and break the tie. If we're nowhere within a day or two or when you circle back (whichever is longest), let's disagree and commit to do it your way.

farmdawgnation: I think the most compelling argument here are the headaches for converting to a correct name. I still think there are some dangers to turning this on because JSON is simple, but still manages to be bananas at finding bugs in code. That said, it's turned off by default and perhaps we can provide enough context as to the meaning and impact of this change in the documentation to be sufficient.

I'll change it to be named `serializeTuplesAsArrays`.

Shadowfiend:
Given no more updates, going to merge this sucker right onnnnn innnn.

PR #1769, opened 2016-02-07
JS settings override

joescii:
Resolving an issue discovered during the discussion in [this ML thread](https://groups.google.com/forum/#!topic/liftweb/0z6SV0ucEW4).  Without this PR, it is not possible to override client settings applied from either `liftJQuery` or `liftVanilla`.  This PR extends the settings object with `liftJQuery` / `liftVanilla` _first_ then apply the usual settings.  This puts the Lift developer back in control and able to provide their own settings.

Here is an example page js after applying this PR:

```javascript
var lift_settings = {};
window.lift.extend(lift_settings,window.liftJQuery);
window.lift.extend(lift_settings,{"liftPath": "/lift", "ajaxRetryCount": 3, "ajaxPostTimeout": 5000, "gcPollingInterval": 75000, "gcFailureRetryTimeout": 15000, "cometGetTimeout": 140000, "cometFailureRetryTimeout": 10000, "cometServer": null, "logError": function(msg) {lift.defaultLogError(msg);}, "ajaxOnFailure": function() {alert("The server cannot be contacted at this time");}, "ajaxOnStart": function() {jQuery('#'+"ajax-loader").show();}, "ajaxOnEnd": function() {jQuery('#'+"ajax-loader").hide();}, "ajaxPost": function() {console.log("hijacked!!");}});
window.lift.init(lift_settings);
```

At the very end of the second extend, you can see where I hijacked the `ajaxPost` function for grins.

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/js/LiftJavaScript.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/js/LiftJavaScript.scala
@@ -75,13 +75,13 @@ object LiftJavaScript {
   }
 
   def initCmd(settings: JsObj): JsCmd = {
-    val extendJqOrVanilla = LiftRules.jsArtifacts match {
+    val extendJsHelpersCmd = LiftRules.jsArtifacts match {
       case JQueryArtifacts => Call("window.lift.extend", JsVar("lift_settings"), JsVar("window", "liftJQuery"))
       case _ => Call("window.lift.extend", JsVar("lift_settings"), JsVar("window", "liftVanilla"))
     }
 
     JsCrVar("lift_settings", JsObj()) &
-    extendJqOrVanilla &
+    extendJsHelpersCmd &
     Call("window.lift.extend", JsVar("lift_settings"), settings) &
     Call("window.lift.init", JsVar("lift_settings"))
   }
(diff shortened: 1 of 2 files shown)


web/webkit/src/main/scala/net/liftweb/http/js/LiftJavaScript.scala
```diff
@@ -75,20 +75,14 @@ object LiftJavaScript {
   }
 
   def initCmd(settings: JsObj): JsCmd = {
-    val extendCmd = LiftRules.jsArtifacts match {
-      case jsa: JQueryArtifacts => Call("window.lift.extend", JsVar("lift_settings"), JsVar("window", "liftJQuery"))
+    val extendJqOrVanilla = LiftRules.jsArtifacts match {
```
Shadowfiend: Why such a specific name? Currently these are the only two things this can extend, but there's no real reason to capture that in the variable name IMO. Perhaps `extendJsHelpersCmd`?

joescii: Yeah, I like yours better.  The original name was so generic as to not be correct because it isn't the only extend.  I over corrected. :)

Shadowfiend:
Dig it, and dig the improved readability of `LiftJavaScript`. I say let's do it but I'll wait in case anyone else wants to say anything. :+1:

farmdawgnation:
I'm game as well. :+1:

PR #1770, opened 2016-02-14
Comet Rehydration

joescii:
As discussed on [this ML thread](https://groups.google.com/forum/#!topic/liftweb/KHjbjev8A0E), comet rehydration is another step towards supporting Lift server instance failover.

This feature is DISABLED by default.  It can be enabled with these two lines in `Boot`:

``` scala
LiftRules.redirectAsyncOnSessionLoss = false
LiftRules.noCometSessionCmd.default.set(() => JsCmds.Run("lift.rehydrateComets()"))
```

Rehydrating comets works by refetching the html from the Lift server.  The html is slapped into an `iframe` and scraped for the comet GUID and version pairs.  The `toWatch` is replaced with the new comet pairs effectively allowing the newly created comets to take over.

As with the [LiftRules.putAjaxFnsInContainerSession PR](https://github.com/lift/framework/pull/1767), this feature can be exercised with [this project](https://github.com/joescii/lift-jetty-cluster/tree/failover-testing) but using the `/comet` page.


joescii:
I vote for marking codacy's complaint about cyclomatic complexity as `WONTFIX`.

farmdawgnation:
Hey @joescii - what are your thoughts on the stability of this? Is this something that could potentially be included in 3.1 or nah?

Shadowfiend:
Going to drop the milestone here. We can restore it when timelines are a little clearer, I think.

farmdawgnation:
@joescii I see you've been periodically merging master here. LMK When you think it's ready for review/  testing

web/webkit/src/main/resources/toserve/lift.js
```diff
@@ -680,12 +715,16 @@
         contentType: contentType
       });
     },
-    ajaxGet: function(url, data, onSuccess, onFailure) {
+    ajaxGet: function(url, data, onSuccess, onFailure, dataType) {
+      if (!dataType) {
+        dataType = "script";
+      }
+
```
joescii: @Shadowfiend This is the change I was talking about in this ML thread https://groups.google.com/forum/#!topic/liftweb/KdH4kgZ-5ug

web/webkit/src/main/resources/toserve/lift.js
```diff
@@ -397,6 +400,37 @@
...
     }
 
+    function rehydrateComets() {
+      function onSuccess(html) {
+        var iframe = document.createElement("iframe");
+        var doc = document.createElement("html");
+        iframe.appendChild(doc);
+        doc.innerHTML = html;
+
+        toWatch = {};
```
farmdawgnation: Is this attached to window or a lift object on the window?

web/webkit/src/main/resources/toserve/lift.js
```diff
@@ -397,6 +400,37 @@
...
+          var version = newComets[i].getAttribute("data-lift-comet-version");
+          var guid = newComets[i].children[0].getAttribute("id");
+          toWatch[guid] = version;
+          oldComets[i].outerHTML = newComets[i].outerHTML;
+        }
+
+        restartComet();
+
+        // Under the impression that a parentless node needn't be deleted.
+        // It should be a candidate for GC after it goes out of scope
```
farmdawgnation: I think this is right. We should be able to remove this comment.

web/webkit/src/main/resources/toserve/lift.js
```diff
@@ -397,6 +400,37 @@
...
+      }
+      function onFailure(err) {
+        // Try again??
+        settings.logError(err);
```
farmdawgnation: Could we fall back to the classic comet refresh method? (Reloading the page?)

web/webkit/src/main/scala/net/liftweb/builtin/snippet/Comet.scala
```diff
@@ -32,8 +32,12 @@ object Comet extends DispatchSnippet with LazyLoggable {
 
   // Take the comet's internal container and annotate it with the unique
   // `containerId`.
-  private def buildContainer(cometHtml: NodeSeq, cometActor: LiftCometActor, containerId: String): NodeSeq = {
-    cometActor.parentTag.copy(child = cometHtml) % ("id" -> containerId)
+  private [this] def buildContainer(cometHtml: NodeSeq, cometActor: LiftCometActor, containerId: String): NodeSeq = {
+    val cometVersion = http.S.requestCometVersions.is
+      .collectFirst { case CometVersionPair(guid, ver) if guid == cometActor.uniqueId => ver }
```
farmdawgnation: Please break this into multiple lines for readability.

web/webkit/src/main/scala/net/liftweb/builtin/snippet/Comet.scala
```diff
@@ -32,8 +32,12 @@ object Comet extends DispatchSnippet with LazyLoggable {
...
+      .getOrElse("")
+
+    cometActor.parentTag.copy(child = cometHtml) % ("id" -> containerId) % ("data-lift-comet-version" -> cometVersion)
```
farmdawgnation: Please break this into multiple lines for readability.

web/webkit/src/main/resources/toserve/lift.js
```diff
@@ -397,6 +400,37 @@
       }
     }
 
+    function rehydrateComets() {
+      function onSuccess(html) {
+        var iframe = document.createElement("iframe");
+        var doc = document.createElement("html");
+        iframe.appendChild(doc);
+        doc.innerHTML = html;
```
farmdawgnation: Just a reminder to be **very** careful about where the `html` value in this method comes from. My understanding is that our risk here is pretty minor as `doc` and`iframe` are never attached to the page, but whenever I see it in a PR I gotta mention it.

(Old habits die hard.)

web/webkit/src/main/resources/toserve/lift.js
farmdawgnation: Because Javascript, this change is backward compatible. I don't think we have to worry too much about this change so long as the `dataType` check works as we expect.

web/webkit/src/main/resources/toserve/lift.js
joescii: Yeah, this is fundamental to the design of this feature. Essentially we reload the page from Lift, and steal all of the comet IDs from the resulting HTML.

web/webkit/src/main/resources/toserve/lift.js
joescii: It's declared up at the top at [line 20](https://github.com/lift/framework/blob/4c1e2ad004282b12f38db6dae1b954e508f4506a/web/webkit/src/main/resources/toserve/lift.js#L20)

web/webkit/src/main/resources/toserve/lift.js
joescii: Do you think we should hard code a refresh here, or make it a `LiftRules` setting?

Shadowfiend:
> Side note

❗️

joescii:
@farmdawgnation I dig that idea. So basically something like
```
def enableCometRehydration(): Unit = {
  LiftRules.redirectAsyncOnSessionLoss = false
  LiftRules.noCometSessionCmd.default.set(() => JsCmds.Run("lift.rehydrateComets()"))
}
```

farmdawgnation:
Yup!

farmdawgnation:
@joescii Bump! We have about 10 days to get this in for the RC! :)

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -836,6 +836,16 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
...
+  /**
+    * Adjusts LiftRules to allow a page's comets to "rehydrate" (i.e. reconnect to the page) after a server restart.
+    * This works by reloading the page into an iframe and stealing all of the comets. When the refresh occurs,
+    * new comet actors will be instantiated, so beware that any state therein will NOT be reinstated.
+    */
+  def enableCometRehydration(): Unit = {
+    LiftRules.redirectAsyncOnSessionLoss = false
+    LiftRules.noCometSessionCmd.default.set(() => JsCmds.Run("lift.rehydrateComets()"))
+  }
+
```
joescii: @farmdawgnation @Shadowfiend I feel like the hardest part to get right here is the comment...

farmdawgnation: I think that the comment strikes the right chord overall. The one guidance we might add would be to say that you should code applications depending on this behavior to be able to reconstruct comet state from cookies or from information available from the client (e.g. JS state, state stores, etc)

If you feel like that would be a good addition I say add it. Otherwise, I'm going to approve this PR.

joescii: Comment updated

farmdawgnation:
it's happening

PR #1771, opened 2016-02-14
Trim page-specific javascript

joescii:
This PR defines `JsCmds.trimNoops(JsCmd):JsCmd` along with helper function `JsCmds.toList(JsCmd):List[JsCmd]` which is now called when we assemble the page-specific JS to remove a large number of unnecessary newlines.

See ML thread [here](https://groups.google.com/forum/#!topic/liftweb/ojR6TfrZhIQ).

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftMerge.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftMerge.scala
@@ -57,7 +57,7 @@ private[http] trait LiftMerge {
       S.jsToAppend() ++
       List(eventJs)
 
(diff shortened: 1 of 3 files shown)


Shadowfiend:
This all seems like a lot of work… Why don't we just make `CmdPair` ignore `Noop`s in its `toJsCmd`?

joescii:
I'm glad you thought of that too. I did it this cumbersome way originally thinking we should be conservative and only modify the behavior for when we create the page specific js. The more I thought about it, the more I felt there is no reason to ever think someone would miss a `Noop`'s blank line. I'll whip up another PR next chance I get.

Shadowfiend:
Great minds… Great minds :)

PR #1772, opened 2016-02-20
Filter out empty lines in CmdPair.toJsCmd

joescii:
This PR filters out all empty lines in `CmdPair.toJsCmd`.  The `CmdPair` case class is what is used when combining `JsCmd`s with `&`.  The extra empty lines are particularly noticeable when we produce the page-specific JS.  In a simple test page, I observed that we produced a 155-line JS file with only 4 non-empty lines.  While this will be slightly more performant, the biggest gains IMO are when we need to examine any JS produced while developing.

Per this [ML thread](https://groups.google.com/forum/#!topic/liftweb/ojR6TfrZhIQ), we decided that rather than write [a lot of code](https://github.com/lift/framework/pull/1771) for only the page-specific JS case, we would always toss out empty lines in this PR.

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/js/JsCommands.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/js/JsCommands.scala
@@ -798,7 +798,7 @@ object JsCmds {
     def toJsCmd: String = {
       val acc = new ListBuffer[JsCmd]()
       appendDo(acc, left :: right :: Nil)
-      acc.map(_.toJsCmd).filterNot(_.isEmpty).mkString("\n")
+      acc.map(_.toJsCmd).mkString("\n")
     }
 
     @scala.annotation.tailrec
@@ -806,6 +806,7 @@ object JsCmds {
       cmds match {
         case Nil =>
         case CmdPair(l, r) :: rest => appendDo(acc, l :: r :: rest)
+        case `_Noop` :: rest => appendDo(acc, rest)
         case a :: rest => acc.append(a); appendDo(acc, rest)
       }
     }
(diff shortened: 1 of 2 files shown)


Shadowfiend:
You'll have to update some of the `HtmlNormalizer` specs that were assuming the newlines would still be there, but this is lookin' good.

web/webkit/src/main/scala/net/liftweb/http/js/JsCommands.scala
```diff
@@ -793,12 +793,12 @@ object JsCmds {
...
 
   case class CmdPair(left: JsCmd, right: JsCmd) extends JsCmd {
-    import scala.collection.mutable.ListBuffer;
+    import scala.collection.mutable.ListBuffer
 
     def toJsCmd: String = {
       val acc = new ListBuffer[JsCmd]()
       appendDo(acc, left :: right :: Nil)
-      acc.map(_.toJsCmd).mkString("\n")
+      acc.map(_.toJsCmd).filterNot(_.isEmpty).mkString("\n")
```
Shadowfiend: Actually, rather than do this, why not change `appendDo` to simply never append Noops? It avoids an extra pass through the list if nothing else, though I doubt there are typically large enough JsCmd combinations for that to make a big difference.

joescii: DONE!

Shadowfiend:
Nailed it. :+1: looks good to me!

farmdawgnation:
:+1: :100:

farmdawgnation:
This is something that has long annoyed me. Glad to see it fixed!

PR #1773, opened 2016-02-24
Dependency version bumps in advance of 3.0-RC1.

farmdawgnation:
Okay here we go!

These are version bumps I want to submit for 3.0-RC1. They spawn a few deprecation warnings in the compile cycle so it may be worth taking care of those before merging.

AFAICT these are the highest version bumps I can make without breaking the compile. I also bumped the version of Scala while I was at it.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,9 +12,9 @@ startYear in ThisBuild             := Some(2006)
 
 organizationName in ThisBuild      := "WorldWide Conferencing, LLC"
 
-scalaVersion in ThisBuild          := "2.11.4"
+scalaVersion in ThisBuild          := "2.11.7"
 
-crossScalaVersions in ThisBuild    := Seq("2.11.4")
+crossScalaVersions in ThisBuild    := Seq("2.11.7")
 
 libraryDependencies in ThisBuild <++= scalaVersion {sv => Seq(specs2, scalacheck, scalatest) }
 
--- a/core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
+++ b/core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
@@ -336,7 +336,7 @@ class BoxSpec extends Specification with ScalaCheck with BoxGenerator {
 
   "A Box equals method" should {
 
-    "return true with comparing two identical Box messages" in check {
+    "return true with comparing two identical Box messages" in prop {
       (c1: Box[Int], c2: Box[Int]) => (c1, c2) match {
         case (Empty, Empty) => c1 must_== c2
         case (Full(x), Full(y)) => (c1 == c2) must_== (x == y)
@@ -407,12 +407,12 @@ trait BoxGenerator {
 
   implicit def genThrowable: Arbitrary[Throwable] = Arbitrary[Throwable] {
     case object UserException extends Throwable
-    value(UserException)
+    const(UserException)
   }
 
   implicit def genBox[T](implicit a: Arbitrary[T]): Arbitrary[Box[T]] = Arbitrary[Box[T]] {
     frequency(
-      (3, value(Empty)),
+      (3, const(Empty)),
       (3, a.arbitrary.map(Full[T])),
       (1, genFailureBox)
     )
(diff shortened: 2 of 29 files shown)


fmpwizard:
Looks like I didn't push to master the bump of common file upload, let me do that later tonight before this or gets in.

fmpwizard:
disregard my comment, there have been other commits to master after I made the bump internally, and I only changed the version number so merging this one in is just fine.

karma4u101:
Hi Matt
I notice you did not bump the servlet_api version to 3.0 is there any particular reason for it or just a oversight ?

farmdawgnation:
javax.servlet.servlet-api looks to only have a 3.0 alpha. Has package identifier changed for a 3.0 final?

eltimn:
It looks like maybe it was moved to: http://mvnrepository.com/artifact/javax.servlet/javax.servlet-api

eltimn:
Should we be using 3.1.0?

karma4u101:
Yea if 3.1 seem like a good candidate (for JavaEE 7) if it dose not break things, what do you all think ?

farmdawgnation:
I think we should take advantage of a major version release and do it. Worst case scenario is we bump it back down during RC because it causes issues.

karma4u101:
Good point, 2.5 is more than 10 years old (September 2005). 3.0 is from December 2009 and 3.1 is from May 2013 so it would be a huge time leap forward ;)

joescii:
Regarding servlet, we should also bump our starter kit's `web.xml`.  That's what caused [the heroku issues I was having](https://groups.google.com/forum/#!topic/liftweb/iRtiEck4s_g).

farmdawgnation:
I'm working on the Servlet bump now.

farmdawgnation:
Uh there are quite a lot of compile errors that resulted from this in the test kit.

And quite a lot of:

```
class MockServletContext needs to be abstract, since: it has 26 unimplemented members.
```

:disappointed:

farmdawgnation:
Alright, I fixed all of the type errors in return types. It looks like the new Servlet API using more restrictive types (yay!) but I had to stop before I could provide sensible implementations for the other stuff since I'm sick today. :mask:

If this doesn't break any tests, then I'll look at getting on a bit later and actually finishing out the implementations we need.

Shadowfiend:
(I'm hacking away at scalaz issues… More to come.)

Shadowfiend:
Just pushed fixes for Scalaz issues, which include a bump to specs2 3.7 (latest) and Scalaz to 7.2.0 (latest). Had to tweak a few things along the way, including correct run mode detection for the latest specs2 and some implicit- and scalacheck-related issues.

Shadowfiend:
Ok, I'm down to merge this. I think we should mention when we cut RC1 that support for Servlet 3.1 is partial, and that we'd like to hear feedback on what features that we haven't implemented are important for folks.

If there are no objections, I'll merge tomorrow.

fmpwizard:
+1

On Thu, Mar 10, 2016 at 1:46 PM, Antonio Salazar Cardozo <
notifications@github.com> wrote:

> Ok, I'm down to merge this. I think we should mention when we cut RC1 that
> support for Servlet 3.1 is partial, and that we'd like to hear feedback on
> what features that we haven't implemented are important for folks.
>
> If there are no objections, I'll merge tomorrow.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/lift/framework/pull/1773#issuecomment-194997248>.
>



-- 
Diego Medina
Lift/Scala Consultant
diego@fmpwizard.com
http://blog.fmpwizard.com/

Shadowfiend:
Alllllll righty. In we go! 👍

karma4u101:
@farmdawgnation @Shadowfiend 
I just noticed a test regression but I am not totally sure this PR is the one affecting some tests but it seem like it could be (sry for being lazy and not checking it thoroughly).
 
I notice both in my own test and lifts FlexMenuBuilderSpec test that using ".... withSFor(url) in {...}" seems to silently drop the entire test. 

The current output from Lifts FlexMenuBuilder Specification, looks like the following, it has 0 example(s) and that's a bit strange it should have 4. 

[info] FlexMenuBuilder Specification
[info] 
[info] FlexMenuBuilder should
[info] 
[info] 
[info] Total for specification FlexMenuBuilder Specification
[info] Finished in 0 ms
[info] 0 example, 0 failure, 0 error

Shadowfiend:
Looking into this… Maybe the way that I fixed `WebSpec` for the latest specs2 isn't quite right.

karma4u101:
Awesome !

Issue #1776, opened 2016-03-03
Props.whereToLook does not work in 3.0-M8


serioga:
There is a change between 3.0-M6 and 3.0-M8 which broke `whereToLook` functionality of `Props`.

The var `providers` in https://github.com/lift/framework/blob/master/core/util/src/main/scala/net/liftweb/util/Props.scala#L397 force to initialize lazy val `props` during initialization of the object `Props`. As result it is not possible to reassign `whereToLook` before initialization of `props`.

This bug is critical for us.

joescii:
Thanks for bringing this to our attention. However per the [Ticket and Pull Request Guidelines](https://github.com/lift/framework/blob/master/CONTRIBUTING.md#ticket-and-pull-request-guidelines) we ask that all issues be first discussed on the [Lift mailing list](https://groups.google.com/forum/#!forum/liftweb).  Please discuss there and we will reopen this issue if needed.

fmpwizard:
reopening as some of us talked about fixing this before RC1

Shadowfiend:
Incidentally, I have a fix for this and a couple of other tiny nits for RC1. Will open a PR tonight! :)

fmpwizard:
awesome!

Shadowfiend:
Incidentally, for us Americans, I dub this the Independence Issue.

Shadowfiend:
Should be fixed in the snapshot!

PR #1779, opened 2016-03-15
RC Fixes

Shadowfiend:
Couple of fixes so we can get the RC out the door:
 - Fix WebSpec's broken fragments (see the tail discussion on #1773).
 - Props respects `whereToLook` again (see #1776) and now has a test
   verifying that this works as expected.

Also some small deprecation, spec, and scaladoc-related fixes.

### DIFF
--- a/core/util/src/main/scala/net/liftweb/util/Props.scala
+++ b/core/util/src/main/scala/net/liftweb/util/Props.scala
@@ -395,7 +395,23 @@ private[util] trait Props extends Logger {
     }
   }
 
-  private[this] var providers: List[PropProvider] = List(props)
+  // None before they've first been looked up/modified/whatever, Some after.
+  // We use this to lazy-load `props`, so that `whereToLook` can be updated
+  // by the user.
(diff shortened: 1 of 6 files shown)


Shadowfiend:
@karma4u101 if you can confirm this fixes the issue you found with `WebSpec` and +1 the PR, I think I'll go for merge.

karma4u101:
@Shadowfiend Yes I tested against my module and this PR fixes the issue. 

I also took a look at the lift-webkit tests and they all (related to WebSec usage) run with executed examples without errors. 

I did however unrelated to this PR notice "java.lang.IllegalStateException: Cannot modify after boot" issues in the end of running the lift-webkit tests.

joescii:
@karma4u101: no worries about that exception AFAIK. I've seen it on my last few PRs and it seems harmless.

Shadowfiend:
I think these are in-process jetty boots that are done during tests, but I think they have the same boot setup. Since they're in the same process, two boots blows up, but because the boot setup is the same things still run smoothly.

PR #1783, opened 2016-03-29
Party Foul: Fix JsCmd handling of extracted event JS

### LINKED ISSUE #1782: Eagerly-evaluated toJsCmd can unexpectedly trigger event extraction JS
Lift defines `toJsCmd` as a `def`. However, in many places
we make it a `val` for efficiency purposes: we want to compute
it only once, and up front. In some cases, we do this with `JsCmd`s
that need to call `fixHtml` and friends. In these cases, we now
do event attribute extraction. However, because that extraction
uses `S.appendJs`, if we're doing this in a place where `toJsCmd`
is defined as a `val`, *merely creating the `JsCmd` is enough to
append the JS*, even if the code that uses that HTML never ends
up being sent down. Additionally, this will probably lead to ordering
problems in these cases (where an element could be added after
we attempt to attach its extracted event handler, thus leading to a
lost event).

We should be doing the event extraction and then returning the
resulting JS as part of `fixHtml` and friends, rather than appending
it globally using `S.appendJs`.

See [the original report on the mailing list](https://groups.google.com/d/msg/liftweb/Z4K-79iObu0/4OCRb893AgAJ).

Shadowfiend:
From the commit:

Before, we were using a global appendJs for these events. However, `fixHtmlAndJs`
already provides a mechanism for returning additional JS. We now use it.

The previous behavior would result in extracted events being appended to the
response (page or AJAX) even if the `JsCmd` that invoked `fixHtmlAndJs` was never
actually sent down to the client. This behavior correctly makes the extracted
event JS part of the `JsCmd`, so that it can be sent down only with that `JsCmd`.

Also ran into a bunch of scaladoc warnings that I finally got tired of and :facepunch:ed
them.

Fixes #1782 . Tested on the example in the ML post linked from there.

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/js/JsCommands.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/js/JsCommands.scala
@@ -646,7 +646,13 @@ trait HtmlFixer {
 }
 
 trait JsCmd extends HtmlFixer with ToJsCmd {
-  def &(other: JsCmd): JsCmd = JsCmds.CmdPair(this, other)
+  def &(other: JsCmd): JsCmd = {
+    if (other == JsCmds.Noop) {
+      this
+    } else {
+      JsCmds.CmdPair(this, other)
+    }
+  }
 
   def toJsCmd: String
 
--- a/web/webkit/src/main/scala/net/liftweb/http/js/JsCommands.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/js/JsCommands.scala
@@ -614,7 +614,12 @@ trait HtmlFixer {
       }
 
     import scala.collection.mutable.ListBuffer
-    val lb = ListBuffer[JsCmd](eventJs)
+    val lb =
+      if (eventJs == JsCmds.Noop) {
+        ListBuffer[JsCmd]()
+      } else {
+        ListBuffer[JsCmd](eventJs)
+      }
 
     val revised = ("script" #> nsFunc(ns => {
       ns match {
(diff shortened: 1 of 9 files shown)


web/webkit/src/main/scala/net/liftweb/http/js/JsCommands.scala
```diff
@@ -646,7 +651,13 @@ trait HtmlFixer {
...
 
 trait JsCmd extends HtmlFixer with ToJsCmd {
-  def &(other: JsCmd): JsCmd = JsCmds.CmdPair(this, other)
+  def &(other: JsCmd): JsCmd = {
+    if (other == JsCmds.Noop) {
+      this
+    } else {
+      JsCmds.CmdPair(this, other)
+    }
+  }
```
joescii: I bet this means that [this change](https://github.com/lift/framework/commit/94bfcbfacafe2e5b7d017f4e5d8153f72499f5b1) is no longer needed :)

Shadowfiend: It's still useful for cases where you do `Noop & ...`, e.g. when you use `Noop` for the start of a `fold` or whatever.

At least I think so ;)

joescii: Ah, good point :+1:

fmpwizard:
Considering this was tested on the example and fixed, and the OP on the mailing list also gave it an ok, :+1: 
and merging

PR #1784, opened 2016-04-09
Support call-by-name for Box's openOrThrowException.

### LINKED ISSUE #1781: Call Box.openOrThrowException parameter by name
**Suggestion to the `Box` API:** make  `openOrThrowException` parameter called by name instead of by value. This will allow to avoid calculation of the “justification” expression for full box.

https://github.com/lift/framework/blob/master/core/common/src/main/scala/net/liftweb/common/Box.scala#L421

farmdawgnation:
Closes #1781.

cc @serioga

### DIFF
--- a/core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
+++ b/core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
@@ -310,7 +310,7 @@ class BoxSpec extends Specification with ScalaCheck with BoxGenerator {
       var sideEffect = false
       def sideEffecting = {
         sideEffect = true
-        "This shouldn't have been invoked."
+        "This should have been invoked."
       }
 
       try {
--- a/core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
+++ b/core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
@@ -427,4 +427,3 @@ trait BoxGenerator {
   } yield Failure(msg.mkString, exception, Box(chain.headOption))
 
 }
-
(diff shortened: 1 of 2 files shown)


Shadowfiend:
We should be able to spec-test the call-by-name nature of the parameter by passing something that computes the value + has a side effect, and confirming that the side effect only happens if the `Box` is non-`EmptyBox`.

Lookin' good though!

farmdawgnation:
Added examples as requested.

core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
```diff
@@ -289,6 +305,22 @@ class BoxSpec extends Specification with ScalaCheck with BoxGenerator {
     "define a 'asA' method returning Empty" in {
       Empty.asA[Double] must_== Empty
     }
+
+    "invoke a call-by-name parameter to openOrThrowException" in {
+      var sideEffect = false
+      def sideEffecting = {
+        sideEffect = true
+        "This shouldn't have been invoked."
```
Shadowfiend: May be worth changing this to "this *should* have been invoked" hehe.

farmdawgnation: Sure.

farmdawgnation:
Heh codacy is complaining about duplication in my specs. :unamused:

Shadowfiend:
I'm 👍 for this. Once Lift 3 goes final ofc ;)

Shadowfiend:
Still 👍 here so let's roll 'er in.

PR #1786, opened 2016-04-11
ajaxMultiSelect onChange

ifedayooladapo:
as discussed
http://stackoverflow.com/questions/9024279/lift-ajax-multi-select-box
https://groups.google.com/forum/#!searchin/liftweb/ajaxMultiSelect/liftweb/mxc19m8eFUE/W6DRROUvH64J
https://groups.google.com/forum/#!searchin/liftweb/ajaxMultiSelect/liftweb/Rg6bm0KMo8E/ObiwM8lVpmYJ

simple test
HtmlMultiSelectSnippet.scala
package code.snippet

import net.liftweb.http.js.JsCmd
import net.liftweb.http._
import net.liftweb.util.Helpers._
import net.liftweb.http.SHtml.ajaxMultiSelect

class HtmlMultiSelectSnippet {
  def select_multi (a : List[String]) : JsCmd ={
    println(a)
  }
  def render = {
    val options = List(("0","0"),("1","1"))
    "#ajaxmulti" #> SHtml.ajaxMultiSelect(options, Seq[String]() , select_multi )

  }
}

index.html

<!DOCTYPE html>
<html >
  <head>
  <body data-lift-content-id="main">
    <div id="main" data-lift="surround?with=default;at=content">
		<div data-lift="HtmlMultiSelectSnippet">
		  <div>
			<select id="ajaxmulti"></select>
		  </div>
		</div>
    </div>
  </body>
</html>

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -227,3 +227,7 @@ Arek Burdach
 
 ### Email: ###
 arek.burdach at gmail dot com
+
+### Name ###
+Ifedayo Oladapo
+ifedayo.oladapo at dawnfuel dot com
--- a/contributors.md
+++ b/contributors.md
@@ -230,4 +230,4 @@ arek.burdach at gmail dot com
 
 ### Name ###
 Ifedayo Oladapo
-ifedayo.oladapo at dawnfuel dot com
+ifedayo.oladapo at gmail dot com
--- a/contributors.md
+++ b/contributors.md
@@ -230,4 +230,6 @@ arek.burdach at gmail dot com
 
 ### Name ###
 Ifedayo Oladapo
+
+### Email: ###
 ifedayo.oladapo at gmail dot com
--- a/web/webkit/src/main/scala/net/liftweb/http/SHtml.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/SHtml.scala
@@ -1028,6 +1028,59 @@ trait SHtml extends Loggable {
     }
   }
 
+  def ajaxUntrustedMultiSelect(opts: Seq[(String, String)], deflt: Seq[String], func: List[String] => JsCmd, attrs: (String, String)*): Elem =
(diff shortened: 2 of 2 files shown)


ifedayooladapo:
Can you tell me what's wrong with the commit quality so I can improve it ?

farmdawgnation:
Hi there,

Thank you for the contribution, but there are a few issues with this PR.

First, according to the guidelines set out in our [contributing guidelines](https://github.com/lift/framework/blob/master/CONTRIBUTING.md) we cannot accept major pull requests from non-committers without explicit request for 3rd party solution from a non-committer.

Second, In this case I typically ask that you open a post on [the Mailing List](https://groups.google.com/forum/?fromgroups=#!forum/liftweb) to get a discussing rolling on this issue, but in this case it might make sense to bump one of the existing posts. That would have been the right first step in our community.

Third, this is using someone else's code (from Stack Overflow) without attribution which is almost certainly a violation of [the license that Stack Overflow uses](https://creativecommons.org/licenses/by-sa/3.0/).

Fourth, this isn't really the approach I would prefer us to use in solving this problem. That's an opinion that I would normally express in the Mailing List post _about_ this solution, but you didn't give us an opportunity to do that before submitting the Pull Request, so I'm noting it here.

I'm going to close this Pull Request. Please feel free to follow up on the Mailing List and we can go from there.

farmdawgnation:
Also worth noting, we recently clarified the verbiage in the CONTRIBUTING.md file, so if that was the source of confusion then that's our bad. :)

ifedayooladapo:
Thanks for the feedback

ifedayooladapo:
I've posted here 
https://groups.google.com/forum/?fromgroups=#!searchin/liftweb/ajaxMultiselect/liftweb/mxc19m8eFUE/W6DRROUvH64J

PR #1791, opened 2016-05-07
Add event return normalization to liftVanilla.

Shadowfiend:
Before, returning false from an event handler attached with
liftVanilla.onEvent would not prevent default behavior or stop
propagation. We now check for event.preventDefault and trigger
it and stopPropagation if the handler returned false. We also
return that return value for older browsers to work with.

### DIFF
--- a/web/webkit/src/main/resources/toserve/lift.js
+++ b/web/webkit/src/main/resources/toserve/lift.js
@@ -707,8 +707,22 @@
       if (typeof elementOrId === 'string') {
         element = document.getElementById(elementOrId);
       }
+      
+      // This is a Lift addition to allow return false to properly do
+      // cross-browser preventDefault/stopPropagation/etc work.
+      function normalizeEventReturn(event) {
+        var result = fn(event);
+        if (result === false) {
+          if (typeof event.preventDefault === "function") {
+            event.preventDefault();
+            event.stopPropagation();
+          }
+        }
+        
+        return result;
+      }
 
-      element[add](pre + eventName, fn, false);
+      element[add](pre + eventName, normalizeEventReturn, false);
     },
     onDocumentReady: function(fn) {
       var settings = this, done = false, top = true,


eltimn:
I tried this locally and it works well with form.ajax, but not so well with LiftScreen. With LiftScreen, I'm getting "Cannot read property 'addEventListener' of null" errors in the console, but I'm not sure why. I'll need more time to investigate, but that's in short supply these days, so I'm not sure when I'll be able to get back to this.

A couple of other thoughts:

1. While calling preventDefault and stopPropagation are equivalent to "return false", one could argue that calling just preventDefault is what you really want. In practice, however, I've never really seen this be a problem, but I'd still vote for not calling stopPropagation.

2. IE didn't support preventDefault until IE9 (for earlier versions you needed to set event.returnValue = false). But, I'm not sure how it handled "return false". I'm also not sure if we need to support IE8- or not, though, so this may be moot.

Shadowfiend:
Re: LiftScreen, any chance there's a project I can spin up to see the behavior?

Re: preventDefault/stopPropagation, changing the way it worked when it was an inline attribute would be extremely bad IMO. The fact that it stops propagation has been annoying to me on multiple occasions, but it's the widely-understood behavior across browsers and frameworks, plus it's what the behavior was before we started lifting event attributes out of the DOM. So I'm pretty strongly opposed to changing that.

Finally, regarding IE pre-9, my understanding is that it didn't support `preventDefault`, but *actually returning false* did what we expect it to. If that understanding's wrong, I'll add an `event.returnValue = result` in there.

eltimn:
I don't have a project you can use, but I'm working on one. I'll hopefully have something this afternoon.

Re: stopPropagation - since liftVanilla is new I don't really see why we need to remain backwards compatible, but, maybe, and this could be for a future version, we could change the code that gets inlined from SHtml.a, button, et al, to call a lift.js function that would be customizable. If not, one could always write their one liftVanilla to handle it they way they want.

Re: IE pre-9 I wasn't sure how the event handler handles return false. Other browsers ignore it when using `addEventListener`, but I guess it wouldn't surprise me that IE would behave differently. I'll try to test it when I get that project setup.

Shadowfiend:
> I don't have a project you can use, but I'm working on one. I'll hopefully have something this afternoon.

:thumbsup:

> Re: stopPropagation - since liftVanilla is new I don't really see why we need to remain backwards compatible, but, maybe, and this could be for a future version, we could change the code that gets inlined from SHtml.a, button, et al, to call a lift.js function that would be customizable. If not, one could always write their one liftVanilla to handle it they way they want.

It's not really a matter of backwards compatibility IMO, this is *now* compatibility. Inline event handlers still exist, and they behave this way. Additionally, jQuery event handlers also behave this way. I think that covers probably 70-80% of the way people use event handlers at the moment in the wild… However, it's worth noting that [React no longer stops propagation on false returns](https://facebook.github.io/react/docs/events.html) (see the note near the top there), so certainly this isn't cut-and-dried and may be worth some ML discussion.

> Re: IE pre-9 I wasn't sure how the event handler handles return false. Other browsers ignore it when using addEventListener, but I guess it wouldn't surprise me that IE would behave differently. I'll try to test it when I get that project setup.

Got it. At this point it may be worth going ahead with the RC build and saying lift vanilla has some known issues in 3.0. What do you think?

eltimn:
Yes. I'm fine with doing the RC and working on this for 3.1.

eltimn:
I got a project to test LiftScreen with and quickly realized what the problem was.  We were wrapping a LiftScreen in `form.ajax`. Why I don't know. Seems to work fine without that :smile: 

I'm going to try to fire up a windows vm now to check on how IE handles return values.

eltimn:
I tested with IE8 on Win7 and it seems to work fine, except it won't submit a form.ajax form by hitting the enter key.

I did find another oddity: The click handler that LiftScreen puts on the submit button does not return false, so the result variable in normalizeEventReturn is undefined. Interestingly, though, addEventListener stops the form from submitting via HTTP in this case.

I pushed the demo project to github here: https://github.com/eltimn/lift-vanilla-demo if you want to try it out.

It has the lift.js from this PR and I added some console.log entries for debugging.

Shadowfiend:
Will circle back to this soon. Just to confirm, the issues here are IE8 on Win7 with that project, or was it with all browsers?

eltimn:
@Shadowfiend sorry, but I don't remember. I'll try to test it again soon.

Shadowfiend:
Cool no worries. I may get to it before you, just wanted to get a clearer handle on what the state was :)

farmdawgnation:
@eltimn @Shadowfiend We should make some progress on this if we want to see it in 3.1

eltimn:
At this point, I don't see any reason to support anything less than IE11, which I tested on Win 8.1 and everything works fine.

So, I'm :+1: for this PR

Shadowfiend:
^^ can't argue with that I suppose <_<

Shadowfiend:
Going to go ahead and merge this guy :tada:

PR #1792, opened 2016-05-07
Implement a sendCometActorMessage that works for all comets of a type.

### LINKED ISSUE #1146: LiftSession.sendCometActorMessage to send to all actors by type regardless of name
Currently LiftSession.sendCometActorMessage lets you send to a single actor of a given type, either one with no name or one with a specified name. It would be nice to have a version that sends to _all_ actors by type. This is slightly different from LiftSession.findComet("Type").foreach(_ ! <message>), because sendCometActorMessage also sends messages to any comet actors _created during this request cycle_. It is this latter functionality that we are looking for, combined with sending existing actors of the type a message.

See https://groups.google.com/d/topic/liftweb/TfbhOCVdq34/discussion for more.

farmdawgnation:
Closes #1146.

This PR implements a variant of `sendCometActorMessage` that will send to all comets of a particular type, regardless of whether or not they have a name or what their name is. I also chose a better name for `setupComet` and deprecated the version of it that existed prior to this PR.

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
@@ -165,7 +165,7 @@ object LiftSession {
    * Cache for findSnippetClass lookups.
    */
   private val snippetClassMap = new ConcurrentHashMap[String, Box[Class[AnyRef]]]()
-  
+
   /*
    * Given a Snippet name, try to determine the fully-qualified Class
    * so that we can instantiate it via reflection.
@@ -428,7 +428,8 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
   private[http] object deferredSnippets extends RequestVar[HashMap[String, Box[NodeSeq]]](new HashMap)
 
   /** Messages saved for CometActors prior to their creation */
-  private object cometPreMessages extends SessionVar[Vector[(CometId, Any)]](Vector.empty)
+  private object cometPreMessagesById extends SessionVar[Vector[(CometId, Any)]](Vector.empty)
+  private object cometPreMessagesByType extends TransientRequestVar[Vector[(String, Any)]](Vector.empty)
 
 
   private[http] def startSession(): Unit = {
@@ -2332,6 +2333,26 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
    * @param name the optional name of the CometActor
    * @param msg the message to send to the CometActor
    */
+  def sendCometActorMessage(theType: String, msg: Any) {
+    testStatefulFeature {
+      findComet(theType).foreach(_ ! msg)
+      setupComet(theType, msg)
+    }
+  }
+
+  /**
+   * This method will send a message to a CometActor, whether or not
+   * the CometActor is instantiated.  If the CometActor already exists
+   * in the session, the message will be sent immediately.  If the CometActor
+   * is not yet instantiated, the message will be sent to the CometActor
+   * as part of setup (@see setupComet) regardless of if it is created as part
+   * of the current HTTP request/response cycle. We plan to restrict this to
+   * the current request/response cycle in the future (as that is the intended beavior).
+   *
+   * @param theType the type of the CometActor
+   * @param name the optional name of the CometActor
+   * @param msg the message to send to the CometActor
+   */
   def sendCometActorMessage(theType: String, name: Box[String], msg: Any) {
     testStatefulFeature {
       findComet(theType, name) match {
@@ -2344,9 +2365,18 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
   /**
    * Allows you to send messages to a CometActor that may or may not be set up yet
    */
+  def setupComet(cometType: String, msg: Any) {
+    testStatefulFeature {
+      cometPreMessagesByType.atomicUpdate(_ :+ cometType -> msg)
+    }
+  }
+
+  /**
+   * Allows you to send messages to a CometActor that may or may not be set up yet
+   */
   def setupComet(cometType: String, cometName: Box[String], msg: Any) {
     testStatefulFeature {
-      cometPreMessages.atomicUpdate(_ :+ CometId(cometType, cometName) -> msg)
+      cometPreMessagesById.atomicUpdate(_ :+ CometId(cometType, cometName) -> msg)
     }
   }
 
@@ -2443,10 +2473,10 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
     val cometInfo = CometId(creationInfo.cometType, creationInfo.cometName)
 
     testStatefulFeature {
-      val existingComet = Box.legacyNullTest(nasyncComponents.get(cometInfo))
+      val existingOrNewComet = Box.legacyNullTest(nasyncComponents.get(cometInfo)).asA[T] or newCometFn(creationInfo)
 
-      (existingComet.asA[T] or newCometFn(creationInfo)).map { comet =>
-        cometPreMessages.atomicUpdate(_.filter {
+      existingOrNewComet.map { comet =>
+        cometPreMessagesById.atomicUpdate(_.filter {
           // Pass messages for this comet on and remove them from pending list.
           case (info, message) if info == cometInfo =>
             comet ! message
@@ -2454,6 +2484,11 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
           case _ =>
             true
         })
+
+        cometPreMessagesByType.is.collect {
+          case (typeForMessage, msg) if typeForMessage == creationInfo.cometType =>
+            comet ! msg
+        }
 
         comet
       }
@@ -2534,7 +2569,7 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
     val attemptedComet = tryo(buildWithNoArgConstructor) or tryo(buildWithCreateInfoConstructor)
 
     attemptedComet match {
-      case fail @ Failure(_, Full(e: java.lang.NoSuchMethodException), _) => 
+      case fail @ Failure(_, Full(e: java.lang.NoSuchMethodException), _) =>
         val message = s"Couldn't find valid comet constructor for ${cometClass.getCanonicalName}. Comets should have a no argument constructor or one that takes the following arguments: (LiftSession, Box[String], NodeSeq, Map[String,String])."
 
         logger.info(message, e)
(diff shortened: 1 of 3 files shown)


farmdawgnation:
Still need to test this, btw, but this is here if anyone is interested. :)

farmdawgnation:
Removing the work in progress designation here since this now has a test.

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
```diff
@@ -2325,32 +2326,68 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
    * the CometActor is instantiated.  If the CometActor already exists
    * in the session, the message will be sent immediately.  If the CometActor
    * is not yet instantiated, the message will be sent to the CometActor
-   * as part of setup (@see setupComet) if it is created as part
+   * as part of setup (@see queueCometMessage) if it is created as part
```
Shadowfiend: Inline, `@see` doesn't actually work. This could be `(see `[[queueCometMessage]]`)` perhaps? With backticks before and after the square brackets, but I can't do that in inline markdown-land seemingly.

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
```diff
@@ -2325,32 +2326,68 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
...
+        case _ => queueCometMessage(theType, name, msg)
       }
     }
   }
 
   /**
    * Allows you to send messages to a CometActor that may or may not be set up yet
    */
-  def setupComet(cometType: String, cometName: Box[String], msg: Any) {
+  def queueCometMessage(cometType: String, msg: Any) {
```
Shadowfiend: Hmm… What's the difference between this and `sendCometActorMessage`? Should we be unifying the two and deprecating `setupComet` and `sendCometActorMessage` both?

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
```diff
@@ -2325,32 +2326,68 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
...
    * in the session, the message will be sent immediately.  If the CometActor
    * is not yet instantiated, the message will be sent to the CometActor
-   * as part of setup (@see setupComet) if it is created as part
+   * as part of setup (@see queueCometMessage) if it is created as part
    * of the current HTTP request/response cycle.
    *
    * @param theType the type of the CometActor
+   * @param msg the message to send to the CometActor
+   */
+  def sendCometActorMessage(theType: String, msg: Any) {
```
Shadowfiend: Do we want to provide a version of this where the type is a type parameter, similar to `S.findComet[T]`?

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
farmdawgnation: Lawl. Gotcha. Will change.

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
farmdawgnation: It's been awhile since i've looked at this code, but I might have just renamed this as a "man that name sucks" reaction instead of anything deeper. Let me reacquaint myself with this PR and I'll let you know.

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
farmdawgnation: I can't find a `findComet[T]` as you describe...

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
farmdawgnation: Okay, in taking a closer look at this - this one doesn't do the setup automatically.

Hmmm. So this is a way of saying, "Hey don't spin this thing up now, but when it does spin up send it this message." I can see that being somewhat useful with regard to queuing something up for a comet in a redirect that will later be instantiated by a page, right?

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
Shadowfiend: It's [`findOrCreateComet[T]`](https://github.com/lift/framework/blob/master/web/webkit/src/main/scala/net/liftweb/http/S.scala#L844-L870), sorry.

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
Shadowfiend: I see. And `sendCometActorMessage` falls back on this if necessary. Got it!

Shadowfiend: I think we could use clarification in the scaladoc here, as this reads very very similarly to `sendCometActorMessage` without anything indicating what the fundamental difference is. In fact, it feels like the scaladoc could be incorrect---if you call `queueCometMessage` and the comet already existed, will it receive the message?

Also, it feels like we should settle on either `sendCometMessage`/`queueCometMessage` or `sendCometActorMessage`/`queueCometActorMessage`, maybe?

farmdawgnation: > if you call queueCometMessage and the comet already existed, will it receive the message?

Ah, probably not. Right you are, sir.

> Also, it feels like we should settle on either sendCometMessage/queueCometMessage or sendCometActorMessage/queueCometActorMessage, maybe?

I like the former since "Comet" implies "Actor" for Lift's purposes.

farmdawgnation:
@Shadowfiend ping here!

Shadowfiend:
Re: `cometPreMessagesByName`: basically, can those messages fire for comets past the current request cycle? I have a vague recollection of discussing this a long while back.

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
```diff
@@ -2327,13 +2327,13 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
    * the CometActor is instantiated.  If the CometActor already exists
    * in the session, the message will be sent immediately.  If the CometActor
    * is not yet instantiated, the message will be sent to the CometActor
-   * as part of setup (@see queueCometMessage) if it is created as part
+   * as part of setup (see queueCometMessage) if it is created as part
```
Shadowfiend: Was suggesting this be \<backtick\>[[queueCometMessage]]\<backtick\>, so that the generated scaladocs (a) style `queueCometMessage` as monospace and (b) link to the other one.

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
```diff
@@ -2345,15 +2345,31 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
...
    * in the session, the message will be sent immediately.  If the CometActor
    * is not yet instantiated, the message will be sent to the CometActor
-   * as part of setup (@see queueCometMessage) regardless of if it is created as part
+   * as part of setup (see queueCometMessage) if it is created as part
+   * of the current HTTP request/response cycle.
+   *
+   * @param msg the message to send to the CometActor
+   */
+  def sendCometMessage[T](msg: Any)(implicit cometManifest: Manifest[T]): Unit = {
+    val castClass = cometManifest.runtimeClass.asInstanceOf[Class[T]]
```
Shadowfiend: What's this bonus cast for?

Shadowfiend:
Ah, there it is: “ We plan to restrict this to the current request/response cycle in the future (as that is the intended beavior).” (Side note: “behavior” is misspelled here.)

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
```diff
@@ -2345,15 +2345,31 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
...
+   * as part of setup (see queueCometMessage) regardless of if it is created as part
    * of the current HTTP request/response cycle. We plan to restrict this to
    * the current request/response cycle in the future (as that is the intended beavior).
    *
    * @param theType the type of the CometActor
    * @param name the optional name of the CometActor
    * @param msg the message to send to the CometActor
    */
-  def sendCometActorMessage(theType: String, name: Box[String], msg: Any) {
+  def sendCometMessage(theType: String, name: Box[String], msg: Any): Unit = {
```
Shadowfiend: I wouldn't mind if we had a master scaladoc for this overload, then described the other overloads in terms of how they differ from this one (with a link to this one)… What do you think?

Also seems like this should be the last overload in the code, since it's the one most of the other ones pipe through.

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
farmdawgnation: This is probably a bit silly, but this is the way the other typed method did it - and I weighed the likelihood that this was a mistake as lower than the likelihood that there was some magic that made this useful that I just didn't know about. Hah.

Could this provide some assistance to dealing with erasure per chance?

I'm actually scratching my head at this now too. Maybe it was just a fluke that we should fix. heh.

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
farmdawgnation: <_< >_>

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
farmdawgnation: I'm ~ on how to order overloads, to be honest. Is this an actual convention we've followed in Lift that I just haven't noticed? I honestly alternate between doing the most unique overload first and doing the most unique overload last depending on my mood, so I don't have a particular opinion either way. <shrug>

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
Shadowfiend: I'd say it's not a very big deal.

web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
Shadowfiend: I thought it more important to make the scaladocs reference the common overload (imagine that the central implementation is done one way and changed to work a different way; the likelihood someone updates the documentation on all the callers is near 0) than to actually move the code around, honestly :)

farmdawgnation:
@Shadowfiend I've stalled out on this. What are your thoughts on shipping as is?

farmdawgnation:
ship it

PR #1796, opened 2016-05-13
Escape jQuery identifiers

serveace:
As discussed on the forum.
See: https://groups.google.com/forum/#!topic/liftweb/jorUqMmCLDQ


eltimn:
If we're going to fix this, should we fix it for all of the meta characters where this is a problem?

This [SO entry](http://stackoverflow.com/questions/605630/how-to-select-html-nodes-by-id-with-jquery-when-the-id-contains-a-dot) has some good info on the problem.

I think using this is probably the best way to go:

    jQuery(document.getElementById(elementOrId))

Shadowfiend:
I don't hate that solution.

serveace:
Good call.  I'll update my local env and test a little bit.  If it looks good there, I'll update the pr.

fmpwizard:
quick question, why is this marked for 3.1 and not for the next RC?

Shadowfiend:
IIRC, it's not a regression, so there was no reason to hold up a 3.0 final release for it (this was opened after th RC cycle began).

fmpwizard:
ok, and I imagine we don't want to add it now to the next RC in case of a regression caused by this fix, right? (because there will be another RC to handle that one js bug we still have)

Shadowfiend:
Soo I think we can move forward on this. @eltimn it sounds like you had a look here? We good to go?

eltimn:
Yes. Looks good.

Shadowfiend:
All in then!

PR #1798, opened 2016-05-18
Parse CSP report as JSON

serveace:
Chrome sends csp-report as application/csp-report, which breaks default json parsing.

As discussed: https://groups.google.com/forum/#!topic/liftweb/WP-csyV0Vbs

-Austen


eltimn:
:+1: from me. I'm glad you found the problem. I was seeing this as well but didn't figure out what was causing it.

Shadowfiend:
Going to go ahead and sneak this into RC4.

PR #1800, opened 2016-05-23
Migration script/docs

Shadowfiend:
Few things here:
 - Quick migration script that highlights things that aren't marked as
   deprecated in Lift 2.6 but behave significantly differently in Lift 3.0,
   as well as giving some instructions on how to port.
 - Fix up CSS selector transform docs, since they are relevant to the
   above migration (from `bind` to CSS selector transforms, specifically).
 - Add a quick `LiftScreen` migration doc for 2.6 to 3.0.

@pbrant could particularly use your eyes on the migration doc, and would
especially appreciate some thoughts on what effect `formName` might have
if it's left blank, if any. Couldn't tell for sure how important it would be if porting
from the old-style `LiftScreen`.

This doesn't need to be in any built JARs or anything, we'll just release it
alongside the final 3.0.0 release.

### DIFF
--- a/docs/css-selectors.adoc
+++ b/docs/css-selectors.adoc
@@ -24,13 +24,14 @@ transforms it according to a set of rules, producing a final `NodeSeq` with all
 of the transformations applied. This means a CSS Selector Transform is
 ultimately simply a function with signature `(NodeSeq)=>NodeSeq`. CSS Selector
 Transforms consist of three main components:
+
  - The selector
  - The subnode modification rule
  - The transformation function
 
-///
-  Nice to have: graphic that shows a selector transform pointing to each
-///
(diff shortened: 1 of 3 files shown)


eltimn:
I can't comment on the LiftScreen stuff, but the CSS selector doc is excellent. Thanks for doing this.

Shadowfiend:
If I don't hear anything from @pbrant or [Alex](https://groups.google.com/forum/#!topic/liftweb/KDdZqqdaH1w), I'll probably merge this in the next couple of days.

farmdawgnation:
What's dat status?

Shadowfiend:
Forgot to merge, but no word so inbound.

PR #1803, opened 2016-06-28
Drop 'small fix' requirement from contributing doc.

Shadowfiend:
Until now, we required that contributions be small fixes or
documentation fixes. We've decided to drop that limitation
and open Lift to wider contributions, though only committers
can merge.

Looking for a +1 or 3 (or suggestions for changes!) and will
post an announcement to the Lift ML once this is merged, with
similar verbiage.


Shadowfiend:
Kk, in we go!

PR #1804, opened 2016-07-02
Add catch method to lift Promise to enhance compatibility

serveace:
As discussed on the [mailing list](https://groups.google.com/forum/#!topic/liftweb/u3Y8lhI6_D0).

### DIFF
--- a/web/webkit/src/main/resources/toserve/lift.js
+++ b/web/webkit/src/main/resources/toserve/lift.js
@@ -507,6 +507,8 @@
 
         return self;
       };
+	  
+      self['catch'] = self.fail;
 
       self.done = function(f) {
         _doneFuncs.push(f);


farmdawgnation:
Flagging this for 3.1.

Shadowfiend:
I'm down here.

@farmdawgnation since you looked at this initially, you want to sign off and merge when ready? Looks like Austen is already in contributors.md.

farmdawgnation:
:eyes:

farmdawgnation:
Left one comment, otherwise this looks good.

web/webkit/src/main/resources/toserve/lift.js
```diff
@@ -507,6 +507,8 @@
 
         return self;
       };
+	  
+      self['catch'] = self.fail;
```
Shadowfiend: Reserved words [only apply to identifiers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Reserved_word_usage), so `self.catch = self.fail` should work just fine I believe. Worth double-checking in IE8 and maybe 7 though.

farmdawgnation: Given that we haven't heard back from @serveace on this since we commented, I'm not inclined to hold up the merge on a minor stylistic detail. WDYT, @Shadowfiend?

Shadowfiend: I'm in 👍

farmdawgnation:
Shipping this. :+1:

PR #1805, opened 2016-08-17
Fix copy-paste error in Prepend example

mdzw:


docs/css-selectors.adoc
```diff
@@ -238,7 +238,7 @@ Append to children rule: `\*<` or `*+`::
 
 Prepend to children rule: `>*` or `-*`::
   The transformation result will be prepended to the children of the matched
-  element(s). For example, `^ *+` will prepend the results of the transformation
+  element(s). For example, `^ -*` will prepend the results of the transformation
```
Shadowfiend: I don't copy-paste… Stop with those vile accusations… <\_< >_>

Haha, thanks for the fix :)

PR #1806, opened 2016-08-17
Fix lift node example markup

mdzw:
Remove '.' from beginning of class name, add closing </form> tag


Shadowfiend:
Mergin' this guy, since it's just a doc fix so shouldn't affect the stability of the release at all.

PR #1808, opened 2016-09-11
Add `flatten` to Box: https://github.com/lift/framework/issues/1765

### LINKED ISSUE #1765: Add `flatten` to `Box`
At the moment, you have to do `flatMap(identity)`. Not the
end of the world, but not ideal either. Would be nice to implement
this directly in the API.

Bhashit:
Added the `flatten` method and added the necessary tests to the specs.

Closes #1765.

### DIFF
--- a/core/common/src/main/scala/net/liftweb/common/Box.scala
+++ b/core/common/src/main/scala/net/liftweb/common/Box.scala
@@ -456,6 +456,13 @@ sealed abstract class Box[+A] extends Product with Serializable{
    */
   def flatMap[B](f: A => Box[B]): Box[B] = Empty
 
+  def flatten[B](implicit ev: A <:< Box[B]): Box[B] = this match {
+    case Full(internal) => ev(internal)
+    case f: Failure => f
+    case Empty => Empty
+  }
+
+
   /**
    * If this `Box` contains a value and it satisfies the specified `predicate`,
    * return the `Box` unchanged. Otherwise, return an `Empty`.
@@ -497,7 +504,7 @@ sealed abstract class Box[+A] extends Product with Serializable{
   def forall(func: A => Boolean): Boolean = true
 
   /**
-   * 
+   *
    * If this `Box` contains a value and it does '''not''' satisfy the specified
    * `f`, return the `Box` unchanged. Otherwise, return an `Empty`.
    */
@@ -648,7 +655,7 @@ sealed abstract class Box[+A] extends Product with Serializable{
    * This method calls the specified function with the specified `in` value and
    * the value contained in this `Box`. If this box is empty, returns the `in`
    * value directly.
-   * 
+   *
    * @return The result of the function or the `in` value.
    */
   def run[T](in: => T)(f: (T, A) => T) = in
@@ -921,7 +928,7 @@ sealed case class Failure(msg: String, exception: Box[Throwable], chain: Box[Fai
     import scala.collection.mutable.ListBuffer
     val ret = new ListBuffer[Throwable]()
     var e: Throwable = exception openOr null
-    
+
     while (e ne null) {
       ret += e
       e = e.getCause
@@ -942,7 +949,7 @@ sealed case class Failure(msg: String, exception: Box[Throwable], chain: Box[Fai
   /**
    * Flatten the `Failure` chain to a List where this Failure is at the head.
    */
-  def failureChain: List[Failure] = 
+  def failureChain: List[Failure] =
     this :: chain.toList.flatMap(_.failureChain)
 
   /**
@@ -1087,7 +1094,7 @@ final case class BoxedBoxOrRaw[T](box: Box[T]) extends BoxOrRaw[T]
  * The `[[BoxOrRaw]]` that represents a raw value.
  */
 final case class RawBoxOrRaw[T](raw: T) extends BoxOrRaw[T] {
-  def box: Box[T] = 
+  def box: Box[T] =
     if (raw.asInstanceOf[Object] ne null) Full(raw) else Empty
 }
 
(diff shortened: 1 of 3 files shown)


core/common/src/main/scala/net/liftweb/common/Box.scala
```diff
@@ -456,6 +456,13 @@ sealed abstract class Box[+A] extends Product with Serializable{
    */
   def flatMap[B](f: A => Box[B]): Box[B] = Empty
 
+  def flatten[B](implicit ev: A <:< Box[B]): Box[B] = this match {
```
Shadowfiend: Let's follow the Scala standard library's convention of naming this implicit `asBox`.

core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
```diff
@@ -141,6 +141,15 @@ class BoxSpec extends Specification with ScalaCheck with BoxGenerator {
     "define a 'flatMap' method transforming its value in another Box. If the value is transformed in an Empty can, the total result is an Empty can" in {
       Full(0) flatMap { x: Int => if (x > 0) Full("full") else Empty } must beEmpty
     }
+    "define a 'flatten' method if it contains another Box. If the inner box is a Full can, the final result is a Full can" in {
```
Shadowfiend: This should read “If the inner box is a Full box, the final result will be identical to that box”, I think.

Shadowfiend: (The word “can” is from when `Box` was named `Can`, a long time ago; haven't had a chance to clean that term out of the whole code base yet.)

core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
```diff
@@ -141,6 +141,15 @@ class BoxSpec extends Specification with ScalaCheck with BoxGenerator {
...
+      Full(Full(1)).flatten must_== Full(1)
+    }
+    "define a 'flatten' method if it contains another Box. If the inner box is a Failure, the final result is a Failure" in {
+      Full(Failure("error", Empty, Empty)).flatten must_== Failure("error", Empty, Empty)
```
Shadowfiend: This is an interesting question. This seems like it diverges from typical Scala collection library behavior; in particular, classic `flatten` behavior preserves the *outer* type, not the inner one:

```scala
@ Vector(List(1,2,3)).flatten
res2: Vector[Int] = Vector(1, 2, 3)
@ Vector(List()).flatten
res3: Vector[Nothing] = Vector()
@ 
```

I think what we want to do is make sure we behave the same as an `identity`-based `flatMap`:

```scala
@ Full(Failure("bam")).flatMap(identity)
res11: Box[Nothing] = Failure(bam,Empty,Empty)
```

So this is indeed the correct behavior based on how we've defined `flatMap`. Whether or not that's “correct”, I think changing the behavior this late in the game would be waay too problematic, so we'll leave it as is for now.

core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
```diff
@@ -141,6 +141,15 @@ class BoxSpec extends Specification with ScalaCheck with BoxGenerator {
...
+    }
+    "define a 'flatten' method if it contains another Box. If the inner box is an Empty can, the final result is an Empty can" in {
```
Shadowfiend: Once again, s/can/box.

Shadowfiend:
We'll have to wait for 3.1, but this looks good 👍

Shadowfiend:
Rockin' work @Bhashit btw!

core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
Bhashit: Thanks @Shadowfiend. That does sound better. I'll push the changes.

core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
Bhashit: @Shadowfiend IMO, In the case of Box/Option, it does make more sense to get the inner instance as that's the Box/Option with the final result (if it exists). And the actual type doesn't change, since we are getting a generic `Box` back.

Bhashit:
Thanks @Shadowfiend for the review. I'll push the necessary changes today.

core/common/src/main/scala/net/liftweb/common/Box.scala
Bhashit: @Shadowfiend Actually, this was taken from scala's `Option#flatten`. Thinking about this a bit, this does serve as evidence here. What do you think?

Shadowfiend: It does indeed serve as evidence, but we also *use* that evidence, so it is not there solely as evidence. I prefer the collections framework's approach in this case, personally.

core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
```diff
@@ -135,20 +135,22 @@ class BoxSpec extends Specification with ScalaCheck with BoxGenerator {
...
-    "define a 'flatten' method if it contains another Box. If the inner box is a Full can, the final result is a Full can" in {
-      Full(Full(1)).flatten must_== Full(1)
-    }
-    "define a 'flatten' method if it contains another Box. If the inner box is a Failure, the final result is a Failure" in {
-      Full(Failure("error", Empty, Empty)).flatten must_== Failure("error", Empty, Empty)
-    }
-    "define a 'flatten' method if it contains another Box. If the inner box is an Empty can, the final result is an Empty can" in {
-      Full(Empty).flatten must_== Empty
+    "define a 'flatten' method if it contains another Box." in {
+      "If the inner box is a Full box, the final result is identical to that box" in {
```
Shadowfiend: 👌👌 nesting for the win.

Shadowfiend:
Ok, merged master here. When travis signs off we're a go!

Shadowfiend:
In we go!

PR #1809, opened 2016-09-11
Escape Call: Add LiftRules.extractInlineJavaScript.

Shadowfiend:
Enabled by default, this can be set to `false` in cases where event extraction
is causing poor or undesirable behavior.

Should be fairly self-explanatory… The only question is whether to default it
to `true` or `false`. Given the major version bump and the desirable aspects
of having this enabled, I'd suggest we go `true` by default. However, if we
prioritize the potentially breaking nature of having it enabled, we could make
a good argument for `false` by default.

I can also post to the list and get input from there if we think that's the best
way to go.

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/HtmlNormalizer.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/HtmlNormalizer.scala
@@ -87,6 +87,7 @@ private[http] final object HtmlNormalizer {
     attributes: MetaData,
     contextPath: String,
     shouldRewriteUrl: Boolean, // whether to apply URLRewrite.rewriteFunc
+    extractInlineJavaScript: Boolean,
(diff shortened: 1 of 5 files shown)


farmdawgnation:
> However, if we prioritize the potentially breaking nature of having it enabled, we could make a good argument for false by default.

So we're adding something potentially breaking in the RC phase? That seems like bad form.

Shadowfiend:
No. The ability to disable this behavior did not exist before this PR. It's potentially breaking with respect to Lift 2.6.

farmdawgnation:
Ah, got it. Yeah, heh. This seems like the kind of thing where the best we can do is big red letters in the release announcement.

Shadowfiend:
Okay, going to merge this so we can cut an RC then!

PR #1811, opened 2016-09-24
remove unused dependencies

xuwei-k:


Shadowfiend:
Per the [contribution guidelines](https://github.com/lift/framework/blob/master/CONTRIBUTING.md), we expect a pull request to be related to an existing post on the [Lift mailing list](https://groups.google.com/forum/#!forum/liftweb) where someone has indicated that a PR should in fact be opened. Please go ahead and start that thread and we can reopen this PR once it's been confirmed that it's good to go.

In this particular case, do list out the dependencies you'd like to remove; some folks who likely monitor the ML more closely that Github will be able to see if anything jumps out as still-needed for whatever reason.

PR #1813, opened 2016-10-08
Req and LiftSession - aware Futures / LAFutures

pdyraga:
Session-aware futures seems to be a recurring pain point in many projects. In the one I work, we often use `Future`s to lazy-load some part of a page, e.g. retrieve a list of users from a database and then apply Lift bindings on them. 

In Lift3 `Future` template binding is straightforward but one thing is still missing: you can’t access session from them which is not so crazy as it might sound. 

Continuing with users from database example, say that we retrieved a `Future[Seq]` of users and now we want to `.map` it to produce bindings that will be later bound to element. In that bindings, we’d like to use `S.?` to render localised user type text. Currently it’s not possible, because session will not be available in thread doing `.map` call.

We already have a PR addressing this issue https://github.com/lift/framework/pull/1730 but there are few things missing there in my opinion:
- lack of `ReqVar` access. I think that we should offer the same access scope as `LazyLoad` that we already have in Lift
- lack of session access in `Future` / `LAFuture` body
- We shouldn’t use `CommonLoanWrapper` for session context wrapping, because `CommonLoanWrapper` acts lazily which may lead to session wrapping to be done in separate thread which is too late. We need to enrich function with session access eagerly. 

Sample usage of feature implemented in this PR can be:

```
<div id="main" data-lift="HelloWorld.render">
  Future: <span id="scala-future">Time goes here</span><br/>
  LAFuture: <span id="lift-future">Time goes here</span>
</div>
```

```
class HelloWorld {

  def render = {
    "#lift-future *" #> LAFutureWithSession.withCurrentSession {
      Thread.sleep(3000);
      S ? ("general-futureCompleted", date)
    }.map(s => s"$s in request from = ${S.request.map(_.userAgent).openOrThrowException("No request!")}") &
    "#scala-future *" #> FutureWithSession.withCurrentSession {
      Thread.sleep(1000);
      S ? ("general-futureCompleted", date)
    }.map(s => s"$s in request from = ${S.request.map(_.userAgent).openOrThrowException("No request!")}")
  }

  private def date: String = {
    new SimpleDateFormat("yyyy/MM/dd HH:mm:ss").format(new Date())
  }
}
```

Mailing list reference: https://groups.google.com/forum/#!topic/liftweb/qXDBuSTfWnw

### DIFF
--- a/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
+++ b/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
@@ -113,10 +113,13 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler, context: Box[LAFutur
     onComplete(v => v match {
       case Full(v) =>
         Box.tryo(contextFn(v)) match {
-          case Full(successfullyComputedFuture) => successfullyComputedFuture.onComplete(v2 => result.complete(v2))
-          case e: EmptyBox => result.complete(e)
+          case Full(successfullyComputedFuture) =>
+            successfullyComputedFuture.onComplete(v2 => result.complete(v2))
+          case e: EmptyBox =>
+            result.complete(e)
         }
-      case e: EmptyBox => result.complete(e)
+      case e: EmptyBox =>
+        result.complete(e)
     })
     result
   }
--- a/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
+++ b/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
@@ -426,7 +426,7 @@ object LAFuture {
     *
     * This is similar to [[net.liftweb.common.CommonLoanWrapper]], however, it decorates the
     * function eagerly. This way, you can access current thread's state which is essential
-    * to set up e.g. HTTP session wrapper.
+    * to do things like set up a HTTP session wrapper
     */
   trait Context {
     def around[T](fn: () => T): () => T
--- a/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
+++ b/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
@@ -176,7 +176,7 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler, context: Box[LAFutur
     synchronized {
       if (satisfied) {LAFuture.executeWithObservers(scheduler, () => contextFn(item))} else
       if (!aborted) {
-        toDo ::= f
+        toDo ::= contextFn
       }
     }
   }
--- a/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
+++ b/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
@@ -15,7 +15,7 @@
  */
 
 package net.liftweb
-package actor 
+package actor
 
 import common._
 
@@ -24,7 +24,7 @@ import common._
  * A container that contains a calculated value
  * or may contain one in the future
  */
-class LAFuture[T](val scheduler: LAScheduler) {
+class LAFuture[T](val scheduler: LAScheduler = LAScheduler, context: Box[LAFuture.Context] = Empty) {
   private var item: T = _
   private var failure: Box[Nothing] = Empty
   private var satisfied = false
@@ -32,10 +32,6 @@ class LAFuture[T](val scheduler: LAScheduler) {
   private var toDo: List[T => Unit] = Nil
   private var onFailure: List[Box[Nothing] => Unit] = Nil
   private var onComplete: List[Box[T] => Unit] = Nil
-
-  def this() {
-    this(LAScheduler)
-  }
 
   LAFuture.notifyObservers(this)
 
@@ -105,16 +101,18 @@ class LAFuture[T](val scheduler: LAScheduler) {
    * @return a Future that represents the function applied to the value of the future
    */
   def map[A](f: T => A): LAFuture[A] = {
-    val ret = new LAFuture[A](scheduler)
-    onComplete(v => ret.complete(v.flatMap(n => Box.tryo(f(n)))))
+    val ret = new LAFuture[A](scheduler, context)
+    val contextFn = LAFuture.inContext(f, context)
+    onComplete(v => ret.complete(v.flatMap(n => Box.tryo(contextFn(n)))))
     ret
   }
 
   def flatMap[A](f: T => LAFuture[A]): LAFuture[A] = {
-    val ret = new LAFuture[A](scheduler)
+    val ret = new LAFuture[A](scheduler, context)
+    val contextFn = LAFuture.inContext(f, context)
     onComplete(v => v match {
       case Full(v) =>
-        Box.tryo(f(v)) match {
+        Box.tryo(contextFn(v)) match {
           case Full(successfullyComputedFuture) => successfullyComputedFuture.onComplete(v2 => ret.complete(v2))
           case e: EmptyBox => ret.complete(e)
         }
@@ -124,7 +122,7 @@ class LAFuture[T](val scheduler: LAScheduler) {
   }
 
   def filter(f: T => Boolean): LAFuture[T] = {
-    val ret = new LAFuture[T](scheduler)
+    val ret = new LAFuture[T](scheduler, context)
     onComplete(v => ret.complete(v.filter(f)))
     ret
   }
@@ -174,8 +172,9 @@ class LAFuture[T](val scheduler: LAScheduler) {
    * @param f the function to execute on success.
    */
   def onSuccess(f: T => Unit) {
+    val contextFn = LAFuture.inContext(f, context)
     synchronized {
-      if (satisfied) {LAFuture.executeWithObservers(scheduler, () => f(item))} else
+      if (satisfied) {LAFuture.executeWithObservers(scheduler, () => contextFn(item))} else
       if (!aborted) {
         toDo ::= f
       }
@@ -188,10 +187,11 @@ class LAFuture[T](val scheduler: LAScheduler) {
    * @param f the function to execute. Will receive a Box[Nothing] which may be a Failure if there's exception data
    */
   def onFail(f: Box[Nothing] => Unit) {
+    val contextFn = LAFuture.inContext(f, context)
     synchronized {
-      if (aborted) LAFuture.executeWithObservers(scheduler, () => f(failure)) else
+      if (aborted) LAFuture.executeWithObservers(scheduler, () => contextFn(failure)) else
       if (!satisfied) {
-        onFailure ::= f
+        onFailure ::= contextFn
       }
     }
   }
@@ -202,10 +202,11 @@ class LAFuture[T](val scheduler: LAScheduler) {
    * @param f the function to execute on completion of the Future
    */
   def onComplete(f: Box[T] => Unit) {
+    val contextFn = LAFuture.inContext(f, context)
     synchronized {
-      if (satisfied) {LAFuture.executeWithObservers(scheduler, () => f(Full(item)))} else
-      if (aborted) {LAFuture.executeWithObservers(scheduler, () => f(failure))} else
-      onComplete ::= f
+      if (satisfied) {LAFuture.executeWithObservers(scheduler, () => contextFn(Full(item)))} else
+      if (aborted) {LAFuture.executeWithObservers(scheduler, () => contextFn(failure))} else
+      onComplete ::= contextFn
     }
   }
 
@@ -259,11 +260,12 @@ object LAFuture {
    * @tparam T the type
    * @return an LAFuture that will yield its value when the value has been computed
    */
-  def apply[T](f: () => T, scheduler: LAScheduler = LAScheduler): LAFuture[T] = {
-    val ret = new LAFuture[T](scheduler)
+  def apply[T](f: () => T, scheduler: LAScheduler = LAScheduler, context: Box[Context] = Empty): LAFuture[T] = {
+    val ret = new LAFuture[T](scheduler, context)
+    val contextFn = inContext(f, context)
     scheduler.execute(() => {
       try {
-      ret.satisfy(f())
+        ret.satisfy(contextFn())
       } catch {
         case e: Exception => ret.fail(e)
       }
@@ -277,8 +279,8 @@ object LAFuture {
    * @tparam T the type that
    * @return
    */
-  def build[T](f: => T, scheduler: LAScheduler = LAScheduler): LAFuture[T] = {
-    this.apply(() => f, scheduler)
+  def build[T](f: => T, scheduler: LAScheduler = LAScheduler, context: Box[Context] = Empty): LAFuture[T] = {
+    this.apply(() => f, scheduler, context)
   }
 
   private val threadInfo = new ThreadLocal[List[LAFuture[_] => Unit]]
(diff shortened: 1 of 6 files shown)


pdyraga:
cc @joescii @Shadowfiend

farmdawgnation:
Looks like this spec is failing, which seems pretty important :)

net.liftweb.http.LAFutureWithSessionSpec

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
```diff
@@ -105,16 +101,18 @@ class LAFuture[T](val scheduler: LAScheduler) {
    * @return a Future that represents the function applied to the value of the future
    */
   def map[A](f: T => A): LAFuture[A] = {
-    val ret = new LAFuture[A](scheduler)
-    onComplete(v => ret.complete(v.flatMap(n => Box.tryo(f(n)))))
+    val ret = new LAFuture[A](scheduler, context)
```
farmdawgnation: While we're in here can we give `ret` and `f` a better name? :angel:

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
```diff
@@ -105,16 +101,18 @@ class LAFuture[T](val scheduler: LAScheduler) {
...
+    val contextFn = LAFuture.inContext(f, context)
+    onComplete(v => ret.complete(v.flatMap(n => Box.tryo(contextFn(n)))))
     ret
   }
 
   def flatMap[A](f: T => LAFuture[A]): LAFuture[A] = {
-    val ret = new LAFuture[A](scheduler)
+    val ret = new LAFuture[A](scheduler, context)
```
farmdawgnation: Ditto here.

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
```diff
@@ -406,5 +408,28 @@ object LAFuture {
...
+  private def inContext[A, T](f: (A) => T, context: Box[LAFuture.Context]): (A) => T = {
+    context.map(_.around(f)) openOr f
+  }
+
+  /**
+    * Allows to wrap function in another function providing some additional functionality.
+    * It may choose to execute or not execute that functionality, but should not interpret
+    * or change the returned value; instead, it should perform orthogonal actions that
+    * need to occur around the given functionality. Typical example is setting up DB
+    * transaction.
```
farmdawgnation: I'm finding this a bit hard to follow, but I'm not entirely sure why... it's entirely possible that I haven't had enough caffeine yet, but I'm flagging this anyway. :coffee:

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
```diff
@@ -406,5 +408,28 @@ object LAFuture {
...
+    *
+    * This is similar to [[net.liftweb.common.CommonLoanWrapper]], however, it decorates the
+    * function eagerly. This way, you can access current thread's state which is essential
+    * to set up e.g. HTTP session wrapper.
```
farmdawgnation: This would read better as `which is essential to do things like set up an HTTP session wrapper`

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
```diff
@@ -0,0 +1,48 @@
...
+object LAFutureWithSession {
+
+  /**
+    * Creates `LAFuture` instance aware of current request and session. Each `LAFuture` returned by chained
+    * transformation method (e.g. `map`, `flatMap`) will be also request/session-aware. However, it's
+    * important to bear in mind that initial session or request are not propagated to chained methods. It's required
+    * that current execution thread for chained method has its own request or session available if reading/writing
+    * some data to it as a part of chained method execution.
+    */
+  def withCurrentSession[T](task: => T, scheduler: LAScheduler = LAScheduler): LAFuture[T] = {
```
farmdawgnation: Maybe `LAFuture.withCurrentSession` should be an alias to this function?

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
```diff
@@ -0,0 +1,48 @@
...
+    S.session match {
+      case Full(session) =>
+        withSession(task, scheduler)
+
+      case empty: EmptyBox =>
+        withFailure(empty ?~! "LiftSession not available in this thread context", scheduler)
```
farmdawgnation: Don't we normally throw an exception if something like this attempts to access the session without a proper session context? I'd prefer we keep that behavior consistent. It feels like a situation where an exception is more or less appropriate.

web/webkit/src/test/scala/net/liftweb/http/FutureWithSessionSpec.scala
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    "have access to session variables in chains of flatMap()" withSFor "/" in {
+      SessionVar1("e")
+      SessionVar2("f")
+
+      val future = FutureWithSession.withCurrentSession("d")
+      val mapped = future
+        .flatMap { s => val out = s + SessionVar1.is; Future(out) }
+        .flatMap { s => val out = s + SessionVar2.is; Future(out) }
```
farmdawgnation: Semicolons? In Scala? :P

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
pdyraga: I can rename `ret` to `result` (any other suggestions appreciated) but have no idea for a better name for `f`. I looked at `Box` and `Option` and they also use just `f` as param name. 🤔

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
pdyraga: The problem is that `lift-actor` does not "see" `lift-webkit` - dependency has an opposite direction.

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
pdyraga: I'd say that *sometimes* we throw an exception if method needs a session but it's not available, just like `S ?` do, but when *something like this attempts to access the session without a proper session context* we usually return a `Failure`. This is also similar for `Req`. 

If you take a look at `S`, you'll notice that `S.session` returns a `Box`. As a result, almost all functions requiring session to be available return `Box` as well (e.g. `findOrCreateComet`, `getSessionAttribute`, `runTemplate`, `eval`). Functions returning `Boolean` in `S` usually return `false` if session is not available (e.g. `legacyIeCompatibilityMode`). I'd rather say that `S.?` is an exception here.

Similar situation is for `S.request`. It returns `Box` and as a result other methods requiring request to be available return `Box` as well (`uri`, `queryString`, `getRequestHeader`, `referer`).

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
pdyraga: I'm not a native speaker, so any improvement ideas for this scaladoc are very appreciated :) 

As an "orthogonal" action, such as DB transaction I mean:

```
def addNewUser(name: String, surname: String, email: String): Box[User] = {
  try {
    DB.createNewTransaction()
    val user = User(name, surname, email)
    repository.persist(user)
    eventBus.publish(NewUserCreatedEvent(user.id))
 } finally {
   DB.flushTransaction()
 }
} 
```

^ Here, that `try - finally` block with all `DB` calls is orthogonal to business logic of this method. It shouldn't be here together with business logic code. That's a typical case when we can use `Context` (or `CommonLoanWrapper`). In Java world, it would be probably just an annotation +  aspect.

web/webkit/src/test/scala/net/liftweb/http/FutureWithSessionSpec.scala
pdyraga: I find tests with chained methods more concise when putting entire action in a single line, but if that semicolon hurts your eyes badly I can remove it :wink:

pdyraga:
I've run all specs created in this PR dozen of times but, as usually, with tests running code concurrently, they can give different results on another machine. Good that it failed here, not later. I think I eliminated all possible issues. Some of them were related to the possible race condition that we discuss here https://groups.google.com/forum/#!topic/liftweb/V1pWy14Wl3A and some of them were my fault in implementation.

Anyway, it's ready for another chance now.

Shadowfiend:
Not as a blocker, but I wonder if we also want to make Lift's default handling of Future/LAFuture in the RHS of CSS selector transforms automatically lift the futures into a session/req context... And same for RestHelper.

web/webkit/src/test/scala/net/liftweb/http/FutureWithSessionSpec.scala
farmdawgnation: It's just that it's usually a code smell. 😬 

Also, isn't this basically the same as a for-comp?

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
pdyraga: Good point. I added `S.sessionFuture` as an alias to `LAFutureWithSession.withCurrentSession`.

pdyraga:
> Not as a blocker, but I wonder if we also want to make Lift's default handling of Future/LAFuture in the RHS of CSS selector transforms automatically lift the futures into a session/req context... And same for RestHelper.

I am not sure about this one. Although this could be handy for developers, it might also cause some bugs and confusion. 

For example:

```
// this is fine - right side of `#>` is computed lazily so we 
// can wrap `LAFuture` with session before `SessionVar.is` gets executed
".content" #> LAFuture { SessionVar.is }
```

```
// here, `SessionVar.is` will be executed before we wrap 
// `LAFuture` in session-aware context
val future = LAFuture { SessionVar.is }
".content" #> future
```

The difference in code is small but serious in execution result. I'd rather prefer if users have to do 
```
".content" #> S.sessionFuture { SessionVar.is }
```
or 
```
val future = S.sessionFuture { SessionVar.is }
".content" #> future
```
No automatic wrapping, user has to do it explicitly, but result is predictable.

pdyraga:
@farmdawgnation @Shadowfiend This is ready for another look.

web/webkit/src/test/scala/net/liftweb/http/FutureWithSessionSpec.scala
pdyraga: I forgot to add: semicolons removed :)

pdyraga:
BTW: Codacy check is not 100% accurate. I can't apply some suggested changes because the code will stop compiling. Also, I need to adjust to API that we already have, so switching the order of parameters is not an option here.

farmdawgnation:
Sorry I lost the thread on this a little bit. Hopefully I'll have time to take a look tomorrow. Leaving this in my email inbox until I do. :)

Shadowfiend:
Re: codacy check, yeah, I noticed. I ignored several of them and removed another few patterns. Mostly leaves the `private[this]` suggestions.

Shadowfiend:
And re: auto-wrapping vs being explicit---good point, I'm :100: on the current solution.

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
```diff
@@ -49,12 +45,12 @@ class LAFuture[T](val scheduler: LAScheduler) {
         if (!satisfied && !aborted) {
           item = value
           satisfied = true
-          val ret = toDo
+          val result = toDo
           toDo = Nil
           onFailure = Nil
           onComplete.foreach(f => LAFuture.executeWithObservers(scheduler, () => f(Full(value))))
           onComplete = Nil
-          ret
```
farmdawgnation: :heart:

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
```diff
@@ -105,28 +101,30 @@ class LAFuture[T](val scheduler: LAScheduler) {
...
+    val result = new LAFuture[A](scheduler, context)
+    val contextFn = LAFuture.inContext(f, context)
     onComplete(v => v match {
       case Full(v) =>
-        Box.tryo(f(v)) match {
-          case Full(successfullyComputedFuture) => successfullyComputedFuture.onComplete(v2 => ret.complete(v2))
-          case e: EmptyBox => ret.complete(e)
+        Box.tryo(contextFn(v)) match {
+          case Full(successfullyComputedFuture) => successfullyComputedFuture.onComplete(v2 => result.complete(v2))
+          case e: EmptyBox => result.complete(e)
```
farmdawgnation: Can we break these case statements up so their result is on the next line?

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
```diff
@@ -105,28 +101,30 @@ class LAFuture[T](val scheduler: LAScheduler) {
...
         }
-      case e: EmptyBox => ret.complete(e)
+      case e: EmptyBox => result.complete(e)
```
farmdawgnation: Ditto here.

web/webkit/src/main/scala/net/liftweb/http/FutureWithSession.scala
```diff
@@ -0,0 +1,128 @@
...
+import scala.util.Try
+
+/**
+  * Decorates `Future` instance to allow access to session and request resources. Should be created with
+  * `FutureWithSession.withCurrentSession` that takes the current Lift request and session and make it available
+  * to all transformation methods and initial `Future` execution body. Each transformation method returns
+  * `FutureWithSession`, thus, they can be all chained together.
+  *
+  * It's important to bear in mind that each chained method requires current thread's `LiftSession` to be available.
+  * `FutureWithSession` does _not_ propagate initial session or request to all chained methods.
```
farmdawgnation: What happens if that's not available?

web/webkit/src/main/scala/net/liftweb/http/FutureWithSession.scala
```diff
@@ -0,0 +1,128 @@
...
+  private def apply[T](task: => T)(implicit executionContext: ExecutionContext): FutureWithSession[T] = {
+    S.session match {
+      case Full(_) =>
+        val sessionFn = withCurrentSession(() => task)
+        new FutureWithSession(Future[T](sessionFn()))
+
+      case _ =>
+        new FutureWithSession(Future.failed[T](
+          new IllegalStateException("LiftSession not available in this thread context")
+        ))
```
farmdawgnation: Ah I see. Okay, that's sensible.

web/webkit/src/main/scala/net/liftweb/http/LAFutureWithSession.scala
farmdawgnation: :+1:

farmdawgnation:
LGTM. @joescii @Shadowfiend Do y'all want to take a closer look at this before merge?

Shadowfiend:
Let me try and give it another pass this week, and if I don't get to it then let's just send it in.

pdyraga:
Folks, I'd appreciate if you could clone this branch and run modified tests on your boxes before merging. They pass on my box, they pass on Jenkins, but since these are tests operating on multiple threads having additional checks would boost my confidence about them :)

farmdawgnation:
Seems reasonable. :)

farmdawgnation:
Ran the tests a bunch of times and didn't run into any issues. I'm merging this.

pdyraga:
:heart:

PR #1817, opened 2016-11-16
JZoom: 2x-10x lift-json performance improvements in some cases

Shadowfiend:
At @elemica we've been using the jawn parser to produce the
lift-json AST due to performance issues in the lift-json parser.
I looked at this a little more deeply over the past few months as
I'm loath to depend on additional libraries if it isn't necessary. The
result has been some work that improves lift-json's performance
2x-10x in several of jawn's benchmarks, as well as an additional
one that involves some large JSON that we've seen at work, bringing
us to the same general performance level as jawn. The benchmarks:

```
Benchmark                           Score      Error      Units        Score     Error
Bla25Bench.jawnStringParse         25.240 ±    2.339      ms/op       25.203 ±   3.041
Bla25Bench.liftParse               44.065 ±    1.594      ms/op       25.772 ±   2.963
                                                                                       
CountriesBench.jawnStringParse      3.854 ±    0.223      ms/op        4.050 ±   0.057
CountriesBench.liftParse            7.122 ±    0.137      ms/op        7.247 ±   0.122
                                                                                       
FlipbookBench.jawnStringParse    1233.372 ±  605.274      ms/op     1187.654 ± 538.381
FlipbookBench.liftParse         20308.247 ± 1467.681      ms/op     1267.820 ± 553.363
                                                                                       
Qux2Bench.jawnStringParse          10.817 ±    0.203      ms/op       11.000 ±   0.353
Qux2Bench.liftParse                19.298 ±    0.516      ms/op       11.413 ±   0.171
                                                                                       
Ugh10kBench.jawnStringParse       307.833 ±   24.139      ms/op      321.164 ±  38.761
Ugh10kBench.liftParse            2024.546 ±   89.060      ms/op      337.198 ±  14.643
```

On the left is the original performance, on the right the performance
with this branch.

The main change here is that we aggressively use mutable state and
aggressively reuse data structures and objects wherever possible to
minimize allocation overhead. Additionally, we rework some of the
conditional logic to be compiled down to switch statements that can
turn into jumps or table lookups in JVM bytecode.

Lastly, we drop the fix for the Java 6 double-parsing vulnerability, since
Lift 3 isn't built for Java 6, Java 7+ are patched, and Java 6 itself has
also been patched.

The changes are a bit gnarly, but ultimately 100% worth it.

### DIFF
--- a/core/json/src/main/scala/net/liftweb/json/JsonParser.scala
+++ b/core/json/src/main/scala/net/liftweb/json/JsonParser.scala
@@ -271,15 +271,6 @@ object JsonParser {
     /** Parse next Token from stream.
      */
     def nextToken: Token = {
-      def isDelimiter(c: Char) = {
-        (c: @switch) match {
-          case ' ' | '\n' | ',' | '\r' | '\t' | '}' | ']' =>
-            true
-          case _ =>
-            false
-        }
-      }
-
       def parseString: String = 
         try {
           unquote(buf)
@@ -321,10 +312,7 @@ object JsonParser {
       }
 
       while (true) {
-        buf.next match {
-          case c if EOF == c => 
-            buf.automaticClose
-            return End
+        (buf.next: @switch) match {
           case '{' =>
             blocks.addFirst(OBJECT)
             fieldNameMode = true
@@ -366,11 +354,19 @@ object JsonParser {
             fieldNameMode = true
             blocks.poll
             return CloseArr
-          case c if Character.isDigit(c) || c == '-' =>
+          case c @ ('0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9' | '-') =>
             fieldNameMode = true
             return parseValue(c)
-          case c if isDelimiter(c) =>
-          case c => fail("unknown token " + c)
+          case ' ' |  '\n' | ',' | '\r' | '\t' =>
+            // ignore
+          case c =>
+            c match {
+              case `EOF` =>
+                buf.automaticClose
+                return End
+              case _ =>
+                fail("unknown token " + c)
+            }
         }
       }
       buf.automaticClose
--- a/core/json/src/main/scala/net/liftweb/json/JsonParser.scala
+++ b/core/json/src/main/scala/net/liftweb/json/JsonParser.scala
@@ -171,34 +171,32 @@ object JsonParser {
     else d.doubleValue
   }
 
+  private[this] case class IntermediateJObject(fields: scala.collection.mutable.ListBuffer[JField])
+  private[this] case class IntermediateJArray(bits: scala.collection.mutable.ListBuffer[JValue])
+
   private val astParser = (p: Parser) => {
     val vals = new ValStack(p)
     var token: Token = null
     var root: Option[JValue] = None
 
-    // This is a slightly faster way to correct order of fields and arrays than using 'map'.
-    def reverse(v: JValue): JValue = v match {
-      case JObject(l) => JObject((l.map { field => field.copy(value = reverse(field.value)) }).reverse)
-      case JArray(l) => JArray(l.map(reverse).reverse)
-      case x => x
-    }
-
     def closeBlock(v: Any) {
       @inline def toJValue(x: Any) = x match {
         case json: JValue => json
+        case other: IntermediateJObject => JObject(other.fields.result)
+        case other: IntermediateJArray => JArray(other.bits.result)
         case _ => p.fail("unexpected field " + x)
       }
 
       vals.peekOption match {
         case Some(JField(name: String, value)) =>
           vals.pop(classOf[JField])
-          val obj = vals.peek(classOf[JObject])
-          vals.replace(JObject(JField(name, toJValue(v)) :: obj.obj))
-        case Some(o: JObject) => 
-          vals.replace(JObject(vals.peek(classOf[JField]) :: o.obj))
-        case Some(a: JArray) => vals.replace(JArray(toJValue(v) :: a.arr))
+          val obj = vals.peek(classOf[IntermediateJObject])
+          obj.fields.append(JField(name, toJValue(v)))
+        case Some(o: IntermediateJObject) => 
+          o.fields.append(vals.peek(classOf[JField]))
+        case Some(a: IntermediateJArray) => a.bits.append(toJValue(v))
         case Some(x) => p.fail("expected field, array or object but got " + x)
-        case None => root = Some(reverse(toJValue(v)))
+        case None => root = Some(toJValue(v))
       }
     }
 
@@ -207,9 +205,9 @@ object JsonParser {
         vals.peekAny match {
           case JField(name, value) =>
             vals.pop(classOf[JField])
-            val obj = vals.peek(classOf[JObject])
-            vals.replace(JObject(JField(name, v) :: obj.obj))
-          case a: JArray => vals.replace(JArray(v :: a.arr))
+            val obj = vals.peek(classOf[IntermediateJObject])
+            obj.fields.append(JField(name,v))
+          case a: IntermediateJArray => a.bits.append(v)
           case other => p.fail("expected field or array but got " + other)
       } else {
         vals.push(v)
@@ -220,7 +218,7 @@ object JsonParser {
     do {
       token = p.nextToken
       token match {
-        case OpenObj          => vals.push(JObject(Nil))
+        case OpenObj          => vals.push(IntermediateJObject(scala.collection.mutable.ListBuffer()))
         case FieldStart(name) => vals.push(JField(name, null))
         case StringVal(x)     => newValue(JString(x))
         case IntVal(x)        => newValue(JInt(x))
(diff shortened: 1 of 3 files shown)


Shadowfiend:
Worth noting, we didn't change at all in our performance of `CountriesBench`. Unclear at the moment what exactly makes us worse at that particular benchmark, unfortunately.

core/json/src/main/scala/net/liftweb/json/JsonParser.scala
```diff
@@ -17,6 +17,8 @@
 package net.liftweb
 package json
 
+import scala.annotation.switch
+
```
farmdawgnation: :eyes:

core/json/src/main/scala/net/liftweb/json/JsonParser.scala
```diff
@@ -83,89 +85,118 @@ object JsonParser {
...
-            case '"'  => s.append('"')
-            case '\\' => s.append('\\')
-            case '/'  => s.append('/')
-            case 'b'  => s.append('\b')
-            case 'f'  => s.append('\f')
-            case 'n'  => s.append('\n')
-            case 'r'  => s.append('\r')
-            case 't'  => s.append('\t')
+          buf.substring(intoBuilder = true)
+          (buf.next: @switch) match {
```
farmdawgnation: Is this similar to `@tailrec` in that Scala can do it automatically in some cases but we've broken some rule here that makes that possible, or do you always have to explicitly call out `@switch` to get the benefit?

core/json/src/main/scala/net/liftweb/json/JsonParser.scala
```diff
@@ -83,89 +85,118 @@ object JsonParser {
...
+            case 't'  => builder.append('\t')
             case 'u' =>
+              var byte = 0
+              var finalChar = 0
               val chars = Array(buf.next, buf.next, buf.next, buf.next)
-              val codePoint = Integer.parseInt(new String(chars), 16)
-              s.appendCodePoint(codePoint)
-            case _ => s.append('\\')
+              while (byte < 4) {
+                finalChar = (finalChar << 4) | HexChars(chars(byte).toInt)
```
farmdawgnation: BIT SHIFFFTTTTTTT

I don't think I've had to think about one of these since college. Hahahaha.

farmdawgnation: Can we de-magic-number this `4`?

farmdawgnation: Also is it right to left shift four times for each loop or should we shift by for minus `n` on each loop?

core/json/src/main/scala/net/liftweb/json/JsonParser.scala
```diff
@@ -83,89 +85,118 @@ object JsonParser {
...
+          c = buf.next
       }
-      c = buf.next
     }
     buf.eofIsFailure = false
-    buf.substring
+    forcedReturn match {
+      case null => new String(buf.substring())
+      case _ => forcedReturn
+    }
```
farmdawgnation: I'm inclined to think that an if statement is syntactically prettier in this case, but that could just be me.

core/json/src/main/scala/net/liftweb/json/JsonParser.scala
```diff
@@ -315,11 +358,19 @@ object JsonParser {
             fieldNameMode = true
             blocks.poll
             return CloseArr
-          case c if Character.isDigit(c) || c == '-' =>
+          case c @ ('0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9' | '-') =>
```
farmdawgnation: ORLY. Huh. I guess I haven't ever thought about this having such a significant performance impact, but yeah, makes sense.

core/json/src/main/scala/net/liftweb/json/JsonParser.scala
```diff
@@ -334,20 +385,37 @@ object JsonParser {
...
+    private[this] var segments = scala.collection.mutable.ArrayBuffer(Segments.apply())
     private[this] var segment: Array[Char] = segments.head.seg
     private[this] var cur = 0 // Pointer which points current parsing location
     private[this] var curSegmentIdx = 0 // Pointer which points current segment
 
-    def mark = { curMark = cur; curMarkSegment = curSegmentIdx }
-    def back = cur = cur-1
+    // Mark the current point so that future substring calls will extract the
+    // value from this point to whatever point the buffer has advanced to.
+    final def mark = {
```
farmdawgnation: I'm mildly concerned about the use of `final` here. This is already a private class. What does restricting overrides actually do for us?

core/json/src/main/scala/net/liftweb/json/JsonParser.scala
Shadowfiend: No, I believe it's [like `tailrec`](http://www.scala-lang.org/api/current/scala/annotation/switch.html).

core/json/src/main/scala/net/liftweb/json/JsonParser.scala
Shadowfiend: The `chars` array contains the bytes that you want in the 64-bit `finalChar` in the “correct” placement. So it goes something like this if the values in `chars` are 1234:

```
finalChar     chars (> is current index)
  0000        Array(1,4,2,3)
  0001        Array(>1,4,2,3)
  0014        Array(1,>4,2,3)
  0142        Array(1,4,>2,3)
  1423        Array(1,4,2,>3)
```

Shadowfiend: (Of course, the integers in this case is first looked up in `HexChars` and produce max 4-bit values.)

core/json/src/main/scala/net/liftweb/json/JsonParser.scala
Shadowfiend: Well, syntactic prettiness was the last of my concerns in this PR, but I think you're right that I was getting match-happy here hehe. I don't think there's a bytecode difference between this and a simple `if`.

core/json/src/main/scala/net/liftweb/json/JsonParser.scala
Shadowfiend: I'll nuke these and see if it has any detrimental effect. `final` generally lets the JVM do better optimizations like inlining because it knows there's no possibility of override, but the visibility might be sufficient in this case.

Shadowfiend:
Should deal with all the mentioned issues.

core/json/src/main/scala/net/liftweb/json/JsonParser.scala
farmdawgnation: Ack.

core/json/src/test/scala/net/liftweb/json/ParserBugs.scala
```diff
@@ -26,11 +26,6 @@ object ParserBugs extends Specification {
     parseOpt(""" {"x":"\uffff"} """).isDefined mustEqual true
   }
 
-  "Does not hang when parsing 2.2250738585072012e-308" in {
-    (allCatch.opt(parse(""" [ 2.2250738585072012e-308 ] """)) mustEqual None) and
-      (allCatch.opt(parse(""" [ 22.250738585072012e-309 ] """)) mustEqual None)
-  }
-
```
farmdawgnation: Did we add a replacement to this spec?

core/json/src/main/scala/net/liftweb/json/JsonParser.scala
```diff
@@ -187,35 +226,38 @@ object JsonParser {
...
     def popAny = stack.poll
     def pop[A](expectedType: Class[A]) = convert(stack.poll, expectedType)
     def push(v: Any) = stack.addFirst(v)
     def peekAny = stack.peek
     def peek[A](expectedType: Class[A]) = convert(stack.peek, expectedType)
-    def replace[A](newTop: Any) = stack.set(0, newTop)
+    def replace[A](newTop: Any) = {
+      stack.pop
+      stack.push(newTop)
+    }
```
farmdawgnation: I'm a bit surprised this would make a significant difference.

Shadowfiend: If you mean the switch from `set(0, ...)` to `pop`/`push`, that's not what happened here. We switched to an [`ArrayDeque`](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayDeque.html) instead of a `LinkedList`, and `ArrayDeque` doesn't support `set` operations, so we had to use `pop`/`push` instead.

Shadowfiend: (I switched to ArrayDeque because it shoves things into an array, which is more cache-efficient than a linked list.)

core/json/src/test/scala/net/liftweb/json/ParserBugs.scala
farmdawgnation: Ack.

core/json/src/main/scala/net/liftweb/json/JsonParser.scala
farmdawgnation: Ack

farmdawgnation:
Okay, I think I'm good on this! Shall we ship it?!?!?!?

Shadowfiend:
ship it

farmdawgnation:
shipping it

Shadowfiend:
Fwiw looks like this may have caused some issues in Scala's community build (see #1820 ). Thinking it may be related to concurrency issues with the [segment size spec](https://github.com/lift/framework/pull/1817/files#diff-dc3a751f9bf5147e8f252aec52719433R91). Need to figure out if we want to make that behave better under concurrency, or either toss the spec or make it so it has to be run separately (e.g. moving it to `it` or something).

PR #1819, opened 2016-12-10
Upgrade mongo-java-driver to v3.4.0

### LINKED ISSUE #1814: Upgrade mongo-java-driver to v3

eltimn:
Closes #1814 

I had to remove some functions that were deprecated in Lift 3.0, but I think it's justified here because otherwise, we'd fall pretty far behind on the version of mongo we supported.

### DIFF
--- a/persistence/mongodb-record/src/main/scala/net/liftweb/mongodb/record/field/MongoMapField.scala
+++ b/persistence/mongodb-record/src/main/scala/net/liftweb/mongodb/record/field/MongoMapField.scala
@@ -91,7 +91,7 @@ class MongoMapField[OwnerType <: BsonRecord[OwnerType], MapValueType](rec: Owner
   def asDBObject: DBObject = {
     val dbo = new BasicDBObject
     value.keys.foreach { k =>
-      dbo.put(k.toString, value.getOrElse(k, ""))
+      value.get(k).foreach(v => dbo.put(k.toString, v.asInstanceOf[Object]))
     }
     dbo
   }
--- a/persistence/mongodb-record/src/main/scala/net/liftweb/mongodb/record/field/MongoMapField.scala
+++ b/persistence/mongodb-record/src/main/scala/net/liftweb/mongodb/record/field/MongoMapField.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010-2013 WorldWide Conferencing, LLC
+ * Copyright 2010-2016 WorldWide Conferencing, LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -90,8 +90,10 @@ class MongoMapField[OwnerType <: BsonRecord[OwnerType], MapValueType](rec: Owner
   */
   def asDBObject: DBObject = {
     val dbo = new BasicDBObject
-    value.keys.foreach { k =>
-      value.get(k).foreach(v => dbo.put(k.toString, v.asInstanceOf[Object]))
+    value.keys.foreach { key =>
+      value.get(key).foreach { innerValue =>
+        dbo.put(key.toString, innerValue.asInstanceOf[Object])
+      }
     }
     dbo
   }
--- a/persistence/mongodb/src/test/resources/logging.properties
+++ b/persistence/mongodb/src/test/resources/logging.properties
@@ -13,4 +13,4 @@ java.util.logging.ConsoleHandler.level = INFO
 java.util.logging.ConsoleHandler.formatter = java.util.logging.SimpleFormatter
 
 # Set the default logging level for the named logger
-com.mongodb.level = SEVERE
+org.mongodb.driver.level = WARNING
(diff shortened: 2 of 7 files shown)


persistence/mongodb/src/test/scala/net/liftweb/mongodb/MongoDirectMongoClient.scala
```diff
@@ -50,15 +50,15 @@ class MongoDirectMongoClientSpec extends Specification with MongoTestKit {
...
 
       doc2.put("name", "MongoSession")
       doc2.put("type", "db")
-      doc2.put("count", 1)
+      doc2.put("count", 1: java.lang.Integer)
 
       doc3.put("name", "MongoDB")
       doc3.put("type", "db")
-      doc3.put("count", 1)
+      doc3.put("count", 1: java.lang.Integer)
```
farmdawgnation: Why is the explicit type case needed here?

persistence/mongodb-record/src/main/scala/net/liftweb/mongodb/record/field/MongoMapField.scala
```diff
@@ -91,7 +91,7 @@ class MongoMapField[OwnerType <: BsonRecord[OwnerType], MapValueType](rec: Owner
   def asDBObject: DBObject = {
     val dbo = new BasicDBObject
     value.keys.foreach { k =>
-      dbo.put(k.toString, value.getOrElse(k, ""))
+      value.get(k).foreach(v => dbo.put(k.toString, v.asInstanceOf[Object]))
```
farmdawgnation: There are enough single letter variable names here to be confusing. Can we do something like...

```scala
value.keys.foreach { key =>
  value.get(key).foreach { innerValue =>
    dbo.put(key.toString, innerValue.asInstanceOf[Object])
  }
}
```

?

persistence/mongodb/src/test/resources/logging.properties
```diff
@@ -13,4 +13,4 @@ java.util.logging.ConsoleHandler.level = INFO
 java.util.logging.ConsoleHandler.formatter = java.util.logging.SimpleFormatter
 
 # Set the default logging level for the named logger
-com.mongodb.level = SEVERE
+org.mongodb.driver.level = WARNING
```
eltimn: Most likely because I blindly copied it from somewhere on  the internet.

persistence/mongodb/src/test/scala/net/liftweb/mongodb/MongoDirectMongoClient.scala
eltimn: I'm not quite sure, but the compiler now complains without it:

    type mismatch;
    [error]  found   : Int(1)
    [error]  required: Object
    [error] Note: an implicit exists from scala.Int => java.lang.Integer, but
    [error] methods inherited from Object are rendered ambiguous.  This is to avoid
    [error] a blanket implicit which would convert any scala.Int to any AnyRef.
    [error] You may wish to use a type ascription: `x: java.lang.Integer`.
    [error]       doc.put("count", 1)

persistence/mongodb-record/src/main/scala/net/liftweb/mongodb/record/field/MongoMapField.scala
eltimn: Certainly.

persistence/mongodb/src/test/scala/net/liftweb/mongodb/MongoDirectMongoClient.scala
farmdawgnation: Weird.

persistence/mongodb/src/test/resources/logging.properties
farmdawgnation: < shrug >

farmdawgnation:
Code changes themselves look good. The one thing that concerns me is that if a deprecation wasn't added until 3.0, that probably means we shouldn't remove it until 4.0. However, we're not truly a semver shop. So, balancing that against the fact that we desperately need to update here makes me wonder if we should go ahead and accept these removals.

My vote, personally, is that we merge this. But I'm going to wait a week to press the button to give folks an opportunity to chime in. If there's any concern over this then we can take it to the ML.

farmdawgnation:
Also, thanks @eltimn! :-)

Shadowfiend:
Agreed with the assessment on removing it in 3.1.

Notably, removal of these particular bits in 3.1 is mentioned in the release notes, both for 3.0.0 final and for 3.0-M8 where the original deprecations were introduced.

Shadowfiend:
Let's maybe hold off on merging this until the 2.12 build goes in and we can confirm we don't have any issues on 2.12 as well?

farmdawgnation:
Merged master in with the 2.12 changes. If the build passes I'm content to shove this in!

eltimn:
I'm going to go ahead and merge this.

PR #1823, opened 2017-01-07
Fix cross compiling between 2.11 and 2.12.

farmdawgnation:
Cross-compiling between Scala 2.11 and 2.12 should now work. Looks like the flag we need to work around scala/scala-dev#249 didn't exist in 2.11.

So **hopefully** this finally gets us compiling again!


Shadowfiend:
Rolling this guy in now, because :shipit: .

PR #1824, opened 2017-01-11
Star Crossed Futures: Split FutureWithSession into 2.11 and 2.12 versions.

Shadowfiend:
This adds proper support for `transform(Try)` and `transformWith` to 2.12,
while leaving 2.11 without unsupported operation exceptions. Also split
the specs so that we can test the two new 2.12 methods.


Shadowfiend:
@pdyraga care to have a look since these are your baby?

pdyraga:
All perfect 👌

Shadowfiend:
:100: going to merge this next opportunity if I don't hear anything then!

PR #1829, opened 2017-03-05
Mongo async support

eltimn:
This contains all of Marek's work, plus I added some tests and updated MongoTestKit.

I also redid MongoAsync.defineDb to take a `MongoDatabase` instead of `(MongoClient, String)` . This allows the user to configure the `MongoDatabase` however they like. Because of this, I removed the setting of the CodecRegistry. This will have to be done by the user. An example is provided in the comments.

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -245,3 +245,9 @@ John Marshall
 
 ### Email: ###
 john at themillhousegroup dot com
+
+### Name: ###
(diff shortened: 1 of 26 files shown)


eltimn:
Unless someone wants to take a look at this, I'll merge it in a couple more days.

PR #1832, opened 2017-04-12
Use proper method to check snippet existence in the snippetClassMap

mroocoo:
"contains" searches by value not by key - "containsKey" should be used

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -251,3 +251,9 @@ Marek Żebrowski
 
 ### Email: ###
 marek.zebrowski at gmail dot com
+
+### Name: ###
+Paweł Mruk
+
+### Email: ###
+mroocoo at gmail dot com
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
@@ -180,7 +180,7 @@ object LiftSession {
       //
       // We're using ConcurrentHashMap, so no `getOrElseUpdate` here (and
       // `getOrElseUpdate` isn't atomic anyway).
-      if (! snippetClassMap.contains(name)) {
+      if (! snippetClassMap.containsKey(name)) {
         snippetClassMap.putIfAbsent(name, {
           // Name might contain some relative packages, so split them out and put them in the proper argument of findClass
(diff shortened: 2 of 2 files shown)


farmdawgnation:
Hey @mroocoo thanks for the contribution!

Was there a discussion about this on our mailing list somewhere? I haven't looked at these changes in depth yet, but having some sort of description of the behavior you were seeing that caused you to want to change this would be super helpful for review. =)

mroocoo:
Hi, i don&#39;t know if there is any discussion about this. Some time ago i&#39;ve noticed that snippet classes was unnecessarily searched each time when rendered and as there were many classes in my project it was slow. I made this small change on my local branch and this fixed my issue. And as you look at the code other methods of snippetClassMap (&#39;putIfAbsent&#39;, &#39;get&#39;) access this map by key, but &#39;contains&#39; do not. With the &#39;contains&#39; method the snippet class cache does not work at all and searches classes each time when invoked.  Dnia 12 kwietnia 2017 15:49 Matt Farmer &lt;notifications@github.com&gt; napisał(a):  Hey  github.com @mroocoo  thanks for the contribution!  Was there a discussion about this on our mailing list somewhere? I haven&#39;t looked at these changes in depth yet, but having some sort of description of the behavior you were seeing that caused you to want to change this would be super helpful for review. =)  —  You are receiving this because you were mentioned.  Reply to this email directly,  github.com view it on GitHub , or  github.com mute the thread .

farmdawgnation:
Got it.

Per our contributing guidelines we typically like to have a discussion on the mailing list to validate that there's not some unexpected edge case as a result of this, but given that this is a fairly straightforward change and _looks_ pretty uncontroversial I think we can waive that requirement.

I'll try to make some time tomorrow to test and grok this, unless someone gets to it before me.

farmdawgnation:
@mroocoo Please add your name and email to the contributors file at the root of the project so this meets that requirement of our contribution guidelines.

Shadowfiend:
Upon closer look, this is probably an artifact of the switch from Scala concurrent maps to the Java `ConcurrentHashMap`. Sadly wasn't flagged by the compiler.

Thanks for the contribution!

farmdawgnation:
These changes are approved. Thank you again @mroocoo, this will ship as a part of Lift 3.1.0-M3 on or around the first of June!

PR #1837, opened 2017-04-20
Update usage of FileInput/OutputStream

### LINKED ISSUE #1835: FileInputStream / FileOutputStream Considered Harmful
https://groups.google.com/d/msg/liftweb/wbUbxzQsyJs/O8zvu2lDAQAJ

Replace usage of ```new FileInputStream``` and ```new FileOutputStream``` with newer java.nio equivalents

eltimn:
Fixes #1835 

I haven't been able to directly test this yet, but I started to upgrade our app to use Scala 2.12 and will as soon as that's done.

### DIFF
--- a/core/util/src/main/scala/net/liftweb/util/IoHelpers.scala
+++ b/core/util/src/main/scala/net/liftweb/util/IoHelpers.scala
@@ -87,6 +87,10 @@ trait IoHelpers {
    * Read an entire file into an Array[Byte]
    */
   def readWholeFile(file: File): Array[Byte] = readWholeStream(Files.newInputStream(file.toPath))
+
+  /**
+   * Read an entire file into an Array[Byte]
+   */
   def readWholeFile(path: Path): Array[Byte] = readWholeStream(Files.newInputStream(path))
 
   /**
@@ -110,7 +114,7 @@ trait IoHelpers {
   }
 
   /**
-   * Executes by-name function f and then closes the Cloaseables parameters
+   * Executes by-name function f and then closes the Closeables parameters
    */
   def doClose[T](is: java.io.Closeable*)(f : => T): T = {
     try {
--- a/core/util/src/main/scala/net/liftweb/util/IoHelpers.scala
+++ b/core/util/src/main/scala/net/liftweb/util/IoHelpers.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2011 WorldWide Conferencing, LLC
+ * Copyright 2006-2017 WorldWide Conferencing, LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@ package net.liftweb
 package util
 
 import java.io._
+import java.nio.file.{Files, Path}
 import scala.collection.mutable.ListBuffer
 import ControlHelpers._
 import common._
@@ -85,7 +86,8 @@ trait IoHelpers {
   /**
    * Read an entire file into an Array[Byte]
    */
-  def readWholeFile(file: File): Array[Byte] = readWholeStream(new FileInputStream(file))
+  def readWholeFile(file: File): Array[Byte] = readWholeStream(Files.newInputStream(file.toPath))
+  def readWholeFile(path: Path): Array[Byte] = readWholeStream(Files.newInputStream(path))
 
   /**
    * Read all data from a stream into an Array[Byte]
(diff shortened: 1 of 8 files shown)


core/util/src/main/scala/net/liftweb/util/IoHelpers.scala
```diff
@@ -85,7 +86,8 @@ trait IoHelpers {
   /**
    * Read an entire file into an Array[Byte]
    */
-  def readWholeFile(file: File): Array[Byte] = readWholeStream(new FileInputStream(file))
+  def readWholeFile(file: File): Array[Byte] = readWholeStream(Files.newInputStream(file.toPath))
+  def readWholeFile(path: Path): Array[Byte] = readWholeStream(Files.newInputStream(path))
```
farmdawgnation: Will the scaladoc appear for both methods if they're grouped like this or should we copy it to be above the second instance, too?

eltimn: I'm not sure. I'll try it locally and see what it looks like.

eltimn: It did *not* appear, so I copied it.

farmdawgnation:
@eltimn Are we waiting on a first-hand testing report from you or is this good to merge?

eltimn:
I did get our app upgraded to scala 2.12, but I'd need to publish a few lift modules compiled against 3.1 before I could test this.

I did make these same changes to our app without issues, so I'm good with merging it.

eltimn:
I added a spec to test ```readWholeFile```, which is what I mainly wanted to make sure still worked properly.

farmdawgnation:
Sweet. I say we ship this, then.

web/webkit/src/main/scala/net/liftweb/http/Req.scala
```diff
@@ -284,14 +284,14 @@ FileParamHolder(name, mimeType, fileName)
  * @param localFile The local copy of the uploaded file
  */
 class OnDiskFileParamHolder(override val name: String, override val mimeType: String,
-                            override val fileName: String, val localFile: File) extends
+                            override val fileName: String, val localFile: Path) extends
```
Shadowfiend: Whelp, I missed this guy. This will break all consumers of `OnDiskFileParamHolder`. Is there a way to get from `Path` to `File`? If not, I suggest we revert the changes to `OnDiskFileParamHolder`. If yes, I suggest we rename this parameter `localPath` and provide a `localFile` accessor that returns a `File` as existing code will expect. We can mark that one deprecated if we really want to.

eltimn: There's an apply function with File, so it shouldn't break anything. Also, there's a toPath function on File, and a toFile function on Path that can convert either way.

Shadowfiend: See https://github.com/lift/framework/pull/1852#issuecomment-302056802 . Folks are using `.localFile` directly, and passing it to things that expect a `File`. Moreover, `localFile` is a poor name for something that isn't a `File`, but rather a `Path`, IMO.

I'll work up a PR to make this API-compatible, should only be a couple of lines.

eltimn: I just realized that's not a case class, though. Do I need to change that to a constructor?

eltimn: I see what I did wrong now. Thanks for fixing it.

Shadowfiend: I'll probably add a constructor overload as well, good point. I also missed that it wasn't a case class hah. I doubt anyone is using the constructor directly, but better safe than sorry.

Shadowfiend: Heh, so the reason `FileInputStream` and `FileOutputStream` are “considered harmful” in the Cloudbees post is because they use finalizers. But `OnDiskFileParamHolder` also has a finalizer. I think this is probably still a win, but worth mentioning.

eltimn: The finalize function is on OnDiskFileParamHolder, not on the FileStream class, though. I would assume that's a bit different, although I guess I don't know exactly how having finalizers affect GC.

Shadowfiend: Yeah not sure either. Will keep it in mind in case anything around this comes up in the future.

PR #1838, opened 2017-04-21
Remove deprecated rendering methods

### LINKED ISSUE #1713: Clean up backward-compatbility stuff in JSON serialization.
In #1710 I introduced changes to rid us of our dependence on `scala.text.Document`. I left in some backward compatibility stuff so that existing code will continue to compile, even if with a slew of warnings. For Lift 3.1, we should rip those out altogether.

farmdawgnation:
Closes #1713.

We originally planned to do this in 3.1, so I wanted to go ahead and open the PR against 3.1 since that's what we said we would be doing. However, we've since adopted a more formal release schedule so it's worth asking the question about whether or not we should leave these in until 3.2.

This essentially removes all the back-compat methods that I had start throwing deprecation warnings in 3.0.


farmdawgnation:
Looks like I broke mongodb record. Whoops!

eltimn:
@farmdawgnation I can update the mongo code if you want.

farmdawgnation:
@eltimn If you're in the mood to do it, then go for it! I won't have time this weekend.

farmdawgnation:
If there are no comments here by EOD I'm probably going to merge this.

PR #1839, opened 2017-04-21
Removed requirement of jQuery in liftJQuery definition

### LINKED ISSUE #1797: lift.js does not allow to get rid of JQuery
When using vanilla js there is still a direct reference to jQuery during initialization in `lift.js`:

```
onDocumentReady: jQuery(document).ready,
```

AFAIK it can be fixed by wrapping a call to jQuery in a function like this:

```
onDocumentReady: function () {
  jQuery(document).ready.apply(jQuery, Array.prototype.slice.call(arguments));
},
```

eltimn:
This wasn't discussed on the ML, but OP was right. jQuery was required just to define liftJQuery, making it a requirement to have jQuery loaded, even if you want to use liftVanilla.

The simple solution is to wrap the ```onDocumentReady``` in a function. I tested this with [lift-vanilla-demo](https://github.com/eltimn/lift-vanilla-demo).

Fixes #1797

### DIFF
--- a/web/webkit/src/main/resources/toserve/lift.js
+++ b/web/webkit/src/main/resources/toserve/lift.js
@@ -507,7 +507,7 @@
 
         return self;
       };
-	  
+
       self['catch'] = self.fail;
 
       self.done = function(f) {
@@ -655,7 +655,9 @@
 
       jQuery(elementOrId).on(eventName, fn);
     },
-    onDocumentReady: jQuery(document).ready,
+    onDocumentReady: function(fn) {
+      jQuery(document).ready(fn);
+    },
     ajaxPost: function(url, data, dataType, onSuccess, onFailure) {
       var processData = true,
           contentType = 'application/x-www-form-urlencoded; charset=UTF-8';


Shadowfiend:
Welllll that's pretty silly. 🤦‍♂️  is me…

I tested this and noted that it works, though I had to muck with the project you linked a good bit---it seems to include jQuery currently. There's also actually current no `JSArtifacts` subclass that can be used with liftVanilla and works with AJAX forms and such heh. If you use `JQueryArtifacts` we don't trigger liftVanilla (of course), while `YUIArtifacts` and `ExtCoreArtifacts` both have dependencies on their respective libraries.

I'll file an issue on that part, since I would say liftVanilla is only really partly usable without a vanilla `JSArtifacts` implementation.

eltimn:
Sorry, I meant to mention you needed to edit the project. I was using that to test something else, so I didn't want to commit those changes. I guess I could have put it on a separate branch.

Anyway, is this good to merge?

Shadowfiend:
I think we can call this good to go, yeah. I've filed #1840 to deal with the lack of “no library” `JSArtifacts`.

Shadowfiend:
Going to merge 'er in.

PR #1841, opened 2017-04-24
Eventful Merger: Fix event extraction for pages without head/body elements

### LINKED ISSUE #1807: Inline JS replacement in requests without head and body
https://groups.google.com/forum/#!topic/liftweb/mHcv9chTN5k

The new inline JS replacement removes event handlers in Ajax requests that use snippets to render HTML but don't return a full HTML page. 

When the HTML content returned does not contain a head and body tag it seems  that the inline JS is replaced but the event handling code is not delivered to the client. 

The onclick events are thereby essentially removed from the nodes. 

Excerpt from LiftMerge:

```
if (!hasHtmlHeadAndBody) {
  val fixedHtml =
    normalizeMergeAndExtractEvents(xhtml, HtmlState(mergeHeadAndTail = false)).nodes //what happens to the js?

  fixedHtml.find {
    case e: Elem => true
    case _ => false
  } getOrElse Text("")
} else { ....
```

Shadowfiend:
Fixes #1807.

Short story here was that in cases where a template had a `head` and `body` element, `LiftMerge` would correctly extract events (when `LiftRules.extractInlineJavaScript` was enabled) and append a page script to the outputted HTML that referenced the extracted event handler attachment code. However, if either of those elements was missing, we went through a path that *still extracted the events*, but never attached the page JS reference. As a result, the events in question were lost.

We now correctly generate the page JS and attach a script reference to the end of the element that is returned from `LiftMerge`.

### DIFF
--- a/web/webkit/src/test/scala/net/liftweb/http/LiftMergeSpec.scala
+++ b/web/webkit/src/test/scala/net/liftweb/http/LiftMergeSpec.scala
@@ -14,6 +14,7 @@ import org.mockito.Mockito._
 import common._
 
 import js.JE.JsObj
+import js.pageScript
 
 trait BaseAround extends Around {
   override def around[T: AsResult](test: =>T): Result = {
@@ -62,10 +63,15 @@ class LiftMergeSpec extends Specification with XmlMatchers with Mockito {
   testRules.autoIncludeAjaxCalc.default.set(() => () => (_: LiftSession) => false)
   testRules.excludePathFromContextPathRewriting.default
     .set(
-      () => { in: String => 
+      () => { in: String =>
         in.startsWith("exclude-me")
       }
     )
+
+  val eventExtractingTestRules = new LiftRules()
+  eventExtractingTestRules.javaScriptSettings.default.set(() => () => Empty)
+  eventExtractingTestRules.autoIncludeAjaxCalc.default.set(() => () => (_: LiftSession) => false)
+  eventExtractingTestRules.extractInlineJavaScript = true
 
   "LiftMerge when doing the final page merge" should {
     "merge head segments in the page body in order into main head" in new WithRules(testRules) {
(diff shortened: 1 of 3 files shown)


web/webkit/src/test/scala/net/liftweb/webapptest/ToHeadUsages.scala
```diff
@@ -167,7 +167,7 @@ object ToHeadUsages extends Specification  {
     }
 
     "Exclude from context rewriting" in {
-      val first = http.Req.fixHtml("/wombat",
+      val first = http.Req.normalizeHtml("/wombat",
```
farmdawgnation: What's the meaning of this change?

web/webkit/src/test/scala/net/liftweb/http/LiftMergeSpec.scala
```diff
@@ -441,10 +447,67 @@ class LiftMergeSpec extends Specification with XmlMatchers with Mockito {
...
+      scripts.map(_ \@ "src") must beLike {
+        case scriptSrc :: Nil =>
+          scriptSrc must beMatching("/context-path/lift/page/F[^.]+.js")
+      }
+      pageScript.is must beLike {
+        case Full(response) =>
+          response.js.toJsCmd must contain("tryme()")
+          response.js.toJsCmd must contain("tryyou()")
+      }
+    }
```
farmdawgnation: Tests! :heart:

web/webkit/src/test/scala/net/liftweb/webapptest/ToHeadUsages.scala
Shadowfiend: Per [the relevant commit](https://github.com/lift/framework/pull/1841/commits/002fe6846401e4d0c6fbdaafa24329e1ff5e9909):

> Fix a couple of webkit test deprecation warnings.
>
> These are related to the renaming of Req.fixHtml.

I renamed `fixHtml` to `normalizeHtml` while doing the event extraction work, to bring consistency to the concept of normalizing HTML throughout the system.

farmdawgnation:
Okay, it's been a few days and there have been no further comments here. Let's merge this up!

PR #1842, opened 2017-04-24
Revolting Expressions: Disable event extraction for `fixHtmlFunc` (and `JsExp` by association)

Shadowfiend:
This allows enabling `extractInlineJavaScript` without being worried about
`JsExp` not behaving correctly.

It is a first pass semi-fix for #1801, but a better, more complete fix should
be done that allows inline JS extraction to operate even in these cases. That's
for later, though.


Shadowfiend:
Okies, since we've got two green lights here I'm going to merge this sucker up.

Issue #1831, opened 2017-04-04
We need to handle missing requests in buildDeferredFunction


arigoldx:
(n.b. the following is copied from [this mailing list thread](https://groups.google.com/forum/#!msg/liftweb/foD9cXlMxSQ/Cw9zU1c6AwAJ)).

We ran across an issue where `S.req` returns a `Req` without a `HTTPRequest` inside. Ergo, when we call `HttpRequest.snapshot` in `Req.snapshot`, we get a NPE. The basic question is what to do with `Req` when you're in a comet context.

To fix this, we modified `currentReq` in both definitions of `buildDeferredFunction` so that they filter out null requests:

`val currentReq = S.request.filter(_.request != null).map(_.snapshot)`

The situation is basically:

 - Someone sends a message to a comet.
 - We try to create a deferred function (in this case, it's for the future-with-session
   helpers that Piotr created) for lazy-loading.
 - The deferred function builder tries to snapshot the current HTTPRequest.
 - That object is null (due to being in a comet context), and things go boom.

Shadowfiend:
@arigoldx Any chance you can put in a PR for this? 😇

arigoldx:
https://github.com/lift/framework/pull/1844?

arigoldx:
Is there some action I should take here? Not really sure what the assignment is for..

Shadowfiend:
I think this was just Matt updating the issue to reflect the fact that you've been working on it.

PR #1844, opened 2017-04-28
Handle missing HTTPRequests in S.request

arigoldx:
This fixes https://github.com/lift/framework/issues/1831

**[Mailing List](https://groups.google.com/forum/#!msg/liftweb/foD9cXlMxSQ/Cw9zU1c6AwAJ) thread**:

# Description

# Discussion

Sometimes `S.req` returns a `Req` without a `HTTPRequest` inside. Ergo, when we call `HttpRequest.snapshot` in `Req.snapshot`, we get a NPE.

To fix this, we modify currentReq in both definitions of `buildDeferredFunction` so that they filter out null requests.

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
@@ -1547,7 +1547,7 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
    * had been executed on the thread that created the function.
    */
   def buildDeferredFunction[T](deferredFunction: () => T): () => T = {
-    val currentReq = S.request.map(_.snapshot)
+    val currentReq = S.request.filter(_.request != null).map(_.snapshot)
     val renderVersion = RenderVersion.get
     val requestVarFunc = RequestVarHandler.generateSnapshotRestorer[T]()
 
@@ -1565,7 +1565,7 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
    * request and session context.
    */
   def buildDeferredFunction[A,T](deferredFunction: (A)=>T): (A)=>T = {
-    val currentReq = S.request.map(_.snapshot)
+    val currentReq = S.request.filter(_.request != null).map(_.snapshot)
     val renderVersion = RenderVersion.get
     val requestVarFunc = RequestVarHandler.generateSnapshotRestorer[T]()
 
--- a/web/webkit/src/test/scala/net/liftweb/http/LiftSessionSpec.scala
+++ b/web/webkit/src/test/scala/net/liftweb/http/LiftSessionSpec.scala
@@ -17,8 +17,10 @@
 package net.liftweb
 package http
 
+import scala.concurrent.ExecutionContext.Implicits.global
 import scala.xml.NodeSeq
-import net.liftweb.common.{Full, Empty}
+import net.liftweb.common.{Full, Empty, Failure}
+import net.liftweb.util.Helpers.tryo
 import org.specs2.specification.BeforeEach
 import org.specs2.mutable.Specification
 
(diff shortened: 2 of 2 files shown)


arigoldx:
@farmdawgnation thanks for the comments.

1. I'm not only unsure whether or not `buildDeferredFunction` is the only place we're at risk of hitting a NPE but I'm sure how to rigorously answer that question. I'll start thinking about it :)

2. I'll also look into adding a spec to `LiftSessionSpec`.

'course it may take me a bit :bowtie:.

Shadowfiend:
I suggest if possible we put in the time to add the spec, but perhaps file an issue for searching for additional missing `null` guards. This seems like a win on its own, and would be nice to roll into 3.1.

farmdawgnation:
Agreed, I think I would be happy at the very least adding a test for `buildDeferredFunction` _itself_ or around the calling code that calls `buildDeferredFunction` that demonstrates the issue.

farmdawgnation:
@arigoldx Would be v cool to get this into 3.1 if possible. Have you had any time to take a further look here? If not I can take a look myself. =)

arigoldx:
@farmdawgnation Sadly last week escaped me -- the busy-ness at work & home was too much.

'not sure when 3.1 is planning on going in but I sure as heck wanna get this in ASAP. It's on my plate, albeit my fun-plate, not my work-life plate :)

Are you on Gitter at all? I actually have been working on this from time to time and, as I'm pretty darn new-new to writing tests, I posted some questions there. Gonna give it a bigger whirl today.

All that said, I'm cool with anything..

arigoldx:
Hmm.. this is where I'm at:

```
  "LiftSession when building deferred functions" should {

    "not fail when the underlying container request is null" in {
      val session = new LiftSession("Test Session", "", Empty)

      def stubFunction: () => Int = () => 3

      S.init(Empty, session) {

        val attempt = tryo(session.buildDeferredFunction(stubFunction))

        attempt must beEmpty
      }
    }
  }
```

I've still got the old code in.. the ol write a failing test routine. (So that should fail).

I'm getting a `Full` but I expected a `Failure`: `Full(net.liftweb.http.LiftSession$$Lambda$10412/1748374708@261d80ea)' is not empty (LiftSessionSpec.scala:107)`. Reason is, I thought `buildDeferredFunction` would fail, giving the `tryo` a `Failure`.

What am I missin'?

farmdawgnation:
The changes you made actually ensure that `buildDeferredFunction` does not fail.

This line:

```scala
val currentReq = S.request.filter(_.request != null).map(_.snapshot)
```

Just makes `currentReq` empty if there's no underlying `request` and `_.snapshot` never attempts to run at all. So, your test should instead read:

```scala
        val attempt = session.buildDeferredFunction(stubFunction)

        attempt() mustBe 3
```

based on how you set this test up, I believe.

arigoldx:
Thanks for writin' back!

> The changes you made actually ensure that `buildDeferredFunction` does not fail.

Yep, but I reverted for testing purposes:

> I've still got the old code in.. the ol write a failing test routine. (So that should fail).

That's why I thought that I'd get a `Failure` in the `tryo`. Hmm.. I'll :eyes: in a bit.

arigoldx:
(oops, mis-clicked)

farmdawgnation:
@arigoldx Ah, I see. Let me know if I should take a look.

arigoldx:
yes please.

arigoldx:
@farmdawgnation I think we might be there! My first spec2 (or spec at all) ever! :muscle:

farmdawgnation:
Woohoo. Will review when I'm back at my computer.

farmdawgnation:
Unless there are any objections / concerns surfaced in the next few days, we'll merge this puppy.

farmdawgnation:
if it fits it ships

arigoldx:
IT'S A NEW DAWN

Issue #1780, opened 2016-03-17
lift-json serializes NaN and Infinity doubles (invalidly)


Shadowfiend:
NaN and Infinity should be serialized to null rather than to their
string representations.

farmdawgnation:
Derrrrrrrp.

Shadowfiend:
Should just be a matter of adding a couple of cases to [this bit of `bufRender`](https://github.com/lift/framework/blob/master/core/json/src/main/scala/net/liftweb/json/JsonAST.scala#L949) to match `JDouble(Double.NegativeInfinity | Double.PositiveInfinity | Double.NaN)` and serialize that to `null`, as well as add specs to check for this behavior.

Shadowfiend:
Still interested in doing this, but… This is potentially a breaking change. Currently lift-json's output can result in invalid JSON. But for folks who are feeding JSON directly into JavaScript, they may be relying on the current functionality.

Going to see about opening a list thread on this, perhaps on Monday.

Shadowfiend:
[Posted the question on what the best path forward is on the list](https://groups.google.com/d/msg/liftweb/7p9CNfClhhM/GaQy3HqDAwAJ).

PR #1845, opened 2017-05-02
The Avengers: Infinity NaNny: JSON-serialize +/- infinity and NaN as null.

Shadowfiend:
Before now, we were serializing these as their Scala string equivalents. This
meant only JavaScript when evaling JSON as actual JavaScript could deserialize 
it. lift-json itself couldn't deserialize the serialized special values, in fact.

We also make this behavior customizable by providing a `RenderSettings`
parameter, `DoubleRenderer`, which is charged with taking a `Double` and
rendering it to `String`. Three variants are provided: an as-is renderer that is
identical to pre-3.1 handling of special `Double` values like `NaN`, the new
null-special-values renderer that has the behavior described above, and a
new behavior, `FailToRenderSpecialValues`, which throws an exception when
a special `Double` value is encountered rather than serializing it in any way.

Still an open question on the list:
 - [x] [Should we provide a Formats toggle for this behavior?](https://groups.google.com/d/msg/liftweb/7p9CNfClhhM/GaQy3HqDAwAJ)

This is a fix for #1780.

### DIFF
--- a/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
+++ b/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
@@ -881,17 +881,74 @@ object JsonAST {
      */
     val compactJs = RenderSettings(0, jsEscapeChars)
   }
+
+  /**
+   * Parent trait for double renderers, which decide how doubles contained in
+   * a JDouble are rendered to JSON string.
+   */
+  sealed trait DoubleRenderer {
+    def renderDouble(double: Double): String
+  }
+  /**
+   * A `DoubleRenderer` that renders special values `NaN`, `-Infinity`, and
+   * `Infinity` as-is using `toString`. This is not strictly valid JSON, but
+   * JavaScript can eval it. Other double values are also rendered the same
+   * way.
+   *
+   * Usage is not recommended.
+   */
+  case object RenderSpecialValuesAsIs extends DoubleRenderer {
+    def renderDouble(double: Double): String = {
+      double.toString
+    }
+  }
+  /**
+   * A `DoubleRenderer` that renders special values `NaN`, `-Infinity`, and
+   * `Infinity` as `null`. Other doubles are rendered normally using
+   * `toString`.
+   */
+  case object RenderSpecialValuesAsNull extends DoubleRenderer {
+    def renderDouble(double: Double): String = {
+      if (double.isNaN || double.isInfinity) {
+        "null"
+      } else {
+        double.toString
+      }
+    }
+  }
+  /**
+   * A `DoubleRenderer` that throws an `IllegalArgumentException` when the
+   * special values `NaN`, `-Infinity`, and `Infinity` are encountered. Other
+   * doubles are rendered normally using `toString`.
+   */
+  case object FailToRenderSpecialValues extends DoubleRenderer {
+    def renderDouble(double: Double): String = {
+      if (double.isNaN || double.isInfinity) {
+        throw new IllegalArgumentException(s"Double value $double is not renderable to JSON.")
+      } else {
+        double.toString
+      }
+    }
+  }
   /**
    * RenderSettings allows for customizing how JSON is rendered to a String.
    * At the moment, you can customize the indentation (if 0, all the JSON is
    * printed on one line), the characters that should be escaped (in addition
    * to a base set that will always be escaped for valid JSON), and whether or
    * not a space should be included after a field name.
+   *
+   * @param doubleRendering Before Lift 3.1.0, the three special double values
+   *    NaN, Infinity, and -Infinity were serialized as-is. This is invalid
+   *    JSON, but valid JavaScript. We now default special double values to
+   *    serialize as null, but provide both the old behavior and a new behavior
+   *    that throws an exception upon finding these values. See
+   *    `[[DoubleRenderer]]` and its subclasses for more.
    */
   case class RenderSettings(
     indent: Int,
     escapeChars: Set[Char] = Set(),
-    spaceAfterFieldName: Boolean = false
+    spaceAfterFieldName: Boolean = false,
+    doubleRenderer: DoubleRenderer = RenderSpecialValuesAsNull
   ) {
     val lineBreaks_? = indent > 0
   }
@@ -946,10 +1003,7 @@ object JsonAST {
     case null          => buf.append("null")
     case JBool(true)   => buf.append("true")
     case JBool(false)  => buf.append("false")
-    // These special double values serialize to null; otherwise, we would
-    // generate invalid JSON.
-    case JDouble(nully) if nully.isNaN || nully.isInfinity => buf.append("null")
-    case JDouble(n)    => buf.append(n.toString)
+    case JDouble(n)    => buf.append(settings.doubleRenderer.renderDouble(n))
     case JInt(n)       => buf.append(n.toString)
     case JNull         => buf.append("null")
     case JString(null) => buf.append("null")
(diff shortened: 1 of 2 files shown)


Shadowfiend:
After thinking about this a bit more and reading a little on what jackson is doing (and [a proposal for additional customization in their renderer](https://github.com/FasterXML/jackson-databind/issues/911)), I decided to do a touch more legwork.

First off, we hook into `RenderSettings`, not `Formats`, since this happens during rendering rather than during composition to the JSON AST. Secondly, we provide a customizable `doubleRenderer`, which defaults to the new `null` rendering for special values. And finally, we add *two* variants here, one that renders doubles as-is, and one that will throw an `IllegalArgumentException` when attempting to render a special double value. This allows a little more customization for folks who may want to actually blow up when code attempts to render something their clients wouldn't expect to see.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -881,17 +881,74 @@ object JsonAST {
...
+   * a JDouble are rendered to JSON string.
+   */
+  sealed trait DoubleRenderer extends Function1[Double,String] {
+    def apply(double: Double): String
+  }
+  /**
+   * A `DoubleRenderer` that renders special values `NaN`, `-Infinity`, and
+   * `Infinity` as-is using `toString`. This is not strictly valid JSON, but
+   * JavaScript can eval it. Other double values are also rendered the same
+   * way.
```
farmdawgnation: We should note that if you render JSON using this thing you may not be able to parse that some JSON back with Lift.

core/json/src/test/scala/net/liftweb/json/JsonPrintingSpec.scala
```diff
@@ -34,6 +36,77 @@ object JsonPrintingSpec extends Specification  with JValueGen with ScalaCheck {
...
+        JsonAST.RenderSettings(0, doubleRenderer = JsonAST.FailToRenderSpecialValues)
+      )
+    }
+
+    "render a standard double as is" in {
+      val double = Random.nextDouble
+      render(JDouble(double)) must_== double.toString
+    }
+
+    "render positive infinity as null" in {
```
farmdawgnation: Should read something like `throw an exception if attempting to render positive infinity`

core/json/src/test/scala/net/liftweb/json/JsonPrintingSpec.scala
```diff
@@ -34,6 +36,77 @@ object JsonPrintingSpec extends Specification  with JValueGen with ScalaCheck {
...
+      render(JDouble(Double.PositiveInfinity)) must throwAn[IllegalArgumentException]
+    }
+
+    "render negative infinity as null" in {
```
farmdawgnation: Should read something like `throw an exception if attempting to render negative infinity`

core/json/src/test/scala/net/liftweb/json/JsonPrintingSpec.scala
```diff
@@ -34,6 +36,77 @@ object JsonPrintingSpec extends Specification  with JValueGen with ScalaCheck {
...
+      render(JDouble(Double.NegativeInfinity)) must throwAn[IllegalArgumentException]
+    }
+
+    "render NaN as null" in {
```
farmdawgnation: Should read something like `throw an exception if attempting to render NaN`

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -881,17 +881,74 @@ object JsonAST {
...
+  }
+  /**
+   * A `DoubleRenderer` that throws an `IllegalArgumentException` when the
+   * special values `NaN`, `-Infinity`, and `Infinity` are encountered. Other
+   * doubles are rendered normally using `toString`.
+   */
+  case object FailToRenderSpecialValues extends DoubleRenderer {
+    def apply(double: Double): String = {
+      if (double.isNaN || double.isInfinity) {
+        throw new IllegalArgumentException(s"Double value $double is not renderable to JSON.")
```
farmdawgnation: Can we change this to `Double value $double is not renderable to JSON when FailToRenderSpecialDoubleValues is used for rendering`

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -881,17 +881,74 @@ object JsonAST {
...
+        double.toString
+      }
+    }
+  }
+  /**
+   * A `DoubleRenderer` that throws an `IllegalArgumentException` when the
+   * special values `NaN`, `-Infinity`, and `Infinity` are encountered. Other
+   * doubles are rendered normally using `toString`.
+   */
+  case object FailToRenderSpecialValues extends DoubleRenderer {
```
farmdawgnation: I would prefer the name `FailToRenderSpecialDoubleValues` for clarity when talking about what the class does or when a curious explorer `println`s it.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -881,17 +881,74 @@ object JsonAST {
...
+    def apply(double: Double): String = {
+      double.toString
+    }
+  }
+  /**
+   * A `DoubleRenderer` that renders special values `NaN`, `-Infinity`, and
+   * `Infinity` as `null`. Other doubles are rendered normally using
+   * `toString`.
+   */
+  case object RenderSpecialValuesAsNull extends DoubleRenderer {
```
farmdawgnation: Would prefer this to be named `RenderSpecialDoubleValuesAsNull` for the same reasons I mentioned elsewhere.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -881,17 +881,74 @@ object JsonAST {
...
+  }
+  /**
+   * A `DoubleRenderer` that renders special values `NaN`, `-Infinity`, and
+   * `Infinity` as-is using `toString`. This is not strictly valid JSON, but
+   * JavaScript can eval it. Other double values are also rendered the same
+   * way.
+   *
+   * Usage is not recommended.
+   */
+  case object RenderSpecialValuesAsIs extends DoubleRenderer {
```
farmdawgnation: Would prefer this to be named `RenderSpecialDoubleValuesAsIs` for the same reasons mentioned elsewhere.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
Shadowfiend: We already do, right? “This is not strictly valid JSON, but JavaScript can eval it”. Is there a rewording you'd prefer? Or is it the emphasis on Lift specifically not being able to reparse you're going for?

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
Shadowfiend: I hate this idea, but I think it's probably still a good idea, so I'll do it hehe.

core/json/src/test/scala/net/liftweb/json/JsonPrintingSpec.scala
Shadowfiend: Whoops, yep. Will update these.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
Shadowfiend: I'm loath to use the class name in a string. I can reword this to make it clearer that it's a choice, however.

Shadowfiend:
Hit those requested changes.

core/json/src/main/scala/net/liftweb/json/JsonAST.scala
farmdawgnation: It's the emphasis on Lift specifically not being able to reparse it. Once could read this and interpret it as "This isn't strictly valid JSON even though we can write/parse it" which would not be true.

farmdawgnation:
Your changes look good. Once we land in agreement on the comment we're discussing above, I'm good.

PR #1846, opened 2017-05-03
Early Tutors: Initial pass at a new tutorial living in the repo

Shadowfiend:
Comments/questions/remarks welcome. For comments on specific bits of the documents, ideally leave a line comment in the appropriate part of the diff :)

### DIFF
--- a/docs/getting-started-tutorial/1-view-first-development.adoc
+++ b/docs/getting-started-tutorial/1-view-first-development.adoc
@@ -68,4 +68,4 @@ smart enough to strip all HTML comments in production mode.].
 When it comes to user testing, notice that our view is fully-valid HTML, with
 placeholder data. It is, in effect, a high-fidelity mockup. And now that we've
 got our view sorted out (and, ideally, tested with users), we can start hooking
-up the Lift side.
+up link:2-the-lift-menu-system.adoc[the Lift side].
--- a/docs/getting-started-tutorial/1-view-first-development.adoc
+++ b/docs/getting-started-tutorial/1-view-first-development.adoc
@@ -14,7 +14,7 @@ worry about the the interface between your backend and the HTML only when
 you're working on the backend.
 
 The flip side of view-first development is that it takes some getting used to
-if one is accustomed the typical web MVC framework. The first stop when
+if one is accustomed to the typical web MVC framework. The first stop when
 figuring out what's going on in a typical web MVC setup is the controller. In
 Lift, your first stop is your HTML file. Everything starts in the HTML, and in
 what it is that you want to present to the user. You don't just think about
@@ -29,7 +29,7 @@ user stories:
  - As a chatter, I want to see messages from me and others so that I can keep
    track of the conversation and contribute in context.
 
-To start with, we'll set up a simple chat.html page in our `src/main/webapp`
+To start with, we'll set up a simple `chat.html` page in our `src/main/webapp`
 directory (where all HTML files go). All we really need in there for now is a
 list of chat messages so far, and a box to put our own chat message into. So,
 here's some base HTML to get us going:
--- a/docs/getting-started-tutorial/1-view-first-development.adoc
+++ b/docs/getting-started-tutorial/1-view-first-development.adoc
@@ -1,0 +1,71 @@
+:idprefix:
+:idseparator: -
+:toc: right
+:toclevels: 2
+
+# View-first Development
+
+If you're developing a user-facing web site or application, one of Lift's
+greatest improvements over existing systems is view-first development.
+View-first development thoroughly separates the process of creating the user
+interface from the process of putting data from the system into it, in a way
+that lets you stay focused on users when you're creating the user interface and
+worry about the the interface between your backend and the HTML only when
+you're working on the backend.
+
+The flip side of view-first development is that it takes some getting used to
+if one is accustomed the typical web MVC framework. The first stop when
+figuring out what's going on in a typical web MVC setup is the controller. In
+Lift, your first stop is your HTML file. Everything starts in the HTML, and in
+what it is that you want to present to the user. You don't just think about
+user interactions first, you *build* them first, and let them guide your
+development forward and inform it at every step of the way. Turning a usability
+tested high fidelity mockup into a live page has never been so straightforward.
+
+For our chat app, we're going to focus first on two use cases, formulated as
+user stories:
+
+ - As a chatter, I want to post a message so that others can see it.
+ - As a chatter, I want to see messages from me and others so that I can keep
+   track of the conversation and contribute in context.
+
+To start with, we'll set up a simple chat.html page in our `src/main/webapp`
+directory (where all HTML files go). All we really need in there for now is a
+list of chat messages so far, and a box to put our own chat message into. So,
+here's some base HTML to get us going:
+
+```html:src/main/webapp/index.html
+<!DOCTYPE html>
+<html>
+<head>
+  <title>Chat!</title>
+</head>
+
+<body>
+  <section id="chat">
+    <ol class="messages">
+      <li>Hi!</li>
+      <li>Oh, hey there.</li>
+      <li>How are you?</li>
+      <li>Good, you?</li>
+    </ol>
+    <form class="send-message">
+      <label for="new-message">Post message</label>
+      <input id="new-message" type="text">
+      <input type="submit" value="Post">
+    </form>
+  </section>
+</body>
+</html>
+```
+
+While we're not using it here, it's probably a good idea to start off with
+http://html5boilerplate.com[HTML5 Boilerplate]. Indeed, the default Lift
+templates all start with exactly that footnote:[Ok, so not exactly. IE
+conditional comments need a little additional work in Lift, because Lift is
+smart enough to strip all HTML comments in production mode.].
+
+When it comes to user testing, notice that our view is fully-valid HTML, with
+placeholder data. It is, in effect, a high-fidelity mockup. And now that we've
+got our view sorted out (and, ideally, tested with users), we can start hooking
+up the Lift side.
--- a/docs/getting-started-tutorial/1-view-first-development.adoc
+++ b/docs/getting-started-tutorial/1-view-first-development.adoc
@@ -8,19 +8,20 @@
 If you're developing a user-facing web site or application, one of Lift's
 greatest improvements over existing systems is view-first development.
 View-first development thoroughly separates the process of creating the user
-interface from the process of putting data from the system into it, in a way
-that lets you stay focused on users when you're creating the user interface and
-worry about the the interface between your backend and the HTML only when
+interface from the process of putting data from the system into it. This way,
+you can stay focused on users when you're creating the user interface and
+focus on the interface between your backend and the HTML only when
 you're working on the backend.
 
 The flip side of view-first development is that it takes some getting used to
-if one is accustomed to the typical web MVC framework. The first stop when
+if you are accustomed to the typical web MVC framework. The first stop when
 figuring out what's going on in a typical web MVC setup is the controller. In
-Lift, your first stop is your HTML file. Everything starts in the HTML, and in
-what it is that you want to present to the user. You don't just think about
-user interactions first, you *build* them first, and let them guide your
+Lift, your first stop is your HTML file. Everything starts in the HTML, where
+you decide what it is that you want to present to the user. You don't just think
+about user interactions first, you *build* them first, and let them guide your
 development forward and inform it at every step of the way. Turning a usability
-tested high fidelity mockup into a live page has never been so straightforward.
+tested, high-fidelity mockup into a live page has never been so
+straightforward.
 
 For our chat app, we're going to focus first on two use cases, formulated as
 user stories:
--- a/docs/getting-started-tutorial/2-the-lift-menu-system.adoc
+++ b/docs/getting-started-tutorial/2-the-lift-menu-system.adoc
@@ -24,7 +24,7 @@ directory:
 ```
 
 The string passed to `i` is the name of this menu. We can use that to
-[automatically render links for our menu](#section-12). It gets processed
+link:menu-links[automatically render links for our menu]. It gets processed
 through Lift's internationalization system, but since we've got no
 internationalization set up for now it'll just go through unchanged. The part
 after the `/` specifies where the template will be found—in our case, in the
--- a/docs/getting-started-tutorial/2-the-lift-menu-system.adoc
+++ b/docs/getting-started-tutorial/2-the-lift-menu-system.adoc
@@ -30,4 +30,4 @@ internationalization set up for now it'll just go through unchanged. The part
 after the `/` specifies where the template will be found—in our case, in the
 `chat.html` file directly under `src/main/webapp`.
 
-With that out of the way, we can move on to bringing our HTML to life.
+With that out of the way, we can move on to link:3-adding-snippet-bindings.adoc[bringing our HTML to life].
--- a/docs/getting-started-tutorial/2-the-lift-menu-system.adoc
+++ b/docs/getting-started-tutorial/2-the-lift-menu-system.adoc
@@ -1,0 +1,33 @@
+:idprefix:
+:idseparator: -
+:toc: right
+:toclevels: 2
+
+# The Lift Menu System
+
+Another distinguishing characteristic of Lift is that it is *secure by
+default*. Amongst other things, this means that you can't access a file in your
+`src/main/webapp` directory through your application unless you explicitly
+define that it's meant to be accessed. You define this using Lift's menu
+system, called `SiteMap`.
+
+Hooking up a simple page like this one is easy, and seems redundant; rest
+assured, we'll explore the real power of `SiteMap` as the application becomes
+more complicated. All you have to do for the chat page is add a line to your
+`SiteMap.scala` that names the page and points to the file in the `webapp`
+directory:
+
+```
+...
+    Menu.i("Chat") / "chat"
+...
+```
+
+The string passed to `i` is the name of this menu. We can use that to
+[automatically render links for our menu](#section-12). It gets processed
+through Lift's internationalization system, but since we've got no
+internationalization set up for now it'll just go through unchanged. The part
+after the `/` specifies where the template will be found—in our case, in the
+`chat.html` file directly under `src/main/webapp`.
+
+With that out of the way, we can move on to bringing our HTML to life.
--- a/docs/getting-started-tutorial/2-the-lift-menu-system.adoc
+++ b/docs/getting-started-tutorial/2-the-lift-menu-system.adoc
@@ -6,12 +6,11 @@
 # The Lift Menu System
 
 Another distinguishing characteristic of Lift is that it is *secure by
-default*. Amongst other things, this means that you can't access a file in your
-`src/main/webapp` directory through your application unless you explicitly
-define that it's meant to be accessed. You define this using Lift's menu
-system, called `SiteMap`.
+default*. Amongst other things, this means that if you enable Lift's `SiteMap`
+menu system, you can't access a file in your `src/main/webapp` directory through
+your application unless you explicitly define that it's meant to be accessed.
 
-Hooking up a simple page like this one is easy, and seems redundant; rest
+Hooking up a simple page in `SiteMap` is easy, and seems redundant; rest
 assured, we'll explore the real power of `SiteMap` as the application becomes
 more complicated. All you have to do for the chat page is add a line to your
 `SiteMap.scala` that names the page and points to the file in the `webapp`
--- a/docs/getting-started-tutorial/2-the-lift-menu-system.adoc
+++ b/docs/getting-started-tutorial/2-the-lift-menu-system.adoc
@@ -16,7 +16,7 @@ more complicated. All you have to do for the chat page is add a line to your
 `SiteMap.scala` that names the page and points to the file in the `webapp`
 directory:
 
-```
+```src/scala/bootstrap/liftweb/Boot.scala
 ...
     Menu.i("Chat") / "chat"
 ...
--- a/docs/getting-started-tutorial/3-adding-snippet-bindings.adoc
+++ b/docs/getting-started-tutorial/3-adding-snippet-bindings.adoc
@@ -29,7 +29,7 @@ list of chat messages, and the text input that lets us actually chat. To the
 `ol` that contains the chat messages, we add:
 
 ```
-<ol class="messages" data-lift="Chat.messages"*>
+<ol class="messages" data-lift="Chat.messages">
 ```
 
 And to the input form:
@@ -40,10 +40,10 @@ And to the input form:
 
 These two indicate two methods in a class called `Chat`, which Lift searches
 for in the `code.snippet` package footnote:[This can be changed using
-`Lift.addPackage`. See ...]. We'll write a very basic version that
-just passes through the contents of the list and form unchanged, and then in
-the next section we'll start adding some behavior. In
-`src/main/scala/code/snippet/Chat.scala`, add:
+link:++https://liftweb.net/api/30/api/index.html#net.liftweb.http.LiftRules@addToPackages(what:String):Unit++[`LiftRules.addPackage`.].
+We'll write a very basic version that just passes through the contents of the
+list and form unchanged, and then in the next section we'll start adding some
+behavior. In `src/main/scala/code/snippet/Chat.scala`, add:
 
 ```
 package code
@@ -59,8 +59,8 @@ object Chat {
 
 Note that the methods referred to from the template can either take a
 `NodeSeq` footnote:[What's a `NodeSeq`? Scala uses a `NodeSeq` to represent an
-arbitrary block of XML. It is a *seq*uence of >= 1 *node*s, which can in turn
-have children.] and return a `NodeSeq`, or they can take no parameters and
+arbitrary block of XML. It is a __seq___uence of >= 1 __node___s, which can in
+turn have children.] and return a `NodeSeq`, or they can take no parameters and
 return a `(NodeSeq)=>NodeSeq` function. The `NodeSeq` that is passed in is the
 element that invoked the snippet in the template, minus the `data-lift`
 attribute. The `NodeSeq` that is returned replaces that element completely in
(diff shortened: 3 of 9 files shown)


docs/getting-started-tutorial/6-adding-usernames.adoc
```diff
@@ -0,0 +1,146 @@
...
+    "li" #> messageEntries.map { entry =>
+      ".poster *" #> entry.poster &
+      ".body *" #> entry.body
+    }
+  }
+
+  def sendMessage = {
+    var message = ChatMessage("", "")
+
+    "#new-message" #> SHtml.text(message, { body: String => message = ChatMessage(username.is, body) }) &
```
marcgrue: Some newbie questions that this line gave for me:

- What is this curly bracket thing? - is it an anonymous function definition in Scala? `text` expects a function as it second parameter, so maybe `String => message` is the function type. But `message` isn't a type, hmm. No, wait, aha - `body` is a self type in an anonymous class and the mutable variable `message` is set to a new ChatMessage instance - or? Having it on one line is really confusing. Do we need an anonymous class with a String self type?

- What is this method `is` on `username`? I pasted the code into a Lift project and realised it was simply a getter method on the username session variable to retrieve the value, duh. There's also a shadow method `get` that points to the `is` method. For a newbie tutorial, I would suggest to simply use the `get` method instead. For seasoned Lift users the `is` method is probably obvious but `get` is more generally understood when coming from the outside (as the audience of a tutorial would likely be).

Since `message` is now a ChatMessage, we need to say `message.body` for the first argument to `text`, no?

Could we instead say something like this?

```scala
"#new-message" #> SHtml.text(message.body, newMessage => { message = ChatMessage(username.get, newMessage) }) &
```

It would be great to describe the line in a similar manner as you so well described the previous lines in part 5:

```scala
    "#new-message" #> SHtml.text(message, message = _) &
    "type=submit" #> SHtml.submitButton(() => {
      messageEntries ::= message
    })
```
Once understanding passing NodeSes => NodeSeq functions around, css-selectors become more natural. So I think it is worth explaining examples of those thoroughly.

Shadowfiend: These are great points. A couple of responses:
 - One of the lines I try to walk here is making sure that I don't stray too far into explaining the basics of Scala. I don't think this tutorial is best served by trying to teach *both* Scala *and* Lift. That said, I do dabble very slightly in that in a few places, and I think it's absolutely a good idea to break this one down a little more.
 - I'm not using the `thing => { body }` style because I'm not a fan :)

Pushed some updates that hopefully address some of this, let me know!

Shadowfiend: Hah! I actually completely fumbled this one, updating the bit further down and forgetting this start part was here. That's what I get for trying to do it quickly and while tired. I'll do another pass tomorrow to clean things up :)

Shadowfiend:
Okay, just completely reworked that whole hunk, and started using asciidoc code callouts to boot… I think it's looking a lot better, let me know if you agree @marcgrue !

marcgrue:
Excellent, Antonio! Clear description that is very helpful. I'm very happy :-)

Shadowfiend:
Awesome! Good to hear :) Thanks for looking through the content!

docs/getting-started-tutorial/1-view-first-development.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+# View-first Development
+
+If you're developing a user-facing web site or application, one of Lift's
+greatest improvements over existing systems is view-first development.
+View-first development thoroughly separates the process of creating the user
+interface from the process of putting data from the system into it, in a way
+that lets you stay focused on users when you're creating the user interface and
+worry about the the interface between your backend and the HTML only when
+you're working on the backend.
```
farmdawgnation: This might be easier to parse if it were broken into two sentences right around `in a way`

docs/getting-started-tutorial/1-view-first-development.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+The flip side of view-first development is that it takes some getting used to
+if one is accustomed to the typical web MVC framework. The first stop when
```
farmdawgnation: `if one is accustomed` — I think we should feel free to use the second-person pronouns where applicable. We do so in the first paragraph, and it makes the sentence feel more personal and conversational to me to read `if you're accustomed`. Using `one` has more of an academic, aloof texture.

docs/getting-started-tutorial/1-view-first-development.adoc
```diff
@@ -0,0 +1,71 @@
...
+figuring out what's going on in a typical web MVC setup is the controller. In
+Lift, your first stop is your HTML file. Everything starts in the HTML, and in
+what it is that you want to present to the user. You don't just think about
```
farmdawgnation: `Everything starts in the HTML, and in what it is that you want to present to the user.`

This sentence took me a few reads to comprehend correctly. It might be a good candidate for some tweaking.

docs/getting-started-tutorial/1-view-first-development.adoc
```diff
@@ -0,0 +1,71 @@
...
+user interactions first, you *build* them first, and let them guide your
+development forward and inform it at every step of the way. Turning a usability
+tested high fidelity mockup into a live page has never been so straightforward.
```
farmdawgnation: Comma after `tested` for clarity since you're chaining adjectives.

docs/getting-started-tutorial/1-view-first-development.adoc
```diff
@@ -0,0 +1,71 @@
...
+```
+
+While we're not using it here, it's probably a good idea to start off with
+http://html5boilerplate.com[HTML5 Boilerplate]. Indeed, the default Lift
+templates all start with exactly that footnote:[Ok, so not exactly. IE
+conditional comments need a little additional work in Lift, because Lift is
+smart enough to strip all HTML comments in production mode.].
+
+When it comes to user testing, notice that our view is fully-valid HTML, with
+placeholder data. It is, in effect, a high-fidelity mockup. And now that we've
```
farmdawgnation: Here you use `high-fidelity` with a hyphen. Earlier you use it without. I think we should use the hyphen version consistently.

docs/getting-started-tutorial/2-the-lift-menu-system.adoc
```diff
@@ -0,0 +1,33 @@
...
+:idseparator: -
+:toc: right
+:toclevels: 2
+
+# The Lift Menu System
+
+Another distinguishing characteristic of Lift is that it is *secure by
+default*. Amongst other things, this means that you can't access a file in your
+`src/main/webapp` directory through your application unless you explicitly
+define that it's meant to be accessed. You define this using Lift's menu
```
farmdawgnation: This isn't entirely true. If a `SiteMap` isn't defined at all, then we default to permitting access to all views under the webapp directory that aren't stored in a folder starting with `_`. Unless, of course, that's changed in a recent commit and I missed it. =)

docs/getting-started-tutorial/3-adding-snippet-bindings.adoc
```diff
@@ -0,0 +1,73 @@
...
+
+In most frameworks, a page's data is looked up by a controller, and backend
+code clutters the HTML to produce the correct rendering of the data. This
+process is usually done through what amounts to little more than string
+munging. Lift throws this paradigm away entirely in favor of a much better
+approach based on entities called snippets.
+
+Snippets let you refer to a block of code that is responsible for rendering a
+particular part of your page. You add these references by augmenting your HTML
+with a few completely valid `data-` attributes that get stripped before the
```
farmdawgnation: Would a link to the HTML5 specification or some other explanation for data attributes be useful here?

docs/getting-started-tutorial/3-adding-snippet-bindings.adoc
```diff
@@ -0,0 +1,73 @@
...
+HTML is then sent to the browser. These snippets then take your HTML, fully
+parsed into a valid DOM tree, and transform it, providing true decoupling
+between your business logic and your template, and an extra level of
+security footnote:[We already mentioned that Lift is secure by default, and
+another way that manifests is that the template HTML is turned into a
+first-class XML tree early in the processing cycle, and snippets just transform
+that tree. That means script injection and a variety of other attacks are
+significantly more difficult against a Lift codebase.].
```
farmdawgnation: I'm torn on whether or not the contents of this footnote should be part of the regular paragraph. No further comment than that, just wanted to bring it to your attention.

docs/getting-started-tutorial/3-adding-snippet-bindings.adoc
```diff
@@ -0,0 +1,73 @@
...
+
+
+Let's look at our chat app specifically. We're going to bind two things: the
+list of chat messages, and the text input that lets us actually chat. To the
+`ol` that contains the chat messages, we add:
```
farmdawgnation: The original markup was two pages back. Perhaps we should reproduce it here so the reader is reminded of the overall structure?

docs/getting-started-tutorial/3-adding-snippet-bindings.adoc
```diff
@@ -0,0 +1,73 @@
...
+<ol class="messages" data-lift="Chat.messages">
+```
+
+And to the input form:
+
+```
+<form class="send-message" data-lift="Chat.sendMessage">
+```
+
+These two indicate two methods in a class called `Chat`, which Lift searches
```
farmdawgnation: "These two" -> "These two changes add data attributes that..."

docs/getting-started-tutorial/3-adding-snippet-bindings.adoc
```diff
@@ -0,0 +1,73 @@
...
+```
+package code
+package snippet
+
+import scala.xml._
+
+object Chat {
+  def messages(contents: NodeSeq) = contents
+  def sendMessage(contents: NodeSeq) = contents
+}
```
farmdawgnation: It would be really cool at some point in the future to take a shot at integrating dexy so all of these have to compile or something. But that's not for now. =)

docs/getting-started-tutorial/3-adding-snippet-bindings.adoc
```diff
@@ -0,0 +1,73 @@
...
+```
+
+Note that the methods referred to from the template can either take a
+`NodeSeq` footnote:[What's a `NodeSeq`? Scala uses a `NodeSeq` to represent an
+arbitrary block of XML. It is a __seq___uence of >= 1 __node___s, which can in
+turn have children.] and return a `NodeSeq`, or they can take no parameters and
+return a `(NodeSeq)=>NodeSeq` function. The `NodeSeq` that is passed in is the
+element that invoked the snippet in the template, minus the `data-lift`
+attribute. The `NodeSeq` that is returned replaces that element completely in
+the resulting output.
```
farmdawgnation: It might be worthwhile to include an example of changing the chat page directly by just returning a static `NodeSeq`. As it is we make reference to the fact that this replacement can happen, but don't provide a way for the user to try it out on their own without them taking some additional logical steps.

I think the addition of that will make the utility of CSS Transforms more obvious, too.

docs/getting-started-tutorial/4-css-selector-transforms.adoc
```diff
@@ -0,0 +1,171 @@
+:idprefix:
+:idseparator: -
+:toc: right
+:toclevels: 2
+
+# CSS Selector Transforms
+
+Because Lift operates by transforming HTML trees, it needs an easy way to
```
farmdawgnation: You use `it needs` in this sentence, and then talk about things `we'd be doing` in the following sentence. It might be work considering changing this to `we need`, instead.

docs/getting-started-tutorial/4-css-selector-transforms.adoc
```diff
@@ -0,0 +1,171 @@
...
+specify those transformations. Otherwise we'd be doing a bunch of recursive
+tree searches and munges and it would get ugly and probably slow to boot. To
```
farmdawgnation: It could just be how I'm reading it today, but the end of this sentence feels a bit awkward. Would recommend changing `ugly and probably slow to boot` to `ugly, unpleasant, and time consuming`.

docs/getting-started-tutorial/4-css-selector-transforms.adoc
```diff
@@ -0,0 +1,171 @@
...
+deal with transformations easily, we use a small subset of CSS selectors, with
+a few Lift idiosyncrasies to maximize performance and address some use cases
```
farmdawgnation: Recommend the addition of a comma after the word `idiosyncrasies`. Also recommend potentially using a word other than `idiosyncrasies` as it has a negative connotation. Perhaps `Lift-specific additions`?

docs/getting-started-tutorial/4-css-selector-transforms.adoc
```diff
@@ -0,0 +1,171 @@
...
+that are particularly useful when transforming trees.
+
+We'll leave forms for the next section, as forms always come with a catalog of
+related functionality, and focus on binding the list of chat messages in this
+section. We'll also add a new one before every page load, so that we can see
```
farmdawgnation: Recommend replacing `new one` with `new message`

docs/getting-started-tutorial/4-css-selector-transforms.adoc
```diff
@@ -0,0 +1,171 @@
...
+
+```
+...
+
+import net.liftweb.util.Helpers._
+
+...
+  def messages = {
+    "li *" #> messageEntries
+  }
```
farmdawgnation: Examples in this form are frequently confusing to me when reading getting started guides. Can we present the entire `Chat` object with imports in each of these since it's fairly self contains?

docs/getting-started-tutorial/4-css-selector-transforms.adoc
```diff
@@ -0,0 +1,171 @@
...
+``String``s. When there's a list on the right side of a transformation, Lift
+repeats the matched element or elements once for each entry in the list, and
+binds the contents of each element in turn.
+
+Let's start up Lift and see what's going on. In your terminal, enter the
+directory of the chat app and start up the application:
+
+```
+$ sbt
+> container:start
```
farmdawgnation: We'll need to update this to `jetty:start` once we upgrade to the new version of the web plugin

docs/getting-started-tutorial/4-css-selector-transforms.adoc
```diff
@@ -0,0 +1,171 @@
...
+  }
+...
+```
+
+Notice that we combine the two CSS selector transforms here by using `&`. You
+can chain together as many CSS selector transforms as you want this way, as long
+as they don't modify the same parts of the same element. We'll deal with that
+limitation link:13-who-knows[a little later] footnote:[This is because CSS
+selector transforms are optimized for speed, and pass through the nodes a
+single time to make all of the transformations happen.].
```
farmdawgnation: It's probably also worth adding a footnote here regarding indentation rules, since any Scala developer passing through will likely wonder why we don't represent this as:

```scala
ClearClearable &
  "li *" #> messageEntries
```

docs/getting-started-tutorial/5-basic-forms.adoc
```diff
@@ -0,0 +1,106 @@
...
+are associated with a callback function that runs when the field is submitted
+with the form. On the client, the field name is always unique to this page
+load, and this unique field name is a cryptographically secure random value
+that is associated to the callback function you specify on the server
+footnote:[What about using your own field names, you may ask? You can always do
+that. To access a submitted field with a given name, you can use
+`S.param("field name")`. You'll get back a `Box` that will be `Full` and
+contain the value you submitted if it was submitted. The `Box` will be `Empty`
+if the field wasn't submitted in this request. However, be very careful about
+using this method, since it exposes you to CSRF attacks.]. This makes Lift
```
farmdawgnation: Should there be a link to the Lift Cookbook entry to [Plain Old Form Processing](http://chimera.labs.oreilly.com/books/1234000000030/ch03.html#_solution_24) here?

docs/getting-started-tutorial/6-adding-usernames.adoc
```diff
@@ -0,0 +1,153 @@
...
+        <span class="poster">Antonio</span> <span class="body">Hi!</span>
+      </li>
+      <li class="clearable">
+        <span class="poster">David</span> <span class="body">Oh, hey there.</span>
+      </li>
+      <li class="clearable">
+        <span class="poster">Antonio</span> <span class="body">How are you?</span>
+      </li>
+      <li class="clearable">
+        <span class="poster">Antonio</span> <span class="body">Good, you?</span>
```
farmdawgnation: You talking to yourself there? 😛

docs/getting-started-tutorial/7-using-actors-for-chat.adoc
```diff
@@ -0,0 +1,159 @@
...
+}
+```
+
+This provides a very basic actor that can receive a new message and add it to
+its internal list. We've moved the `ChatMessage` class from the `Chat` snippet
+to this file. Typically, messages to actors are case classes. This is because
+they're easy to pattern match (as you can see, message handling is done via
+pattern matching footnote:[Strictly speaking, `messageHandler` is a
+`PartialFunction`. This means that it can match any subset of objects that it
+wants to.]) and because they're generally immutable, so there's no chance of
```
farmdawgnation: Technically, there's `little` change of someone else trying to modify the message as we're processing it. But, eh, technicalities. heh.

docs/getting-started-tutorial/7-using-actors-for-chat.adoc
```diff
@@ -0,0 +1,159 @@
...
+very handy utility function, `Box.asA[T]`; it attempts to convert its parameter
+to the type `T`, and, if it succeeds, provides a `Full` `Box` with the
+converted value of the appropriate type. If it fails, it provides an `Empty`
+`Box` instead.
+
+To deal with the fact that the `Box` may be `Full` or `Empty`, we use `flatten`
+on the `Box`. We do this because the type of `messageEntries` is now a
+`Box[List[ChatMessage]]`, meaning a box that *contains* a list of chat
+messages. `flatten` will give us the plain list of messages if the `Box` is
+`Full`, and an empty list if it's `Empty`, which is perfect.
```
farmdawgnation: Are you sure about that?

```scala
@ Box(List("abc")).flatten
cmd2.sc:1: Cannot prove that String <:< net.liftweb.common.Box[B].
val res2 = Box(List("abc")).flatten
                            ^
Compilation Failed
```

docs/getting-started-tutorial/7-using-actors-for-chat.adoc
```diff
@@ -0,0 +1,159 @@
...
+
+It's worth mentioning that it seems like we *know* we'll be getting a
+`List[ChatMessage]` from the actor. However, the compiler *doesn't*, and that
+means it can't guarantee to us that future changes won't render that assumption
+false. Using `Box.asA` ensures that, if someone changes the `ChatActor` later
+to reply with something else, our snippet won't blow up in the user's face—it
+will just not display the existing messages. The intrepid reader can then go
+and fix the issue.
```
farmdawgnation: It's probably worth noting that someone should probably have unit tests on that actor to ensure people don't just go changing its contract all willy nilly 😉

docs/getting-started-tutorial/8-customizable-usernames.adoc
```diff
@@ -0,0 +1,69 @@
...
+  <section id="chat">
+    <header>
+      Posting as:
+      <input data-lift="Chat.nameField" class="poster-name">
+    </header>
+...
+```
+
+The ideal way for this to work would be for you to be able to change the value
+of the field, and have it save. We can do exactly that using Lift's `ajaxText`
```
farmdawgnation: Not sure about the comma after `field`. Might also be worthwhile to make mention that the save occurs on blur?

docs/getting-started-tutorial/9-comet-actors.adoc
```diff
@@ -0,0 +1,20 @@
...
+we're already using to track chat messages, only `CometActor`s have a couple
+more tricks up their sleeve. Most notably for our purposes: they can re-render
+themselves in response to stuff going on inside the server, and send the
+updated version to the client. This finally gives us a way to update Jane's
+chat message list when Jill posts a message, without Jane having to do
+anything.
+
+Our first move will be to change how exactly we handle binding chat messages.
+First, we'll do a quick conversion that puts everything in a `CometActor`, but
+doesn't add any additional functionality. Instead of calling
```
farmdawgnation: AMG SUSPENSE

Shadowfiend:
🎵 I have not forgotten youuuuu 🎵

But I'll be trying to wrap up #1876 before circling back here.

farmdawgnation:
This PR is...
http://i0.kym-cdn.com/entries/icons/mobile/000/003/619/ForeverAlone.jpg

docs/getting-started-tutorial/1-view-first-development.adoc
Shadowfiend: Excellent point!

docs/getting-started-tutorial/1-view-first-development.adoc
Shadowfiend: I prefer no contractions, since this is written, but I'm down to go second-person to make it less academic.

docs/getting-started-tutorial/1-view-first-development.adoc
Shadowfiend: Rephrased, great point.

docs/getting-started-tutorial/1-view-first-development.adoc
Shadowfiend: :bow:

docs/getting-started-tutorial/1-view-first-development.adoc
Shadowfiend: Your eye for detail brings a tear to my eye ;)

docs/getting-started-tutorial/2-the-lift-menu-system.adoc
Shadowfiend: ! Nice catch. Rephrased a bit to take that into account.

docs/getting-started-tutorial/3-adding-snippet-bindings.adoc
Shadowfiend: Because this is a tutorial, I decided I wanted to keep the top-level content super-focused on implementation rather than reasoning.

docs/getting-started-tutorial/3-adding-snippet-bindings.adoc
Shadowfiend: Yes… But let me punt on that a bit. I'm wanting to do a little preprocessing pass on these guys that handles building some of that for you.

docs/getting-started-tutorial/3-adding-snippet-bindings.adoc
Shadowfiend: 100% part of my plan, yeah.

docs/getting-started-tutorial/3-adding-snippet-bindings.adoc
Shadowfiend: That's a good point… There are definitely some good spots here to suggest exploration, I think. e.g., “try changing `contents` to <p>I am the walrus!</p>” and see what happens!

docs/getting-started-tutorial/4-css-selector-transforms.adoc
Shadowfiend: Yeah I went too conversational here, trying to go for a rhythm that shows all the bad things snowballing haha.

docs/getting-started-tutorial/4-css-selector-transforms.adoc
Shadowfiend: Totally rephrased that bit, it was definitely a little rambly.

docs/getting-started-tutorial/4-css-selector-transforms.adoc
Shadowfiend: I actually want the compiled output here to produce the whole thing with highlighted changes… I have some more thoughts here. Bit off a little more than I could chew the first time I tried to do this, but I think there are some small things we can do to get to that point quickly.

docs/getting-started-tutorial/4-css-selector-transforms.adoc
Shadowfiend: I'm going to punt on this… Not sure that note really belongs in a beginner tutorial. I've also been rethinking the general principle, although I like the look of same-indentation. Mostly because if you consider auto-indentation tools your friend, they can actually catch the common error of forgetting an `&`... Anyway, I don't have a strong thought there for now, but I think we can get away with not mentioning it in the tutorial.

docs/getting-started-tutorial/5-basic-forms.adoc
Shadowfiend: I guess we'll need to figure out the Lift Cookbook situation… So will punt here.

docs/getting-started-tutorial/6-adding-usernames.adoc
Shadowfiend: See, here I am now by myself, talking to myself. That's Chaos Theory

docs/getting-started-tutorial/7-using-actors-for-chat.adoc
Shadowfiend: I am… Not, all of a sudden, lol. Reworking to use `openOr`, which is a better pattern in this case anyway I think.

docs/getting-started-tutorial/7-using-actors-for-chat.adoc
Shadowfiend: True, though even so you can imagine someone changing the test + actor and missing the snippet because the compiler can't complain about it.

docs/getting-started-tutorial/9-comet-actors.adoc
Shadowfiend: Lawl yeah this guy is… Incomplete :tada:

Shadowfiend:
Ok, pushed a variety of changes motivated by the above comments :tada:

I would say we can probably merge with these changes and then iterate on additional tweaks, but if you have time to give it a second read-through or at least look at the tweaks, @farmdawgnation, I won't complain :)

PR #1852, opened 2017-05-13
Rushed Responder: Always suspend comet requests before resuming them.

### LINKED ISSUE #1848: Race-condition in asynchronous response handling
**[Mailing List](https://groups.google.com/forum/#!forum/liftweb) thread:** https://groups.google.com/forum/#!topic/liftweb/-Fve3lAyO7M

My client has an application running Lift 3.0.1 and Jetty 9.2.2. They have a problem with requests that "hang" for a very long time before they are completed with empty responses.

I think this could be due to a race-condition in the method handleGenericContinuation of the class LiftServlet. This method (directly and indirectly) calls the suspend and resume methods of the ServletAsyncProvider interface but does not guarantee that suspend is called before resume. As far as I can see, the order of the calls can depend on the scheduling of threads. And indeed, after having inserted debug traces and attaching a debugger to the server I have the impression that the resume method is sometimes called before the suspend method, leading to the hanging requests we're seeing. Please see mailing list thread for more information.

Minor correction to what I wrote in the mailing list thread: I claimed that the NullPointerExceptions we are seeing are thrown from the complete method of the underlying Servlet 3.0 API. This is wrong -- it is thrown when invoking the getResponse method of the AsyncContext interface (in the resume method of Servlet30AsyncProvider, of course). Sorry for the confusion.

Shadowfiend:
When a container supports request suspension, we were first scheduling a
function to resume the request on a different thread and then suspending the
request. While the function was scheduled in the future, the fact that it was
scheduled before the suspend had definitively occurred meant there were thread
schedulings where the resume could occur before the suspend had executed,
leading to exceptions.

We now suspend the request *before* scheduling the resume. We schedule the
resume in a finally block to most closely track current behaviors when the
thread scheduling was correct, in which a resume would occur whether or not the
suspend threw an exception. I'm not certain about the full shape of that edge
case, so for now we take the least-changed path to dealing with it while still
addressing the core reported problem.

Should fix #1848.

@waern any chance you could give this a test drive?

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftServlet.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftServlet.scala
@@ -114,14 +114,17 @@ class LiftServlet extends Loggable {
     }
 
     if (reqOrg.request.suspendResumeSupport_?) {
-      runFunction(liftResponse => {
-        // do the actual write on a separate thread
-        Schedule.schedule(() => {
-          reqOrg.request.resume(reqOrg, liftResponse)
-        }, 0.seconds)
-      })
+      try {
+        reqOrg.request.suspend(cometTimeout)
+      } finally {
+        runFunction(liftResponse => {
+          // do the actual write on a separate thread
+          Schedule.schedule(() => {
+            reqOrg.request.resume(reqOrg, liftResponse)
+          }, 0.seconds)
+        })
+      }
 
-      reqOrg.request.suspend(cometTimeout)
       false
     } else {
       val future = new LAFuture[LiftResponse]
@@ -880,7 +883,7 @@ class LiftServlet extends Loggable {
 
     val jsCommands =
       new JsCommands(JsCmds.Run(jsUpdateTime) :: jsUpdateStuff)
-    
+
     // If we need to, ensure we capture JS from this request's render version.
     // The comet actor will already have handled the comet's version.
     S.request.flatMap(req => extractRenderVersion(req.path.partPath)) match {
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftServlet.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftServlet.scala
@@ -117,6 +117,10 @@ class LiftServlet extends Loggable {
       try {
         reqOrg.request.suspend(cometTimeout)
       } finally {
+        // FIXME This should be removed from the finally block in the next major
+        // FIXME Lift version; it is currently here to ensure any code that
+        // FIXME depends on a response handler running even if requests suspension
+        // FIXME fails continues to work as-is during the 3.x series.
         runFunction(liftResponse => {
           // do the actual write on a separate thread
           Schedule.schedule(() => {
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftServlet.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftServlet.scala
@@ -118,9 +118,9 @@ class LiftServlet extends Loggable {
         reqOrg.request.suspend(cometTimeout)
       } finally {
         // FIXME This should be removed from the finally block in the next major
-        // FIXME Lift version; it is currently here to ensure any code that
-        // FIXME depends on a response handler running even if requests suspension
-        // FIXME fails continues to work as-is during the 3.x series.
+        // FIXME Lift version; it is currently here to ensure code continues to
+        // FIXME work as-is during the 3.x series even if it depends on response
+        // FIXME handlers running when request suspension fails.
         runFunction(liftResponse => {
           // do the actual write on a separate thread
           Schedule.schedule(() => {


waern:
Thanks for the fix! If suspend throws, I think the right thing to do is to log the exception and fail with an error response to the client. Otherwise the error is simply delayed until resume is called.

farmdawgnation:
> Otherwise the error is simply delayed until resume is called.

The delay there seems negligible though, right?

waern:
The delay is not negligible, it depends on how long it takes to process the request. But the request shouldn't be processed at all if suspend fails since there is no point in computing an answer that cannot be sent back to the client.

Shadowfiend:
My goal here was basically minimal change to behavior, i.e. ensuring that if, before, we ran the resume every time, we would do the same here, while still fixing the resume-before-suspend issue.

It's important to note that the scheduled function does not *only* run the resume, it also runs the code that computes the result. There is a lot of stuff going on in the background, and there are a lot of complex interactions that lead to a request being resumed---most critically, part of these interactions happen in client code. If someone has code out there *today* that expects that computation to run irrespective of whether there's an issue handling the request, I don't want to change that behavior with this targeted fix.

It's also worth noting that the decision here is between throwing one exception and two exceptions. If it were between throwing *no* exceptions and throwing one exception, that would change the equation significantly for me.

So that's my reasoning for keeping this fix as-is. Happy to discuss those points, though.

waern:
Sure, it's always possible that someone has code that relies on buggy behaviour. In this case the combination of suspend failing and someone relying on still being able to compute something in that case seems quite unlikely IMHO. But if you want to keep the current behaviour, can you log the exception?

Shadowfiend:
Keep in mind that from the browser perspective, "suspend failing" really just looks like "the request failed". So imagine that you are sending telemetry back over a comet connection, and suspend fails occasionally---the request doesn't work, but the telemetry still gets reported, and that is valuable.

The current code should bubble the `suspend` exception up for handling by any logging that's configured at the jetty level. The `resume` exception would be logged by the configured actor logger, as it occurs during a `Schedule` block, IIRC. We can also do additional logging of the `suspend` exception at the exception site, but leaving that to higher-up code *seems* like the right decision. Once again, happy to discuss---I am no logging ninja :)

waern:
I missed the fact that the failing suspend would actually be logged with the current fix, so I'm fine with it (except I think it would be better to failing fast -- but let's just agree to disagree there).

Minor detail: are you sure the resume exception would be logged for all async provider classes? I remember looking at the code and not seeing the exception caught without logging and when resume failes with `Jetty7AsyncProvider`. And I had to use a debugger to figure out what happened.

Shadowfiend:
I'm not 100% positive, no; I'll check when I get a chance.

Shadowfiend:
(Of note: I agree it would be better to fail fast in general---but backwards compatibility is generally quite important in Lift-land. I'll probably drop a FIXME on it to change this behavior in the next major version.)

waern:
Great! Is it possible for you to upload a jar file somewhere that I can test?

Shadowfiend:
[lift-webkit-jar.zip](https://github.com/lift/framework/files/1006332/lift-webkit-jar.zip)

Above, a ZIP with two lift-webkit JARs, one for 2.11 and one for 2.12.

Also added the FIXME.

waern:
Thanks a lot!

I tried the JAR for 2.11 (we're using Scala 2.11.8) but I got some compilation errors. We're on Lift 3.0.1 -- do you think the JAR is compatible with that?

waern:
Actually, I tried upgrading to Lift 3.1.0-M2 and I still get the compilation errors with the bug fixed webkit jar. They look like this:

```
[error] /tmp/nix-build-ksdental.drv-0/evsdental/src/main/scala/code/service/AdminService.scala:98: overloaded method constructor FileInputStream with alternatives:
[error]   (x$1: java.io.FileDescriptor)java.io.FileInputStream <and>
[error]   (x$1: java.io.File)java.io.FileInputStream <and>
[error]   (x$1: String)java.io.FileInputStream
[error]  cannot be applied to (java.nio.file.Path)
[error]         createThumbnail(new FileInputStream(f.localFile))
[error]                         ^
[error] /tmp/nix-build-ksdental.drv-0/evsdental/src/main/scala/code/service/AdminService.scala:101: ambiguous reference to overloaded definition,
[error] both constructor SerialBlob in class SerialBlob of type (x$1: java.sql.Blob)javax.sql.rowset.serial.SerialBlob
[error] and  constructor SerialBlob in class SerialBlob of type (x$1: Array[Byte])javax.sql.rowset.serial.SerialBlob
[error] match argument types (Nothing)
[error]     val blobObj = thumbnail.map(new SerialBlob(_)).headOption
[error]                                 ^
[error] /tmp/nix-build-ksdental.drv-0/evsdental/src/main/scala/code/rest/AdminRest.scala:301: type mismatch;
[error]  found   : java.nio.file.Path
[error]  required: java.io.File
[error]         AdminService.saveHipaaFile(fileName = f.fileName, sourceFile = f.localFile)
[error]                                                                          ^
[error] /tmp/nix-build-ksdental.drv-0/evsdental/src/main/scala/code/rest/InsuranceManagementRest.scala:214: type mismatch;
[error]  found   : java.nio.file.Path
[error]  required: java.io.File
[error]         FeeService.uploadFeePrice(f.fileName, f.localFile, feeGuideId)
[error]                                                 ^
[error] /tmp/nix-build-ksdental.drv-0/evsdental/src/main/scala/code/rest/PatientRest.scala:323: type mismatch;
[error]  found   : java.nio.file.Path
[error]  required: java.io.File
[error]           sourceFile = f.localFile, attributes = imageAtrributes(req))
[error]                          ^
[warn] two warnings found
[error] 5 errors found
[error] (evsdental/compile:compileIncremental) Compilation failed
[error] Total time: 211 s, completed May 17, 2017 10:59:19 AM
```

Shadowfiend:
Derp. There were some [changes to use newer `java.nio` classes](https://github.com/lift/framework/pull/1837/files#r116997218), but I missed that one of these resulted in a public API change (`localFile` is now a `java.nio.file.Path`). We'll work to get that fixed up by the 3.1 final release. In the meantime, do you mind updating references to `localFile` to instead call `localFile.toFile` to test this?

As a side note, `FileParamHolder` has a `.fileStream` method that will return an `InputStream` for you directly.

Shadowfiend:
(Alternatively I've [opened a PR that fixes this](https://github.com/lift/framework/pull/1854), so I can try to spin a JAR build for you tonight with those changes included.)

waern:
Thanks, I fixed the compilation errors and have deployed the fix our test system. We will run it for a few days and see how it behaves.

Shadowfiend:
Great! If things look good in your tests, we'll merge this in for the June 15th RC build so it's rolled into 3.1.0 :)

waern:
We have run with the fix now for about a week in our test system without seeing any hanging request. So it seems to be working!

Shadowfiend:
Good deal! @farmdawgnation perhaps you'd be willing to merge? I'm always loath to merge my own PRs, but will probably do it in a day or two if you don't have time :)

web/webkit/src/main/scala/net/liftweb/http/LiftServlet.scala
```diff
@@ -114,14 +114,21 @@ class LiftServlet extends Loggable {
...
-        Schedule.schedule(() => {
-          reqOrg.request.resume(reqOrg, liftResponse)
-        }, 0.seconds)
-      })
+      try {
+        reqOrg.request.suspend(cometTimeout)
+      } finally {
+        // FIXME This should be removed from the finally block in the next major
+        // FIXME Lift version; it is currently here to ensure any code that
+        // FIXME depends on a response handler running even if requests suspension
```
farmdawgnation: I think you mean `request suspension` here instead of `requests suspension`

web/webkit/src/main/scala/net/liftweb/http/LiftServlet.scala
```diff
@@ -114,14 +114,21 @@ class LiftServlet extends Loggable {
...
+        // FIXME fails continues to work as-is during the 3.x series.
+        runFunction(liftResponse => {
```
farmdawgnation: Should we log a warning in the even we catch an exception here advising people of the impending change of behavior? Actually logging a `debug` or `warn` in general seems like a reasonable idea in the even suspension isn't working, but I also don't know how frequently that actually occurs.

web/webkit/src/main/scala/net/liftweb/http/LiftServlet.scala
farmdawgnation: Also, do we have a ticket tracking this clean up? Should we link to such a ticket from this comment?

Shadowfiend: We certainly can create one. Probably would be worth making it a catchall to check all FIXMEs.

web/webkit/src/main/scala/net/liftweb/http/LiftServlet.scala
Shadowfiend: I don't think it's worth the log noise, personally. The exception will already show up in the logs unless someone's configured exceptions to be suppressed, so that's its own thing. I think clearly noting the change in behavior in a hypothetical future release is a better approach.

Shadowfiend:
Tweaked the FIXME wording and added #1858 to track checking into FIXMEs during the next major release cycle.

farmdawgnation:
ship it

Issue #1777, opened 2016-03-11
Implement caching for FullTypeHints


Shadowfiend:
`FullTypeHints` in lift-json does a classloader lookup whenever
it tries to instantiate a type, to look the class up based on its
fully-qualified name. Unfortunately, classloader `loadClass` is
a blocking, synchronized action, which can cause serious thread
contention in high load scenarios.

We added a `ConcurrentHashMap` cache of FQNs to `Class`
objects at Elemica to a drop-in replacement class and saw
significant improvements. We should move this into Lift proper.

farmdawgnation:
:+1:

farmdawgnation:
Oh I could have used a reaction if I wasn't mobile.

Shadowfiend:
`FastFullTypeHints` implementation for reference, will try to get this folded in soon:

```scala
case class FastFullTypeHints(hints: List[Class[_]]) extends TypeHints {

  private val hintsToClass: concurrent.Map[String, Class[_]] =
    new ConcurrentHashMap[String, Class[_]]().asScala ++= hints.map(clazz => hintFor(clazz) -> clazz)

  override def hintFor(clazz: Class[_]): String = clazz.getName

  override def classFor(hint: String): Option[Class[_]] =
    hintsToClass.get(hint).orElse {
      val clazz = Thread.currentThread.getContextClassLoader.loadClass(hint)
      hintsToClass.putIfAbsent(hint, clazz).orElse(Some(clazz))
    }
}
```

PR #1853, opened 2017-05-13
Speedy GonTypeHints: Add class name->Class caching to FullTypeHints.

Shadowfiend:
FullTypeHints in lift-json does a classloader lookup whenever it tries to
instantiate a type, to look the class up based on its fully-qualified name.
Unfortunately, classloader loadClass is a blocking, synchronized action, which
can cause serious thread contention in high load scenarios.

Implements #1777.

### DIFF
--- a/core/json/src/main/scala/net/liftweb/json/Formats.scala
+++ b/core/json/src/main/scala/net/liftweb/json/Formats.scala
@@ -18,6 +18,10 @@ package net.liftweb
 package json
 
 import java.util.{Date, TimeZone}
+import java.util.concurrent.ConcurrentHashMap
+
+import scala.collection.concurrent.{Map=>ConcurrentScalaMap}
+import scala.collection.JavaConverters._
 
 /** Formats to use when converting JSON.
  * Formats are usually configured by using an implicit parameter:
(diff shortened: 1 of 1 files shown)


core/json/src/main/scala/net/liftweb/json/Formats.scala
```diff
@@ -230,9 +234,16 @@ case class ShortTypeHints(hints: List[Class[_]]) extends TypeHints {
 /** Use full class name as a type hint.
  */
 case class FullTypeHints(hints: List[Class[_]]) extends TypeHints {
+  private val hintsToClass: ConcurrentScalaMap[String, Class[_]] =
+    new ConcurrentHashMap[String, Class[_]]().asScala ++= hints.map(clazz => hintFor(clazz) -> clazz)
```
farmdawgnation: What's the performance impact of using the scala wrapper around CHM? It might be nothing, but my anecdotal impression is that using that is met with a bit of skepticism by other Scalans. I could be wrong, though, in which case I would be glad to know it.

Shadowfiend: It should be minimal for what we're using it for. Certainly we've seen a giant improvement from this change at Elemica. Doesn't mean that there isn't some performance left on the table, but compared to the big win we got from adding the caching it's negligible.

Shadowfiend:
Re: json benchmark, which one are you referring to? The ones we've used for recent JSON PRs test JSON serialization and deserialization, not extraction, which is where this comes into play. Is there one you're aware of that does test that?

farmdawgnation:
Oh sorry. I thought I had code somewhere that we could just plug in and do this. I was wrong.

farmdawgnation:
If there are no remarks on this by the end of the day I think we merge it.

PR #1854, opened 2017-05-17
Improve API compatibility of OnDiskFileParamHolder.

Shadowfiend:
When #1837 shifted to using java.nio.file.* classes, we made `localFile` on
`OnDiskFileParamHolder` a `java.nio.file.Path` instead of a `java.io.File`.
Additionally, though we provided an `apply` overload that took a `File` and
converted to a `Path`, we didn't provide a *constructor* overload that did the
same.

We unify the constructor and apply overload, and we change `localFile` to
`localPath`. We also add a `localFile` accessor that converts the `localPath`
to a `java.io.File` for compatibility.


Shadowfiend:
Whoops, accidental bad commit; fixed it! <_<

Issue #1856, opened 2017-05-22
Explore additional Box combinators and hierarchical improvements


Shadowfiend:
A few possibilities:
 - Add `collect` to `Box`. This can be useful for `PartialFunction`-based combined filter+map operations on boxes as they are on collections.
 - Add some sort of `mapFailure` that behaves similarly to `map` but only runs with a `Failure`. This would allow running a function only when the box is a `Failure`, without having to do a `match` that has to deal with the other three box states.
 - Restructure the `Box` hierarchy a little to allow a type that indicates “only `Empty` or `Full`” and another that indicates “only `Failure` or `Full`”, for example.

For the last one, an untested proposal I posted to gitter a while back:

```scala
sealed trait Box[+T]
sealed trait PresenceBox[+T] extends Box[T]
sealed trait TryBox[+T] extends Box[T]
sealed abstract class EmptyBox extends Box[Nothing] // Empty or Failure
case class Failure(msg: String, exception: PresenceBox[Throwable], chain: PresenceBox[Failure]) extends EmptyBox with TryBox[Nothing]
case object Empty extends EmptyBox
case class Full[T](item: T) extends TryBox[T] with PresenceBox[T]
```

Bhashit:
Can I take this up?

Shadowfiend:
Sure, happy to see any attempts. My example above forgot to have `Empty` extend `PresenceBox`.

Bhashit:
Alright.

Also, what kind of logging helpers are you looking for? Any approximate examples?

Shadowfiend:
Oh, logging helper wise I actually just want to port a bunch of stuff we have at Elemica into Lift. So I've more or less already figured this out, if that's okay… This was definitely a brain dump for myself, sorry.

Shadowfiend:
Another thought on the hierarchy:

```scala
sealed trait ParamFailableBox[+T, A] extends Box[T]

case class ParamFailure[A](..., param: A) extends ParamFailableBox[Nothing, A]
case class Full[T](item: T) extends TryBox[T] with PresenceBox[T] with ParamFailableBox[T, Nothing]
```

Not 100% sure that bit'll work, but it allows you to declare that something expects only a `Full` and a `ParamFailure` with a particular type to be produced.

Bhashit:
Looks good. That will be helpful.

[...]

PR #1857, opened 2017-05-25
Loggable Boxes: Add BoxLogging helpers.

Shadowfiend:
Direct lift from the commit message:

`BoxLogging` provides helpers for logging boxes and then continuing to operate on
the box, alloing logging to become a fluent part of the box interaction. It
provides methods that can allow for logging empties and failures, or just
failures. It also allows logging these to different log levels, and allows for
different adapters to be applied.

For example, one could write a snippet:

```scala
  class MySnippet(id: String) extends LoggableBoxLogging {
    def render = {
      "#content" #> {
        Content.getById(id).flatMap { content =>
          content.text.asHtml // assume this returns Box[NodeSeq]
        }.infoLogEmptyBox(s"Failed to retrieve content HTML for $id")
    }
  }
```

When this snippet gets rendered, if the provided content lookup fails, or if
the conversion to `NodeSeq` fails, there will be an INFO level log that will
capture this failure with the provided message prefix.

`BoxLogging` provides handling for arbitrary-depth failure chains, as well as
providing logging for failure params, thus ensuring all of the information in a
failure will be logged. It also correctly passes any captured Throwable to the
logger methods, provided that the adapter handles them correctly.

Two adapters are provided out of the box: `LoggableBoxLogging`, which can be
mixed in as above and is based on Lift's `Loggable`, and `SLF4JBoxLogging`,
which is based on the raw SLF4J interface, and requires a protected `logger`
`var` with an SLF4J logger to be available.

Lastly, for simplicity, you can `import net.liftweb.common.BoxLogging._` to have
the logging helpers available without mixing anything else into your classes.

We've been using this at @elemica with great success for a while now, so it's time
to contribute it on back to Lift. This version even grew some extra capabilities, with
the possibility of doing this logging at various log levels.

See #1856 .

### DIFF
--- a/core/common/src/test/scala/net/liftweb/common/HListSpec.scala
+++ b/core/common/src/test/scala/net/liftweb/common/HListSpec.scala
@@ -81,6 +81,8 @@ object HListSpec extends Specification {
 
           success
         }
+
+        case Right(_) => failure
         case Left(_) => failure
       }
     }
@@ -90,7 +92,7 @@ object HListSpec extends Specification {
       import HLists._
 
       val res = for {
-        a :+: one :+: lst :+: _ <- 
+        a :+: one :+: lst :+: _ <-
         (Full("a") ?~ "Yak" :&: Full(1) :&: Full(List(1,2,3))) ?~! "Dude"
       } yield a.length * one * lst.foldLeft(1)(_ * _)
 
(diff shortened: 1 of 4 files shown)


farmdawgnation:
This looks familiar. =)

core/common/src/test/scala/net/liftweb/common/HListSpec.scala
```diff
@@ -81,6 +81,8 @@ object HListSpec extends Specification {
 
           success
         }
+
+        case Right(_) => failure
```
farmdawgnation: Not sure I track the relevance of this change.

Shadowfiend: This just fixed an exhaustive match warning during lift-common spec compilation.

PR #1859, opened 2017-05-25
Failure Collection: Fix LAFuture.collect/collectAll when sub-futures fail.

### LINKED ISSUE #1850: LAFuture.collect does not handle failures
Current `LAFuture.collect`, when passed a future that will fail, will produce
a future that will never complete, as success or failure.

Need to update [the `foreach` invocation](https://github.com/lift/framework/blob/master/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala#L356) to use `onComplete` instead, and
fail the collected future if there is a failure in a contained future.

We should also update `collectAll` accordingly.

Shadowfiend:
Before, this would cause the overall future to hang forever. We now fail the
overall future if any contained future fails.

We also refactor the way that collect and collectAll are written to go through
a common path, which is also available for third-party use, and we add some
extra specs around collect and collectAll behavior.

Fixes #1850 .

### DIFF
--- a/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
+++ b/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
@@ -176,7 +176,7 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler, context: Box[LAFutur
   /**
    * Java-friendly alias for completed_?.
    */
-  def isCompleted: Boolean = complete_?
+  def isCompleted: Boolean = completed_?
   /**
    * Has the future completed?
    */
@@ -389,10 +389,10 @@ object LAFuture {
               gotCnt += 1
               onFutureSucceeded(value, result, accumulator, index)
 
-              if (gotCnt >= len && ! result.complete_?) {
+              if (gotCnt >= len && ! result.completed_?) {
                 onAllFuturesCompleted(result, accumulator)
 
-                if (! result.complete_?) {
+                if (! result.completed_?) {
                   result.fail(Failure("collect invoker did not complete result"))
                 }
               }
@@ -403,10 +403,10 @@ object LAFuture {
               gotCnt += 1
               onFutureFailed(failureBox, result, accumulator, index)
 
-              if (gotCnt >= len && ! result.complete_?) {
+              if (gotCnt >= len && ! result.completed_?) {
                 onAllFuturesCompleted(result, accumulator)
 
-                if (! result.complete_?) {
+                if (! result.completed_?) {
                   result.fail(Failure("collect invoker did not complete result"))
                 }
               }
--- a/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
+++ b/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
@@ -155,19 +155,35 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler, context: Box[LAFutur
   }
 
   /**
+   * Java-friendly alias for satisfied_?.
+   */
+  def isSatisfied: Boolean = satisfied_?
+
+  /**
    * Has the future been satisfied
    */
-  def isSatisfied: Boolean = synchronized {satisfied}
+  def satisfied_? = synchronized {satisfied}
+  /**
+   * Java-friendly alias for aborted_?.
+   */
+  def isAborted: Boolean = aborted_?
 
   /**
    * Has the future been aborted
    */
-  def isAborted: Boolean = synchronized {aborted}
+  def aborted_? = synchronized {satisfied}
 
   /**
-   * Has the future been satisfied or
+   * Java-friendly alias for completed_?.
    */
-  def isCompleted: Boolean = synchronized { ! (isSatisfied || isAborted) }
+  def isCompleted: Boolean = complete_?
+  /**
+   * Has the future completed?
+   */
+  def completed_? : Boolean = synchronized(satisfied || aborted)
+
+  @deprecated("Please use completed_? instead.", "3.1.0")
+  def complete_? : Boolean = completed_?
 
   /**
    * Abort the future.  It can never be satified
@@ -247,11 +263,6 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler, context: Box[LAFutur
       }
     }
   }
-
-  /**
-   * Has the future completed?
-   */
-  def complete_? : Boolean = synchronized(satisfied || aborted)
 }
 
 /**
--- a/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
+++ b/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
@@ -17,6 +17,8 @@
 package net.liftweb
 package actor
 
+import scala.collection.mutable.ArrayBuffer
+
 import common._
 
 
@@ -161,6 +163,11 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler, context: Box[LAFutur
    * Has the future been aborted
    */
   def isAborted: Boolean = synchronized {aborted}
+
+  /**
+   * Has the future been satisfied or
+   */
+  def isCompleted: Boolean = synchronized { ! (isSatisfied || isAborted) }
 
   /**
    * Abort the future.  It can never be satified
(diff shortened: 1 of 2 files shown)


farmdawgnation:
Didn't have time to review this morning, but it's on y list for my next OSS stint.

core/actor/src/test/scala/net/liftweb/actor/LAFutureSpec.scala
```diff
@@ -61,30 +61,81 @@ class LAFutureSpec extends Specification {
...
-      collectResult.isSatisfied shouldEqual true
-      collectResult.get(timeout) shouldEqual Nil
+      "report a failed LAFuture as a failure for the overall future" in {
+        val one: LAFuture[Int] = new LAFuture
+        val two: LAFuture[Int] = LAFuture(() => 5)
+
+        one.fail(Failure("boom boom boom!"))
+
+        val collectResult = LAFuture.collect(one, two)
+        collectResult.get(timeout) shouldEqual Failure("boom boom boom!")
```
farmdawgnation: Is it weird that the return type here is `List[_]` on success and then a `Failure` when things go sideways? Am I missing something here?

core/actor/src/test/scala/net/liftweb/actor/LAFutureSpec.scala
```diff
@@ -61,30 +61,81 @@ class LAFutureSpec extends Specification {
...
-      val collectResult = LAFuture.collectAll(Nil : _*)
-      collectResult.isSatisfied shouldEqual true
-      collectResult.get(timeout) shouldEqual Nil
+    "when collecting Boxed results with collectAll" in {
+      "collectAll collects an EmptyBox immediately" in {
+        val one: LAFuture[Box[Int]] = LAFuture(() => { Failure("whoops"): Box[Int] })
+        val two: LAFuture[Box[Int]] = LAFuture(() => { Thread.sleep(10000); Full(1) })
+
+        val collectResult = LAFuture.collectAll(one, two)
+        collectResult.get(5000) shouldEqual Failure("whoops")
```
farmdawgnation: Should we also assert here that `two` does actually execute?

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
```diff
@@ -333,32 +351,64 @@ object LAFuture {
...
+   * completed. For the failure handler, the value is the `Box` of the failure.
    */
-  def collect[T](future: LAFuture[T]*): LAFuture[List[T]] = {
-    val result = new LAFuture[List[T]]
-    if (future.isEmpty) {
-      result.satisfy(Nil)
+  def collect[T, A](
+    valueSucceeded: (T, LAFuture[A], ArrayBuffer[Box[T]], Int)=>Unit,
+    valueFailed: (Box[Nothing], LAFuture[A], ArrayBuffer[Box[T]], Int)=>Unit,
+    allValuesCompleted: (LAFuture[A], ArrayBuffer[Box[T]])=>Unit,
```
farmdawgnation: These function signatures are complex enough to warrant a type alias.

Also I'm a bit confused by the fact that`valueFailed` has `Box[Nothing]` instead of just `Failure`. We'll always get a `Failure` in that position, right?

farmdawgnation: Should these be named in event handler style? e.g. `onValueSucceeded` instead of `valueSucceeded`?

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
```diff
@@ -368,6 +418,21 @@ object LAFuture {
...
+   * Collect all the future values into the aggregate future
+   * The returned future will be satisfied when all the
+   * collected futures are satisfied
+   */
+  def collect[T](future: LAFuture[T]*): LAFuture[List[T]] = {
+    collect[T, List[T]](
+      { (value, result, values, index) => values.insert(index, Full(value)) },
+      { (valueBox, result, values, index) => result.fail(valueBox) },
+      { (result, values) => result.satisfy(values.toList.flatten) },
+      future: _*
```
farmdawgnation: For clarity, I think we should either...

1. Use the argument names in the method invocation (so, `valueSucceeded = { (value result...`
2. Declare these callbacks as their own `def`s within `collect[T]` and pass them into the collect call by name

I like option (2) the best, but could be convinced either is better. =)

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
```diff
@@ -376,40 +441,21 @@ object LAFuture {
...
+          case Full(realValue) =>
+            values.insert(index, Full(Full(realValue)))
+          case other: EmptyBox =>
+            result.satisfy(other)
+        }
+      },
+      { (valueBox, result, values, index) => result.fail(valueBox) },
+      { (result: LAFuture[Box[List[T]]], values: ArrayBuffer[Box[Box[T]]]) =>
+        result.satisfy(Full(values.toList.flatten.flatten))
+      },
```
farmdawgnation: Ditto here on somehow indicating how each of these functions are mapped to the underlying collect implementation without having to refer back to the positional order of things in that function.

Though here I feel more strongly that these functions should be their own `def`s nested inside `collectAll`

core/actor/src/test/scala/net/liftweb/actor/LAFutureSpec.scala
Shadowfiend: It is not only weird, it's incorrect heh. The test above should be `shouldEqual Full(Nil)`.... Not entirely sure why that test is passing O.o

core/actor/src/test/scala/net/liftweb/actor/LAFutureSpec.scala
Shadowfiend: We could… I don't know that that's particularly core to the behavior, but I'm not opposed to having that check to make sure that we don't break that behavior in the future unless we want to.

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
Shadowfiend: Hmmm… I feel like a type alias would just obfuscate the types involved further. Will think about it a bit and see if I can come up with a better way to handle it.

farmdawgnation: So my thinking here is that when I read this I want to know "What types does this take?"

```scala
valueSucceeded: ValueSucceededCallback,
valueFailed: ValueFailedCallback,
allValuesCompleted: AllValuesCompleteCallback
```

makes more sense to me. Sure, figuring out what each of those are requires a bit of effort on my part. But, it's a lot easier to grok in passing when I don't really care about the underlying function params.

Shadowfiend: That looks to me like using the type to convey the intended use, which is what the argument names do. It doesn't help me at all to know that `valueSucceeded` is a `ValueSucceededCallback`, except perhaps to know that it's a callback (IMO better achievable with your suggestion below to rename this in event-handling style, which I'm not opposed to).

farmdawgnation: Well, we're possibly in some uncanny valley where the type and the intended use are tightly coupled. Unlike a `String` this exact sequence of (arugments, result) only have one possible connotation in this file and enabling a human to identify that sequence correctly across multiple occurrences has value. "Do function X and Y take the same argument at some point?" is a question we frequently have to answer as developers, and function signatures in parameter lists makes answering that more challenging.

That said, I'm not interested in bike shedding until we're both blue in the face. Renaming the parameters to more clearly indicate the event-handling nature would satisfy me enough to rubber stamp this PR. =)

core/actor/src/test/scala/net/liftweb/actor/LAFutureSpec.scala
Shadowfiend: Turns out I misread this. `.get` returns a `Box[List]` for `LAFuture.collect`. If any of the futures fails, that fails the collecting future immediately with that future's failure. This behavior is shared with `collectAll`.

The return type of `.get` in case of success is `Full[List]`, not `List`.

core/actor/src/test/scala/net/liftweb/actor/LAFutureSpec.scala
Shadowfiend: Per my response above, we could but we shouldn't. `collectAll` doesn't guarantee the *execution* of any of the futures---in fact, it doesn't affect their execution at all. All it does is gather their results. So if there were a way to furnish a future that would never get scheduled, `collectAll` would simply never complete.

Of course for Lift futures the act of creating the future schedules it, but basically that's orthogonal to this code. So I don't think we should test it here.

Shadowfiend:
@farmdawgnation I believe addressed the remaining concerns if you're willing to clear :)

PR #1849, opened 2017-05-11
Rewrite xml literals to interpolator

allanrenucci:
I am writing an xml interpolator and a scalafix rewrite.

The purpose of this PR is to make sure the tests still pass without choking my cpu.

PR #1860, opened 2017-05-26
Rewrite xml literals to xml interpolator

allanrenucci:
Second attempt at #1849


Shadowfiend:
Looks like a successful run?

PR #1861, opened 2017-06-11
Add special handling for UnavailableException

### LINKED ISSUE #1843: HTTPProvider.bootlift catch UnavailableException but should not
**[Mailing List](https://groups.google.com/forum/#!forum/liftweb) thread:** https://groups.google.com/forum/#!topic/liftweb/BHYjE5C-kIk

The HTTPProvider.bootLift method is catching javax.servlet.UnavailableException, which it should not because the exception is here to tell the application server that the servlet failed to boot.

In Lift 2.6.x, it was even worse because the method was catching ALL exceptions, even Errors. 

*How to reproduce*
Throws a javax.servlet.UnavailableException in Lift boot() method. The exception is caught during boot and the log displays a big warning message:
```
------------------------------------------------------------------
------------------------------------------------------------------
------------------------------------------------------------------
------------------------------------------------------------------
******** Failed to Boot! Your application may not run properly
[here the exception message]
------------------------------------------------------------------
------------------------------------------------------------------
------------------------------------------------------------------
------------------------------------------------------------------
```

*Use case* 
I'm doing a lot of initialization and resource availability checking at bootstrap, and it was natural to put them in Lift boot() method. The UnavailableException was a conventionnal way to tell the web-app server that something went horribly wrong during init and the servlet is  unavailable. Jetty (for ex) will then respond with "error 503: service unavailable" when a query is done to that app. 

*Workaround*
For now, I added a ServletContextListener where I'm doing the init and throw UnavailableException, but it  splits the initialization logic between it and Lift boot() and make the code harder to follow.

farmdawgnation:
https://groups.google.com/forum/#!topic/liftweb/BHYjE5C-kIk

This code adds special handling for javax.servelet.UnavailableException.
This exception is an idiomatic way to signal a full abort to the Java
Application server running the application.

From now on, we will log and re-throw this exception if we see it.

Fixes #1843

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/provider/HTTPProvider.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/provider/HTTPProvider.scala
@@ -95,7 +95,7 @@ trait HTTPProvider {
           logger.error("------------------------------------------------------------------")
           logger.error("------------------------------------------------------------------")
           logger.error("------------------------------------------------------------------")
-          logger.error("********** Failed to Boot! An unavailable exception was thrown an all futher boot activities are aborted", unavailableException);
+          logger.error("********** Failed to Boot! An UnavailableException was thrown and all futher boot activities are aborted", unavailableException);
           logger.error("------------------------------------------------------------------")
           logger.error("------------------------------------------------------------------")
           logger.error("------------------------------------------------------------------")
--- a/web/webkit/src/main/scala/net/liftweb/http/provider/HTTPProvider.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/provider/HTTPProvider.scala
@@ -14,9 +14,9 @@
  * limitations under the License.
  */
 
-package net.liftweb 
-package http 
-package provider 
+package net.liftweb
+package http
+package provider
 
 import net.liftweb.common._
 import net.liftweb.util._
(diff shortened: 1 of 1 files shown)


farmdawgnation:
@fanf This PR resolves the issue you logged regarding `javax.servlet.UnavailableException`. Can you please take it out for a spin and verify that this fixes your issue completely before we ship it?

web/webkit/src/main/scala/net/liftweb/http/provider/HTTPProvider.scala
```diff
@@ -87,12 +87,29 @@ trait HTTPProvider {
...
       } catch {
+        // The UnavailableException is the idiomatic way to tell a Java application container that
+        // the boot process has gone horribly, horribly wrong. That _must_ bubble to the application
+        // container that is invoking the app. See https://github.com/lift/framework/issues/1843
+        case unavailableException: javax.servlet.UnavailableException =>
+          logger.error("------------------------------------------------------------------")
+          logger.error("------------------------------------------------------------------")
+          logger.error("------------------------------------------------------------------")
+          logger.error("------------------------------------------------------------------")
+          logger.error("********** Failed to Boot! An unavailable exception was thrown an all futher boot activities are aborted", unavailableException);
```
Shadowfiend: I might say “An UnavailableException was thrown”; either way, it should be “and” rather than “an”.

web/webkit/src/main/scala/net/liftweb/http/provider/HTTPProvider.scala
```diff
@@ -87,12 +87,29 @@ trait HTTPProvider {
...
         b.boot
       } catch {
+        // The UnavailableException is the idiomatic way to tell a Java application container that
+        // the boot process has gone horribly, horribly wrong. That _must_ bubble to the application
+        // container that is invoking the app. See https://github.com/lift/framework/issues/1843
+        case unavailableException: javax.servlet.UnavailableException =>
+          logger.error("------------------------------------------------------------------")
+          logger.error("------------------------------------------------------------------")
+          logger.error("------------------------------------------------------------------")
+          logger.error("------------------------------------------------------------------")
```
Shadowfiend: Do we really need the four lines of dashes? Feels like one should be more than enough……

web/webkit/src/main/scala/net/liftweb/http/provider/HTTPProvider.scala
farmdawgnation: Aye.

web/webkit/src/main/scala/net/liftweb/http/provider/HTTPProvider.scala
farmdawgnation: I'm not super opinionated on this, personally, but I'm inclined to leave it as is for now and reconsider when we next touch this code.

fanf:
This is ok with me, the behavior is what is needed. I'm not fan of the big error message, but I understand the need for consistancy.

farmdawgnation:
Yeah, I think we'll revisit that error message at some point in the future.

farmdawgnation:
Cool, then I think this is good to merge.

Shadowfiend:
Let us, as they say in the lands of Sekulampa, do it.

PR #1862, opened 2017-06-11
Use a custom implicitNotFound error for CSS transforms

### LINKED ISSUE #930: Use @scala.annotation.implicitNotFound in CssSelectors etc.
I noticed a tweet from Josh Seureth that there's an annotation (at least in 2.8.1) that lets you customize the error message when the compiler cannot find an implicit of a given type. It's possible this would make a better developer experience where implicits are used, such as in the css selectors (anywhere else?). Indrajit agreed and told me to file a ticket, so here it is. :)

farmdawgnation:
N/A

This adds a custom error message that provides a bit more direction when we can't find an implicit converter for a CSS transform. Tested it out by creating an error in a spec... it looks like this!

```
[error] /Users/matt/Projects/framework/core/util/src/test/scala/net/liftweb/util/CssSelectorSpec.scala:502: The value you're attempting to use on the right hand side of the #> operator can't be automatically converted to a (NodeSeq)=>NodeSeq function. Consider binding the individual members of the class you're trying to use, or define an implicit CanBind[T] that tells Lift how to convert this to a (NodeSeq)=>NodeSeq.
[error]       ".foo" #> Bacon("boo!")
[error]              ^
[error] one error found
```

Fixes #930

### DIFF
--- a/core/util/src/main/scala/net/liftweb/util/CssSel.scala
+++ b/core/util/src/main/scala/net/liftweb/util/CssSel.scala
@@ -4,6 +4,7 @@ package util
 import common._
 import xml._
 import collection.mutable.ListBuffer
+import scala.annotation.implicitNotFound
 
 /**
  * Created with IntelliJ IDEA.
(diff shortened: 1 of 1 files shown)


Shadowfiend:
This fixes #930, right?

farmdawgnation:
Yes, updated the PR description to mention that.

I can haz merge? :D

Shadowfiend:
Ya okay :shipit:

PR #1863, opened 2017-06-11
Resolve some issues with swallowing exceptions in lift-json

### LINKED ISSUE #1774: `MappingException` doesn't include cause when triggered by another Exception
In certain cases, lift-json can throw a `MappingException` due to
another exception being caught. However, we don't add the original
exception as the `cause` of the `MappingException`, so debugging
can be more complicated because the underlying failure can't be
seen.

Problematic place in the code is in [`Extraction.scala:204`](https://github.com/lift/lift/blob/master/framework/lift-base/lift-json/src/main/scala/net/liftweb/json/Extraction.scala#L204), though
there may be others.

See [this ML thread](https://groups.google.com/d/msg/liftweb/omdcCBwGb2c/5UaKyuBLCAAJ) for the original report.

farmdawgnation:
This should resolve some issues with swallowing exceptions in lift-json extraction/parsing. I did an audit of everywhere `fail` is used and everywhere exceptions are caught in general and I think I got them all.

Closes #1774

### DIFF
--- a/core/json/src/main/scala/net/liftweb/json/Extraction.scala
+++ b/core/json/src/main/scala/net/liftweb/json/Extraction.scala
@@ -272,6 +272,9 @@ object Extraction {
                      args.mkString(",") + "\narg types=" + args.map(a => if (a != null)
                        a.asInstanceOf[AnyRef].getClass.getName else "null").mkString(",") +
                      "\nconstructor=" + jconstructor, matchedException)
+
+              case unmatchedException =>
+                throw unmatchedException
             }
         }
       }
--- a/core/json/src/main/scala/net/liftweb/json/Extraction.scala
+++ b/core/json/src/main/scala/net/liftweb/json/Extraction.scala
@@ -265,11 +265,14 @@ object Extraction {
           val instance = jconstructor.newInstance(args.map(_.asInstanceOf[AnyRef]).toArray: _*)
           setFields(instance.asInstanceOf[AnyRef], json, jconstructor)
         } catch {
-          case e @ (_:IllegalArgumentException | _:InstantiationException) =>
-            fail("Parsed JSON values do not match with class constructor\nargs=" +
-                 args.mkString(",") + "\narg types=" + args.map(a => if (a != null)
-                   a.asInstanceOf[AnyRef].getClass.getName else "null").mkString(",") +
-                 "\nconstructor=" + jconstructor)
+          case exception: Exception =>
+            exception match {
+              case matchedException @ (_:IllegalArgumentException | _:InstantiationException) =>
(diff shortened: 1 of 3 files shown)


core/json/src/main/scala/net/liftweb/json/Extraction.scala
```diff
@@ -265,11 +265,14 @@ object Extraction {
           val instance = jconstructor.newInstance(args.map(_.asInstanceOf[AnyRef]).toArray: _*)
           setFields(instance.asInstanceOf[AnyRef], json, jconstructor)
         } catch {
-          case e @ (_:IllegalArgumentException | _:InstantiationException) =>
-            fail("Parsed JSON values do not match with class constructor\nargs=" +
-                 args.mkString(",") + "\narg types=" + args.map(a => if (a != null)
-                   a.asInstanceOf[AnyRef].getClass.getName else "null").mkString(",") +
-                 "\nconstructor=" + jconstructor)
+          case exception: Exception =>
```
Shadowfiend: Hmm… This went from a targeted catch to catching all exceptions and swallowing any exception that isn't an `IllegalArgumentException` or an `InstantiationException`, right?

Should we add a spec or two to confirm this is doing what we think it should be? For this particular one I think it's when the class that you're deserializing is either throwing an instantiation exception or an illegal argument exception.

farmdawgnation: Ah, you're right, so I should have a

```
case e =>
  throw e
```

For context, this is the result of a peculiarity about how "or" ing cases works. Without the outer match statement the type of the variable is `Throwable` even though they're both `Exception`.

farmdawgnation: With regard to the test, I'm not sure, but I can look at doing that.

farmdawgnation:
@Shadowfiend I think I addressed your concern in the most recent commit. Unfortunately, the method this code is in is nested inside another def (`extract0`) which is a mega def that does many things. I think that disqualifies this for a targeted unit test automatically. But I also am not sure what I could do to trigger an exception except trying to `extract` a `java.lang.Object` and asserting that the expected MappingException appears at the calling code.

Is that worthwhile here?

Shadowfiend:
I think you'd need to create a constructor that throws an exception (to trigger the instantiation exception). I think it's worth ensuring that the `MappingException` appears *with the appropriate cause*, since that's the particular thing we're trying to fix here and I think it's worth guarding against regressions in that behavior.

In a perfect world we'd check the re-throw as well, but what are ya gonna do.

farmdawgnation:
Updated the PR and even added a little bit extra to improve the top-level error message that we surface. Let me know if this is good enough to ship.

PR #1864, opened 2017-06-17
Add collect and mapFailure methods.

Bhashit:
Issue #1856 

Adds `mapFailure` method. Creating an early PR for review to see if it's going in the right direction.

### DIFF
--- a/core/common/src/main/scala/net/liftweb/common/Box.scala
+++ b/core/common/src/main/scala/net/liftweb/common/Box.scala
@@ -793,9 +793,9 @@ sealed abstract class Box[+A] extends Product with Serializable{
     * Transforms this box using the `transformFn`. If `transformFn` is defined for this box,
     * returns the result of applying `transformFn` to it. Otherwise, returns this box unchanged.
     *
-    * If you want to change the content of a `Full` box, using [[map]] or [[collect]] might be better
+    * If you want to change the content of a `Full` box, using `[[map]]` or `[[collect]]` might be better
     * suited to that purpose. If you want to convert an `Empty`, `Failure` or a `ParamFailure` into a
-    * `Full` box, you should use [[flip]].
+    * `Full` box, you should use `[[flip]]`.
     *
     * @example {{{
     *
--- a/core/common/src/main/scala/net/liftweb/common/Box.scala
+++ b/core/common/src/main/scala/net/liftweb/common/Box.scala
@@ -809,7 +809,27 @@ sealed abstract class Box[+A] extends Product with Serializable{
     collect(pf)
   }
 
-  final def collectFailure[B](pf: PartialFunction[Failure, Failure]): Box[Failure] = this match {
+  /**
+    * If this box is a `Failure` and if `pf` is defined for this `Failure` instance, returns
+    * a `Full` box containing the result of applying `pf` to this `Failure`, otherwise, returns
+    * `Empty`.
+    *
+    * @example {{{
+    *  // Returns Full("alternative") because the partial function covers the case.
+    *  Failure("error") collectFailure { case Failure("error", Empty, Empty) => "alternative" }
+    *
+    *  // Returns Empty because the partial function doesn't cover the case.
+    *  Failure("another-error") collectFailure { case Failure("error", Empty, Empty) => "alternative" }
+    *
+    *  // Returns Empty for an Empty box
+    *  Empty collectFailure { case _ => Failure("error") }
+    *
+    *  // Returns Empty for a Full box
+    *  Full(1) collectFailure { case _ => Failure("error") }
+    *
+    *  }}}
+    */
+  final def collectFailure[B](pf: PartialFunction[Failure, B]): Box[B] = this match {
     case f: Failure if pf.isDefinedAt(f) => Full(pf(f))
     case _ => Empty
   }
--- a/core/common/src/main/scala/net/liftweb/common/Box.scala
+++ b/core/common/src/main/scala/net/liftweb/common/Box.scala
@@ -757,7 +757,7 @@ sealed abstract class Box[+A] extends Product with Serializable{
 
 
   /**
-   * If the `Box` is `Fuvall`, apply the transform function `f` on the value `v`;
+   * If the `Box` is `Full`, apply the transform function `f` on the value `v`;
    * otherwise, just return the value untransformed.
    *
    * The transform function is expected to be a function that will take the
--- a/core/common/src/main/scala/net/liftweb/common/Box.scala
+++ b/core/common/src/main/scala/net/liftweb/common/Box.scala
@@ -167,8 +167,9 @@ sealed trait BoxTrait {
    * @return A `Full` containing the transformed value if
    *         `pf.isDefinedAt(value)` and `Empty` otherwise.
    */
-  def apply[InType, OutType](pf: PartialFunction[InType, OutType])(value: InType): Box[OutType] =
-  if (pf.isDefinedAt(value)) Full(pf(value)) else Empty
+  def apply[InType, OutType](pf: PartialFunction[InType, OutType])(value: InType): Box[OutType] = {
+    apply(value)(pf)
+  }
 
   /**
    * Apply the specified `PartialFunction` to the specified `value` and return
@@ -180,8 +181,9 @@ sealed trait BoxTrait {
    * @return A `Full` containing the transformed value if
    *         `pf.isDefinedAt(value)` and `Empty` otherwise.
    */
-  def apply[InType, OutType](value: InType)(pf: PartialFunction[InType, OutType]): Box[OutType] =
-  if (pf.isDefinedAt(value)) Full(pf(value)) else Empty
+  def apply[InType, OutType](value: InType)(pf: PartialFunction[InType, OutType]): Box[OutType] = {
+    pf.andThen(Full.apply).applyOrElse(value, (_:  InType) => Empty)
+  }
 
   /**
    * This implicit transformation allows one to use a `Box` as an `Iterable` of
@@ -755,7 +757,7 @@ sealed abstract class Box[+A] extends Product with Serializable{
 
 
   /**
-   * If the `Box` is `Full`, apply the transform function `f` on the value `v`;
+   * If the `Box` is `Fuvall`, apply the transform function `f` on the value `v`;
    * otherwise, just return the value untransformed.
    *
    * The transform function is expected to be a function that will take the
@@ -797,7 +799,9 @@ sealed abstract class Box[+A] extends Product with Serializable{
    * If the partial function is defined at the current Box's value, apply the
    * partial function.
    */
-  final def collect[B](pf: PartialFunction[A, B]): Box[B] = filter(pf.isDefinedAt).map(pf)
+  final def collect[B](pf: PartialFunction[A, B]): Box[B] = flatMap { value =>
+    Box(value)(pf)
+  }
 
   /**
    * An alias for `collect`.
@@ -831,6 +835,16 @@ sealed abstract class Box[+A] extends Product with Serializable{
     */
   final def collectFailure[B](pf: PartialFunction[Failure, B]): Box[B] = this match {
     case f: Failure if pf.isDefinedAt(f) => Full(pf(f))
+    case _ => Empty
+  }
+
+  /**
+    * Returns a `Full` box containing the results of applying `flipFn` to this box if it is a `Failure`,
+    * `ParamFailure` or `Empty`. Returns `Empty` if this box is `Full`. In other words, it "flips" the
+    * full/empty status of this Box.
+    */
+  def flip[B](flipFn: EmptyBox => B): Box[B] = this match {
+    case e: EmptyBox => Full(flipFn(e))
     case _ => Empty
   }
 }
--- a/core/common/src/main/scala/net/liftweb/common/Box.scala
+++ b/core/common/src/main/scala/net/liftweb/common/Box.scala
@@ -814,37 +814,30 @@ sealed abstract class Box[+A] extends Product with Serializable{
   }
 
   /**
-    * If this box is a `Failure` and if `pf` is defined for this `Failure` instance, returns
-    * a `Full` box containing the result of applying `pf` to this `Failure`, otherwise, returns
-    * `Empty`.
+    * Returns a `Full` box containing the results of applying `flipFn` to this box if it is a `Failure`,
+    * `ParamFailure` or `Empty`, and `flipFn` is defined for it. Returns `Empty` if this box is `Full`.
+    * In other words, it "flips" the full/empty status of this Box.
     *
     * @example {{{
     *  // Returns Full("alternative") because the partial function covers the case.
-    *  Failure("error") collectFailure { case Failure("error", Empty, Empty) => "alternative" }
+    *  Failure("error") flip { case Failure("error", Empty, Empty) => "alternative" }
     *
     *  // Returns Empty because the partial function doesn't cover the case.
-    *  Failure("another-error") collectFailure { case Failure("error", Empty, Empty) => "alternative" }
+    *  Failure("another-error") flip { case Failure("error", Empty, Empty) => "alternative" }
     *
-    *  // Returns Empty for an Empty box
-    *  Empty collectFailure { case _ => Failure("error") }
+    *  // Returns Full("alternative") for an Empty box since `partialFn` is defined for Empty
+    *  Empty flip { case Empty => "alternative" }
+    *
+    *  // Returns Empty because the partial function is not defined for Empty
+    *  Empty flip { case Failure("error", Empty, Empty) => "alternative" }
     *
     *  // Returns Empty for a Full box
-    *  Full(1) collectFailure { case _ => Failure("error") }
+    *  Full(1) flip { case _ => Failure("error") }
     *
     *  }}}
     */
-  final def collectFailure[B](pf: PartialFunction[Failure, B]): Box[B] = this match {
-    case f: Failure if pf.isDefinedAt(f) => Full(pf(f))
-    case _ => Empty
-  }
-
-  /**
-    * Returns a `Full` box containing the results of applying `flipFn` to this box if it is a `Failure`,
-    * `ParamFailure` or `Empty`. Returns `Empty` if this box is `Full`. In other words, it "flips" the
-    * full/empty status of this Box.
-    */
-  def flip[B](flipFn: EmptyBox => B): Box[B] = this match {
-    case e: EmptyBox => Full(flipFn(e))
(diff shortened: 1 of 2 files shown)


Bhashit:
Will be adding tests for this as well. Just wanted to get this in early for a review.

Bhashit:
The suggestion by `codacy` is about the fact that `mapFailure` includes the word `Failure`, which is also the name of one of the classes that has that method.

farmdawgnation:
🙄 Codacy.

Feel free to ignore it's output if it's about something silly. I feel like there are many, many times where I need to disregard one of its suggestions because it's not actionable.

Shadowfiend:
Looks like you removed the only `collect` method, not a duplicate? `mapFailure` looks sweet.

Bhashit:
I removed the duplicate `collect` method I had added. We already have an implementation at https://github.com/lift/framework/blob/master/core/common/src/main/scala/net/liftweb/common/Box.scala#L776

Shadowfiend:
> About the collect method

< rubs eyes > lol, guess I'd missed that! Need to make a note to myself to change that to use `applyOrElse`.

Shadowfiend:
As I'm working on lift-formality, it occurred to me it could also be particularly useful to add a `collectFailure`, since pattern matching is used heavily when dealing with `ParamFailure`. Thoughts?

core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
```diff
@@ -152,14 +152,17 @@ class BoxSpec extends Specification with ScalaCheck with BoxGenerator {
         Full(Empty).flatten must_== Empty
       }
     }
+    "define a 'mapFailure' method returning itself." in {
+      Full(1).mapFailure(identity) must_==  Full(1)
```
Shadowfiend: Even if `mapFailure` didn't map just over a failure, this would return `Full(1)`. How about a parameter that returns some standard `Failure`?

core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
```diff
@@ -298,14 +301,17 @@ class BoxSpec extends Specification with ScalaCheck with BoxGenerator {
     "define a 'flatten' method returning Empty" in {
       Empty.flatten must beEmpty
     }
+    "define a 'mapFailure' method returning itself." in {
+      Empty.mapFailure(identity) must beEmpty
```
Shadowfiend: Ditto re: above.

core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
```diff
@@ -358,7 +364,11 @@ class BoxSpec extends Specification with ScalaCheck with BoxGenerator {
       Failure("error", Empty, Empty) flatMap {x: String => Full(x.toString)} must_== Failure("error", Empty, Empty)
       Failure("error", Empty, Empty).flatten must_== Failure("error", Empty, Empty)
     }
-    "return a itself when asked for its status with the operator ?~" in {
+    "define a 'mapFailure' method that transforms it into another Failure instance." in {
+      val exception = new Exception("transformed")
+      Failure("error", Empty, Empty) mapFailure { _ => Failure("new-error", Full(exception), Empty) } must_=== Failure("new-error", Full(exception), Empty)
```
Shadowfiend: Any particular need for `===` here?

Bhashit:
> Need to make a note to myself to change that to use applyOrElse

I'll do that.

> it could also be particularly useful to add a collectFailure, since pattern matching is used heavily when dealing with ParamFailure. Thoughts?

I'll push an implementation for review.

core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
Bhashit: Yeah, I did that get the niggling feeling when I wrote this. You are right. That could be confusing. I'll change it to something more appropriate.

core/common/src/main/scala/net/liftweb/common/Box.scala
```diff
@@ -797,11 +797,7 @@ sealed abstract class Box[+A] extends Product with Serializable{
    * If the partial function is defined at the current Box's value, apply the
    * partial function.
    */
-  final def collect[B](pf: PartialFunction[A, B]): Box[B] = {
-    flatMap(value =>
-    if (pf.isDefinedAt(value)) Full(pf(value))
-    else Empty)
-  }
+  final def collect[B](pf: PartialFunction[A, B]): Box[B] = filter(pf.isDefinedAt).map(pf)
```
Bhashit: @Shadowfiend thought this was more readable than

```
final def collect[B](pf: PartialFunction[A, B]): Box[B] = flatMap { value =>
  pf.andThen(Full.apply).applyOrElse(value, _ => Empty)
}

```

Bhashit:
@Shadowfiend just pushed an initial version of `collectFailure`. If it looks good, I'll add some comments and tests, and if it doesn't, I'll make the necessary changes.

core/common/src/main/scala/net/liftweb/common/Box.scala
```diff
@@ -808,6 +808,11 @@ sealed abstract class Box[+A] extends Product with Serializable{
   final def collectFirst[B](pf: PartialFunction[A, B]): Box[B] = {
     collect(pf)
   }
+
+  final def collectFailure[B](pf: PartialFunction[Failure, Failure]): Box[Failure] = this match {
+    case f: Failure if pf.isDefinedAt(f) => Full(pf(f))
+    case _ => Empty
+  }
```
Shadowfiend: It feels like this should be a `PartialFunction[Failure, B]` and result in a `Box[B]`, no?

Bhashit: That was my first instinct as well. Almost sent a comment asking about these two signatures. I think I'll go with that one then. Now that you said it, it makes more sense.

Bhashit:
@Shadowfiend I think this might be ready for review. I.e. now it contains all the methods that I set out to add along with their respective tests.

Shadowfiend:
@Bhashit `flip` can be a separate PR if you'd like. Up to you.

core/common/src/main/scala/net/liftweb/common/Box.scala
Shadowfiend: I do think this version will result in a double-application of the `PartialFunction` guards/pattern matches (a problem that the previous version in the repo also shared):

```scala
@ def slap(in: String) = println(s"Uh-oh ${in}"); false 
defined function slap
@ val booyan: PartialFunction[String, Int] = { case "boom" if slap("jam") => 5; case _ => 8 } 
booyan: PartialFunction[String, Int] = <function1>

@ Full("boom").filter(booyan.isDefinedAt).map(booyan) 
Uh-oh jam
Uh-oh jam
res13: Box[Int] = Full(8)
@ booyan.andThen(Full.apply).applyOrElse("boom", { _: String => Empty }) 
Uh-oh jam
res12: Box[Int] = Full(8)
```

So while I agree this is slightly more readable, I think you nailed it with the original alternative formulation.

Bhashit:
@Shadowfiend added the `flip` method. And corrected the `collect` implementation. It turns out that `Box` has a couple of `apply` implementations that take a `PartialFunction` and do what we need. So, used those. And used `applyOrElse` in those existing methods. The remaining codacy complaint is about the method `mapFailure` having "Failure" in its name.

core/common/src/main/scala/net/liftweb/common/Box.scala
```diff
@@ -731,7 +757,7 @@ sealed abstract class Box[+A] extends Product with Serializable{
 
 
   /**
-   * If the `Box` is `Full`, apply the transform function `f` on the value `v`;
+   * If the `Box` is `Fuvall`, apply the transform function `f` on the value `v`;
```
Shadowfiend: Whoops!

core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
```diff
@@ -152,14 +152,32 @@ class BoxSpec extends Specification with ScalaCheck with BoxGenerator {
         Full(Empty).flatten must_== Empty
       }
     }
+    "define a 'mapFailure' method returning itself." in {
+      val failure = Failure("new-error", Empty, Empty)
+      Full(1).mapFailure(_ => failure) must_==  Full(1)
+    }
```
Shadowfiend: This looks like it's a duplicate of one below?

Shadowfiend: Oh wait jk I just remembered how this spec is structured heh.

core/common/src/test/scala/net/liftweb/common/BoxSpec.scala
```diff
@@ -358,7 +392,29 @@ class BoxSpec extends Specification with ScalaCheck with BoxGenerator {
...
+      val exception = new Exception("transformed")
+      Failure("error", Empty, Empty) mapFailure { _ => Failure("new-error", Full(exception), Empty) } must_== Failure("new-error", Full(exception), Empty)
+    }
+    "define a 'collect' method returning itself" in {
+      Failure("error", Empty, Empty) collect { case _ => "Some Value" } must_== Failure("error", Empty, Empty)
+    }
+    "define a 'collectFailure' method that takes a PartialFunction to transform this Failure into something else" in {
+      "If the partial-function is defined for this Failure, returns a full box containing the result of applying the partial function to it" in {
+        Failure("The Phantom Menace") collectFailure { case Failure("The Phantom Menace", Empty, Empty) => "Return Of The Jedi" } must_== Full("Return Of The Jedi")
+        Failure("The Phantom Menace") collectFailure { case Failure("The Phantom Menace", Empty, Empty) => Failure("Attack") } must_== Full(Failure("Attack"))
```
Shadowfiend: So a random interesting observation here tying into your question about making `flip` take a `PartialFunction`… Now that we've added `flip`, `collectFailure` is really just a special subcase of `flip`… Do we want to keep both? Or do we, as you initially suggested, make `flip` take a `PartialFunction`, and drop `collectFailure`?

Also of interest is that `mapFailure` and `collectFailure` work very differently despite their very similar names (probably the reason why initially you defined `collectFailure` as `PartialFunction[Failure,Failure]` instead of `PartialFunction[Failure,B]`).

Not sure what the best way forward is, but I'm slightly concerned about these two things… Would be interested in thoughts/ideas before we commit to this setup as it currently stands. Going with `mapFailure` + `flip` seems like a good way to handle the common use cases without adding the oddity of `collectFailure`.

Bhashit: I do like `flip` that takes a `PartialFunction`, in which case, `collectFailure` is indeed a special case of `flip`. I'll go ahead and make the changes over this weekend. 

Also, on a related note, let me know if we later need to adapt these methods (or any existing ones)  to the new hierarchy you are implementing. I'd be happy to take that up.

Shadowfiend: Sounds good. I expect this PR will go in before mine, and if that's the case I'll take on the task of pulling it into my refactoring. If mine goes in first, I'll let you take it on :)

Bhashit:
@Shadowfiend changes pushed. Although, the following case of `flip` did make me somewhat uncomfortable.

```scala
Empty flip { case Failure("error", Empty, Empty) => "alternative" }
```
This would return `Empty`. The thing that bothered me was that `flip` sounds definitive, whereas `collect` has a well-known meaning that it will only apply in some cases. Although, I think that should be okay.

Bhashit:
Just a gentle reminder (not sure if I'm supposed to do that): I think this is ready for review. The only remaining codacy complaint is about the `Failure` having a method `mapFailure` ("You should not name methods after their defining object").

Shadowfiend:
You are well within your rights to poke me, yes :) I've been short on time, but will have another look soon. Can you remove the WIP from the PR title?

Shadowfiend:
> Although, the following case of flip did make me somewhat uncomfortable.

This is a very, *very* fair concern... Enough so that I think it maybe calls into question the idea of giving flip a `PartialFunction`.

That point made me think a little bit about `mapFailure` as well. The trouble with that name is that it implies a tie to `map`, but `map` deals with the contents of a `Box`, not with the `Box` itself. I'm wondering if we tilt this a little bit… We keep `flip` as a total function that can flip `EmptyBox` into `Full`. And we introduce `def transform[A](transformFn: PartialFunction[Box[A], Box[A]): Box[A]`. `transform` can transform the box in any way it wants, as long as the contained value type remains unchanged (to change the contained value type you'd have to use `collect` or `map`.

This definition of `transform` encompasses what we currently consider `mapFailure` (by simply passing a `PartialFunction[Failure, Failure]`). It also encompasses the `PartialFunction` variant of `flip`. And it leaves `flip` to be as definitive as it sounds.

Sorry to come back with yet another “let's rework this yet again”---let me know what you think, and if you have time to look at doing it. If it sounds good to you, I can also make the change and let you review it. Whatever works! I think it would be great to see this land by M2 next week.

Bhashit:
Sounds good. I'll try to get that in over this weekend. Please don't worry about revising. That's what gets us a good API eventually.

Shadowfiend:
Absolutely, but I try to keep in mind how long these feedback cycles have been…  I appreciate your patience!

Bhashit:
I'm somewhat embarrassed that I didn't spot this until I started working on it: since Box is covariant in type A, and the return type param of PartialFunction is contravariant, we can't use `PartialFunction[Box[A], Box[A]]` here.

We'll need to use something like:

```scala
transform[B >: A](transformFn: PartialFunction[Box[A], Box[B]]): Box[B]
```

One caveat with that signature is that, unless `A` and `B` are the same type, the returned box will be some common super-type of both (and in may cases, that turns out to be `Any`).

What do you think? Any other ideas, or is this the way we want to go?

I'm pushing some changes, as I think this might be okay since users will probably know they are doing.  (just for review, we can revise as needed)

core/common/src/main/scala/net/liftweb/common/Box.scala
```diff
@@ -576,18 +552,18 @@ sealed abstract class Box[+A] extends Product with Serializable{
...
    * Returns an `[[scala.collection.Iterator Iterator]]` over the value
    * contained in this `Box`, if any.
    */
   def elements: Iterator[A] = Iterator.empty
 
   /**
+    * Return this Box if `Full`, or the specified alternative if it is
+    * empty. Equivalent to `Option`'s `[[scala.Option.orElse orElse]]`.
+    */
+  def or[B >: A](alternative: => Box[B]): Box[B] = transform { case _: EmptyBox => alternative }
```
Shadowfiend: While this is quite clean, I wonder if we should leave `or` as-is. The JVM will be able to turbo-inline the simple definitions of `or`, whereas this definition will probably (though not necessarily) be near-impossible to optimize. Given how common it is to call `or`, I'm tempted to say we should leave it as-is.

core/common/src/main/scala/net/liftweb/common/Box.scala
```diff
@@ -814,30 +790,39 @@ sealed abstract class Box[+A] extends Product with Serializable{
   }
 
   /**
-    * Returns a `Full` box containing the results of applying `flipFn` to this box if it is a `Failure`,
-    * `ParamFailure` or `Empty`, and `flipFn` is defined for it. Returns `Empty` if this box is `Full`.
-    * In other words, it "flips" the full/empty status of this Box.
+    * Transforms this box using the `transformFn`. If `transformFn` is defined for this box,
+    * returns the result of applying `transformFn` to it. Otherwise, returns this box unchanged.
```
Shadowfiend: I think we should at a note here that if you are looking to change the value *inside* a `Full` box, you should really be using `map`, and if you're looking to change a `Failure` or `Empty` into a `Full`, you should really be using `flip`.

There's enough complexity in `Box` that providing some helpful guides in some of this method-specific documentation starts making sense for me. What do you think?

core/common/src/main/scala/net/liftweb/common/Box.scala
Bhashit: Hmm. Yeah, I didn't think it along those lines. I'll change that back.

core/common/src/main/scala/net/liftweb/common/Box.scala
Bhashit: Yeah, that sounds pretty helpful. Will do so.

core/common/src/main/scala/net/liftweb/common/Box.scala
Bhashit: Do you think keeping the `or` abstract in the `Box` trait will make much of a difference? I mean, that way, there's one implementation in `Full` and one in `EmptyBox`. It sort of seems cleaner.

Shadowfiend: I think that's fine.

core/common/src/main/scala/net/liftweb/common/Box.scala
```diff
@@ -793,6 +793,10 @@ sealed abstract class Box[+A] extends Product with Serializable{
     * Transforms this box using the `transformFn`. If `transformFn` is defined for this box,
     * returns the result of applying `transformFn` to it. Otherwise, returns this box unchanged.
     *
+    * If you want to change the content of a `Full` box, using [[map]] or [[collect]] might be better
+    * suited to that purpose. If you want to convert an `Empty`, `Failure` or a `ParamFailure` into a
+    * `Full` box, you should use [[flip]].
```
Shadowfiend: I'd put backticks around `flip`, `map`, and `collect` (in addition to the `[[` and `]]`).

Bhashit: Generally, as a part of the review, I would take the words `I'd put backticks` to mean that I (as the author) should make that change. But since you approved, I am guessing you are making that change? Let me know if I should do it, and I'll push the change.

Bhashit:
I think you did the most of the hard work here though. So, a 'thank you' is in order. :bowing_man:

Shadowfiend:
Yeah, the request was for you to make the change, but I approved the PR because there's really no reason to block on it :)

Bhashit:
When this gets merged, do you need to implement them in your branch? More specifically, you need an extra hand with that? Otherwise, I'm planning to work on the vanilla JS artifacts this week.

Shadowfiend:
I'll disentangle my branch, no worries. Have to finish up test coverage in there before starting on that anyway.

Shadowfiend:
If anyone else wants to have a glance here and sign off, I'm good on merging… I'll probably merge later this week otherwise.

Shadowfiend:
Bit late, buuuut in we go!

Shadowfiend:
Tweaked the description and title to show what we ended up with ;)

Issue #1794, opened 2016-05-09
buildDeferredFunction, OfflineRequestSnapshot and serverPort as a lazy val.


Bhashit:
This is related to the post I made [here, in the Google group](https://groups.google.com/forum/#!topic/liftweb/2CdURDdKigA). I'll repeat the contents here:

I have a piece of code where the snippet just puts a comet on the page, and then the rest of the work is carried out using Futures. When those Futures complete, the results are sent to the page via the comet-actor.

Now, here's the problem I am facing.
1. I am using `LiftSession#buildDeferredFunction` in order to be able to execute my async code in context of the original request. 
2. `buildDeferredFunction` snapshots the current request, along with a lot of other stuff. As a result, an instance of `net.liftweb.http.provider.servlet.OfflineRequestSnapshot` is created. 
3. In this instance, all the details of the original request are stored using vals. However, only the `serverPort` is specifed as a lazy val. This `serverPort` val refers to a `org.eclipse.jetty.server.Request`, which happens to be mutable.
4. Now, in my Future code, when I need to get the `serverPort`, my `OfflineRequestSnapshot` tries to get it by accessing the jetty Request instance which it has a reference to. 
5. However, there seems to be a race-condition and, intermittently, the jetty Request instance gets changed in the middle (Jetty is reusing it?) and the details get lost, including the uri (which is set as a _uri field in that class).

Now, I can of course, get the server-port in the snippet, and pass it to the Future, along with the comet-name. However, it does seem like we could do something about it while creating the snapshot. This is because there could be quite a bit of code that might use this information. For ex. `S.hostAndPath` will refer to it.

@Shadowfiend suggested in that post that we could do something about it without losing the performance benefits.

Thanks.

Bhashit:
@Shadowfiend I can take care of this issue, and submit a PR. P.S. I'm familiar with the contributing guidelines.

Shadowfiend:
Go for it!

Shadowfiend:
Looks like this got all closed and stuff :tada:

PR #1865, opened 2017-06-17
Save the server-port of the original request

Bhashit:
Issue #1794 

store the server-port of the original request since it's used in a lazy val, and Jetty will recycle the request instance if the lazy val is realized too late

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/provider/servlet/HTTPRequestServlet.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/provider/servlet/HTTPRequestServlet.scala
@@ -19,9 +19,9 @@ package http
 package provider 
 package servlet 
 
-import java.io.{InputStream}
-import java.util.{Locale}
-import javax.servlet.http.{HttpServletRequest}
+import java.io.InputStream
+import java.util.Locale
+import javax.servlet.http.HttpServletRequest
 import org.apache.commons.fileupload.servlet._
 import org.apache.commons.fileupload.ProgressListener
 import net.liftweb.common._
@@ -127,7 +127,7 @@ class HTTPRequestServlet(val req: HttpServletRequest, val provider: HTTPProvider
     } yield id
 
   def extractFiles: List[ParamHolder] = (new Iterator[ParamHolder] {
-    val mimeUpload = (new ServletFileUpload)
+    val mimeUpload = new ServletFileUpload
     mimeUpload.setProgressListener(new ProgressListener {
       lazy val progList: (Long, Long, Int) => Unit = S.session.flatMap(_.progressListener) openOr LiftRules.progressListener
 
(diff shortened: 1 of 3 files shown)


web/webkit/src/main/scala/net/liftweb/http/provider/servlet/HTTPRequestServlet.scala
```diff
@@ -234,8 +236,8 @@ private class OfflineRequestSnapshot(req: HTTPRequest, val provider: HTTPProvide
 
   val scheme: String = req.scheme
 
-  lazy val serverPort: Int = req.serverPort match {
-    case 80 => headers("X-SSL").flatMap(Helpers.asBoolean _).filter(a => a).map(a => 443).headOption getOrElse 80
+  lazy val serverPort: Int = _serverPort match {
+    case 80 => headers("X-SSL").flatMap(Helpers.asBoolean).filter(identity).map(a => 443).headOption getOrElse 80
```
farmdawgnation: While we're here, can we change this to be a bit more readable?

```scala
case 80 =>
  headers("X-SSL")
    .flatMap(Helpers.asBoolean)
    .filter(identity)
    .map(a => 443)
    .headOption
    .getOrElse(80)
```

Also, what's the point of the `filter`?

Bhashit: Formatting done. 

And about that `filter` call: It seems to exist because the `headers` method returns a `List[String]` and we want to check if there is any `true` value.

But, now that I looked into it a bit more, I think there might be something wrong with how the `headers` method is used here. It doesn't return header values, it just returns the names of the headers, which are never going to be converted to a `true` value and the result of this entire chain is always 80.

I am not sure if I should push a change here or create a ticket.

farmdawgnation:
> And about that filter call: It seems to exist because the headers method returns a List[String] and we want to check if there is any true value.

The `filter` call is probably there because the headers method returns one type and we need a different type. (e.g. got an Iterator need a List) Can you look into that?

> But, now that I looked into it a bit more, I think there might be something wrong with how the headers method is used here. It doesn't return header values, it just returns the names of the headers, which are never going to be converted to a true value and the result of this entire chain is always 80.

Hm, that doesn't look good. I don't have time to look at it right now, but can you build out a unit test demonstrating that? If you're right, I think we should look to fix it here.

Bhashit:
About the `filter` call, yeah, it is like that. I mean, just the way of checking if there really is a header named `X-SSL`.

About the `headers` method: I'll push some changes here later

Bhashit:
@farmdawgnation Added a quick-n-dirty test, just for demo, to show that the current `headers` implementation is incorrect. Can be run with `test-only net.liftweb.http.provider.servlet.OfflineRequestSnapshotSpec`.

Also added a demo of a correct implementation. These are throwaway changes, and I'll clean them up if it looks okay to you.

web/webkit/src/main/scala/net/liftweb/http/provider/servlet/HTTPRequestServlet.scala
```diff
@@ -178,7 +178,7 @@ class HTTPRequestServlet(val req: HttpServletRequest, val provider: HTTPProvider
   }
 }
 
-private class OfflineRequestSnapshot(req: HTTPRequest, val provider: HTTPProvider) extends HTTPRequest {
+private [servlet] class OfflineRequestSnapshot(req: HTTPRequest, val provider: HTTPProvider) extends HTTPRequest {
```
Shadowfiend: If we're making this `private[servlet]`, I wonder if we just move it outside of `HTTPRequestServlet` and into its own file altogether? Not necessarily a pre-req for merging, just occurred to me while reading.

web/webkit/src/test/scala/net/liftweb/http/provider/servlet/OfflineRequestSnapshotSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+import net.liftweb.mockweb.WebSpec
+import org.mockito.Mockito._
+import org.specs2.mock.Mockito
+
+
+object OfflineRequestSnapshotSpec extends WebSpec with Mockito {
+
+  "A snapshot request should" in {
+    val mockHttpRequest = mock[HTTPRequest]
+    val mockProvider = mock[HTTPProvider]
```
Shadowfiend: This could probably just be `new HTTPProvider { val context = _ }`---no huge difference, just a preference for using the language instead of mock magic when possible.

web/webkit/src/test/scala/net/liftweb/http/provider/servlet/OfflineRequestSnapshotSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+
+import net.liftweb.http.provider._
+import net.liftweb.mockweb.WebSpec
+import org.mockito.Mockito._
+import org.specs2.mock.Mockito
+
+
+object OfflineRequestSnapshotSpec extends WebSpec with Mockito {
+
+  "A snapshot request should" in {
```
Shadowfiend: This can be `"A snapshot request" should {`. Also, I would suggest folding in the name of the class under spec, i.e. `"An OfflineRequestSnapshot" should {` or even (my personal preference) `"OfflineRequestSnapshot" should {`. This makes it easier when seeing a failed test to know exactly what class failed.

web/webkit/src/main/scala/net/liftweb/http/provider/servlet/HTTPRequestServlet.scala
Bhashit: Sounds like a good idea.

web/webkit/src/test/scala/net/liftweb/http/provider/servlet/OfflineRequestSnapshotSpec.scala
Bhashit: Absolutely. I just wrote this up as a quick-n-dirty example to show the current implementation didn't work correctly. To be improved if approved. I guess this means I should go ahead and do the cleanup.

web/webkit/src/test/scala/net/liftweb/http/provider/servlet/OfflineRequestSnapshotSpec.scala
Bhashit: Setting the `context` value to `_` results in `unbound placeholder parameter`. I think that only works for class level `vars`.

Setting `context` to null doesn't feel right to me. I also thought about setting it to an instance of `net.liftweb.http.LiftFilter`, which implements the `HTTPProvider` trait, but then it would look like it has some purpose. So, considering these two scenarios, I think `mock` is probably appropriate here?

Shadowfiend: Yeah, right you are! `= null` it would have to be.

It's not “right” in the sense that if this were live code it would be a bad idea. Since this is a test, I usually care a lot less.

Think of it this way: with the mock, if someone tried to call something on `context`, things would blow up (because there's no expectation set up for that mock). So in that sense it's the same as if you set a `null` there: interacting with the object in question will cause issues. But, we save ourselves whatever weird plumbing and other magic the mocking framework does, instead favoring straightforward language constructs.

Bhashit: Now that you said it, `null` seems quite fitting. Pushed a change.

Bhashit:
The Codacy issues are about converting `private` to `private [this]`, and there's one about using that `null` we discussed about in a test.

Shadowfiend:
Fwiw I do think the `private[this]` advice is worth following. Not a big deal though.

Bhashit:
Yeah. In fact, I started using private[this] as a default for my current project from today. Ha ha. Will be making that change here as well.

Bhashit:
Found that the `param` method was incorrect in the same manner that the `header` method was. So, corrected it as well.

web/webkit/src/test/scala/net/liftweb/http/provider/servlet/OfflineRequestSnapshotSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+
+      snapshot.param("tennis") shouldEqual tennisParams
+      snapshot.param("Tennis") should beEmpty
+      snapshot.param("swimming") shouldEqual swimmingParams
+    }
+  }
+
+  private[this] val xSSLHeader = HTTPParam(X_SSL, List("true")) :: Nil
+
+  private[this] def getRequestSnapshot(originalPort: Int, headers: List[HTTPParam] = xSSLHeader, params: List[HTTPParam] = Nil) = {
```
farmdawgnation: `headers` should probably default to `Nil` otherwise it's not obvious in a casual reading of your tests that the SSL header is included.

web/webkit/src/test/scala/net/liftweb/http/provider/servlet/OfflineRequestSnapshotSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+      req.headers("X-SSL") shouldEqual List("true")
+      req.headers("Unknown") must beEmpty
+    }
+
+    "have the serverPort value" in {
+      "443 when the 'X-SSL' header is set to the string 'true' (case-insensitive) and original port is 80" in {
+        val port80Req = getRequestSnapshot(originalPort = 80)
+        port80Req.serverPort shouldEqual 443
+      }
+      s"equal to the original request-port when the '$X_SSL' header is absent or not set to the string 'true' (case-insensitive), or if the original port is not 80" in {
```
farmdawgnation: This looks ripe for splitting into three separate specifications. :)

PR #1871, opened 2017-06-29
Bumping logback version to 1.2.3 for a security vulnerability fix

joescii:
Just to be super conservative, bumping the dependency on logback to version `1.2.3`. Since this dependency is marked as `"provided"`, downstream consumers are not affected and hence this PR isn't strictly necessary. 

Vulnerability details: https://nvd.nist.gov/vuln/detail/CVE-2017-5929
Related play framework security advisory: https://www.playframework.com/security/vulnerability/20170407-LogbackDeser

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -57,7 +57,7 @@ object Dependencies {
 
   // Provided scope:
   // Scope provided by container, available only in compile and test classpath, non-transitive by default.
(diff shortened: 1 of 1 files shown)


farmdawgnation:
Not sure if we merge this before or after tomorrow when we declare Lift 3.1.0 final...

farmdawgnation:
Oh well, 3.1.0 has shipped so moot point now. Let's :shipit:.

PR #1876, opened 2017-07-07
Float Like A Butterfly: Introduce PresenceBox, TryBox, and ParamTryBox.

Shadowfiend:
These three intermediary types allow APIs and consumers of boxes to indicate
that they expect a particular subset of `Box` types. This is useful to restrict
types to what is expected, while still being able to interoperate with
APIs designed to interact with many all `Box` types.

For example, you can declare that a function expectes a `PresenceBox` to ensure
that interactors guarantee either a `Full` or `Empty` is passed. Or, you can
specify that your function returns a `TryBox` so callers only need to worry about
handling a `Full` or a `Failure`. Lastly, `ParamTryBox` allows indicating that a
given situation will only get a `Full` or a `ParamFailure` with a particular error
type.

A lot of tests and additional API conversions are required to fully bring this
vision to fruition, but this is the foundation for that work.

Required before this is ready to merge:
 - [x] Proper typing of all methods in `Box.scala`
 - [x] Initial battery of specs (with appropriate type annotations)

Of note, the compiler will already properly infer, say, `PresenceBox` if your
code only produces `Full` or `Empty`. It will also infer `ParamTryBox[String, Int]`,
for example, if your code only produces `Full` or `ParamFailure[Int]`.

Hoping I can get this wrapped up in time for 3.2.0-M2, but we'll see ;)


Bhashit:
Ah, looks like I was a bit late in pushing these changes. I did have most of it, was stuck with a serializer and didn't get enough time for that one. Looking great. :+1:

Shadowfiend:
Ugh, sorry @Bhashit… I kind of did this all in one day, and in my head you were working on the combinators but not the typing changes. Re-reading the original issue, I had no reason to think that, and should have let you know I was going to put pedal to the metal on cranking something out.

Shadowfiend:
Pushed a bunch more changes here… One note for you @Bhashit particularly for `collectFailure` in #1864---I had to un-`final` the `collect` methods to provide more specific types in `PresenceBox` for them.

Bhashit:
> in my head you were working on the combinators but not the typing changes

No problem at all. I am pretty sure you'd do a better job of it.

> I had to un-final the collect methods

Sounds good.

farmdawgnation:
Holding off on review here until the tests are passing, but this looks exciting. :)

farmdawgnation:
Tests still are not passing. :(

Shadowfiend:
Well no changes have been pushed so that seems perfectly reasonable :p

farmdawgnation:
BUT I WANT THE FEATURES NOWWWWWWWWWWWWW :baby:

Shadowfiend:
Sadly I had to jettison a properly-typed `flatten` implementation, at least for the time being. Details in [the commit](https://github.com/lift/framework/pull/1876/commits/a8f140202f3e3ac19effeec5d0401ffdfae866e8), but net-net I couldn't figure out how to get the compiler to play nice on making `TryBox[TryBox].flatten=>TryBox` and `PresenceBox[PresenceBox].flatten=>PresenceBox`. For now, that means we have `flattenTry` on `TryBox` and `flattenPresence` on `PresenceBox` that only compile if both types match and produce the right return type.

I don't like this gotcha one bit :(

core/common/src/main/scala/net/liftweb/common/Box.scala
```diff
@@ -785,34 +822,124 @@ sealed abstract class Box[+A] extends Product with Serializable{
...
+ * A `Full` or `Empty` box (stands in trivially for `Option`).
+ */
+sealed trait PresenceBox[+A] extends Box[A] {
+  override def isA[B](clsOrg: Class[B]): PresenceBox[B] = Empty
+  override def asA[B](implicit m: Manifest[B]): PresenceBox[B] = Empty
+
+  def or[B >: A](alternative: => Full[B]): Full[B]
+  def or[B >: A](alternative: => PresenceBox[B])(implicit a: DummyImplicit): PresenceBox[B]
+  def or[B >: A](alternative: => TryBox[B])(implicit a: DummyImplicit, b: DummyImplicit): TryBox[B]
+  def or[B >: A, E](alternative: => ParamTryBox[B, E])(implicit a: DummyImplicit, b: DummyImplicit, c: DummyImplicit): ParamTryBox[B, E]
```
farmdawgnation: I'm idly wondering if the presence of these `DummyImplicit`s might be a sign we've tried to be a bit too clever by incorporating this into the overall box structure instead of having them as a parallel data type with implicit conversions.

core/common/src/main/scala/net/liftweb/common/Box.scala
```diff
@@ -785,34 +822,124 @@ sealed abstract class Box[+A] extends Product with Serializable{
...
+
+  override def map[B](f: A => B): PresenceBox[B] = Empty
+  def flatMap[B](f: A => PresenceBox[B]): PresenceBox[B]
+  override def collect[B](pf: PartialFunction[A, B]): PresenceBox[B] = ???
```
farmdawgnation: We should remove the presence of `???` from anything that'll be production code.

core/common/src/main/scala/net/liftweb/common/Box.scala
```diff
@@ -785,34 +822,124 @@ sealed abstract class Box[+A] extends Product with Serializable{
...
+      new WithPresenceFilter(x => p(x) && q(x))
+  }
+}
+/**
+ * A `Full` or `Failure` box (somewhat similar to `Try`, but can carry an error
+ * message without needing an exception, and a parameter as in `ParamFailure`).
+ */
+sealed trait TryBox[+A] extends Box[A] {
+  def or[B >: A](alternative: => Full[B]): Full[B]
+  def or[B >: A](alternative: => PresenceBox[B])(implicit a: DummyImplicit): PresenceBox[B] // *
```
farmdawgnation: Rouge comment on this line?

core/common/src/main/scala/net/liftweb/common/Box.scala
```diff
@@ -785,34 +822,124 @@ sealed abstract class Box[+A] extends Product with Serializable{
...
+ * A `Full` or `ParamFailure` box; `A` is the type in the box, `E` is the type
+ * of the `ParamFailure`'s param.
+ */
+sealed trait ParamTryBox[+A, +E] extends Box[A] with TryBox[A] {
+  def or[B >: A](alternative: => Full[B]): Full[B]
+  def or[B >: A](alternative: => PresenceBox[B])(implicit a: DummyImplicit): PresenceBox[B] // *
+  def or[B >: A](alternative: => TryBox[B])(implicit a: DummyImplicit, b: DummyImplicit): TryBox[B]
+  def or[B >: A, E2](alternative: => ParamTryBox[B, E2])(implicit a: DummyImplicit, b: DummyImplicit, c: DummyImplicit): ParamTryBox[B, E2]
+
+  override def map[B](f: A => B): ParamTryBox[B, E] = ???
```
farmdawgnation: Ditto here with `???`

core/common/src/main/scala/net/liftweb/common/Box.scala
Shadowfiend: Parallel data types with implicit conversions are only an option for a small subset of these methods, because of the fact that some of the methods overlap with Iterable and such, which means if someone imports `Box._` (which is not an unreasonable thing to do in a few situations), they'll be unable to get our implicits at a higher specificity than the iterable conversions.

It ain't fun.

That said, these dummy implicits don't have anything to do with the way in which we attach these methods; instead, they are here because we're trying to overload on an parameter type with a call-by-name parameter. Overloading on parameter type is directly supported in the JVM, but by making these by-name we turn them into parameterized `Function1`s, or something roughly equivalent. That means we're suddenly overloading by the parameter's *generic* type, which isn't directly supported. Dummy implicits step in to fill this gap.

core/common/src/main/scala/net/liftweb/common/Box.scala
Shadowfiend: I'll double-check this. I introduced it for a reason, but it may have been an earlier iteration.

core/common/src/main/scala/net/liftweb/common/Box.scala
Shadowfiend: Tru.

Shadowfiend:
> I'm wondering whether or not we've made Box harder to understand conceptually and therefore increased the barrier to entry for using it and contributing to it.

We've definitely increased complexity here, but a lot of the hoops we jump through here are to make the intuition of this added complexity work as expected.

Specifically, we added `PresenceBox`, `TryBox`, and `ParamTryBox`. These let you specify at a type level that you expect certain subsets of the box hierarchy---a situation we find ourselves in quite frequently, more often than not because we're trying to communicate failure-or-full and there's no way to do that without also having to handle the empty state. From a user perspective, this should be the only complication. The rest of the code in this PR (and there is a good bit of added code) is to ensure that the various `Box` combinators produce the correct subsets in their type signatures, so that if , say, I want to get a `TryBox` and I have to do some `flatMap`s to get there, I don't immediately have to resort to type-casting (because `flatMap` is aware of what's being `flatMap`ped and can maintain proper typing throughout the operation).

With that said, it definitely increases the barrier to entry for contributing to `Box`. Personally I can't think of anything else that we would need to add, but of course you can't predict the future. I think from continuous experience that this gets us:

 - A solution to a common problem.
 - A setup to prevent potentially dangerous type conversions in the future. This one specifically is about the implicit `box2Option` conversion, which today is always potentially lossy. In a world where `PresenceBox` is a thing, we can imagine giving `PresenceBox` an implicit conversion to `Option` but `Box` would not (obviously we wouldn't do this until a breaking release cycle, though).

I definitely understand the complexity concern here though, so if we collectively think this is too much, I'm happy to leave it on the table.

core/common/src/main/scala/net/liftweb/common/Box.scala
Shadowfiend: Worth noting though---neither of the `???` implementations are used in practice, because they are overridden further in all of their subclasses, and dynamic binding ensures it's those implementations, not these, that are executed.

farmdawgnation: Yeah, I'm not concerned about that. More about accident proofing the code. If there is a `???` implementation here, the compiler would say, for example, "Hey this interface is satisfied!" If someone were to somehow break the override of a child implementation, this would become the used implementation.

Whereas, if we _don't_ provide `???` here, breaking the override implementation should break the compile.

farmdawgnation:
I think the complexity concern here is probably worth it given what we're getting. And I can't think of any way to simplify things significantly.

Bhashit:
FWIW, I'd love having this new hieararchy. The pattern matches with `Empty` and `Failure`, even when they are not possible outcomes, always feels impure.

Shadowfiend:
Not quite complete yet, but I've added the structure I'll be using for the typing tests, as well as a first tranche of tests.

I also figured out why `collect` has `???` implementations---without it, the concrete `collect` implemented at the `Box` level interferes with the one defined on the traits, and things won't compile. I have an idea or two I need to see if I can try before sticking to this workaround though, and if we do stick with it I'll add an appropriate note.

Shadowfiend:
Ok, so slight complication here which we'll need to decide if we're okay with… In a sense, this branch is too effective heh. `?~` now returns `TryBox`, as it should. This means that code that has a `for` comprehension where each step uses `?~` to ensure a `Failure` pops out even in case of `Empty`, and then pattern-matches on the result, will fail compilation if the developer pattern-matched against `Empty` (if they didn't, they would have gotten a compiler warning prior to this).

We can change `?~` to return `Box` instead of `TryBox`, but we start winking at “what even is the point” territory. The compilation error would also be very easy to fix: The `Empty` case just needs to be nuked.

An example is this chunk of `WithParam`:

https://github.com/lift/framework/blob/aa24c2d966d005cd21d725a8ffa7a4790b6c6983/web/webkit/src/main/scala/net/liftweb/builtin/snippet/WithParam.scala#L44-L58

Shadowfiend:
Ok, wrapped up the specs, made a fix or two, nuked the `= ???` definitions, and merged in master with @Bhashit's changes...

I think I consider this PR good to go, provided travis doesn't slap me repeatedly in the face with red :x:es.

Shadowfiend:
Ohai Scala 2.11. Looks like still some cleanup to be done. Will circle back!

Shadowfiend:
> this is acceptable

Don’t know what I was thinking. Minor releases are API-compatible, by definition. So... that’s awkward heh. Will noodle.

Shadowfiend:
Moved this to hypothetical 4.0 for now… Really bummed about it though :(

Shadowfiend:
(May be worth discussing if this is worth doing an earlier-than-expected 4.0 release with the community… Perhaps after another couple of minor releases.)

farmdawgnation:
Hey there, I'm closing PR as orphaned since it's been more than a year since it's had any updates. Feel free to reopen when you have time to update it!

Shadowfiend:
Not much to update here, tbh. This PR was left open because it can't land before a 4.0 release; it may have some merge conflicts but I suspect I'd be able to dispense with them when the time comes.

farmdawgnation:
Hm, gotcha. I can reopen it.

farmdawgnation:
Also converted it to a Draft so we know it's not an "actually actionable thing right now

PR #1881, opened 2017-07-20
Lift json performance optimizations

chriswebster:
https://groups.google.com/forum/#!topic/liftweb/uBzyiDEQsyI

This motivation for this PR is to improve performance of json case class serialization. This change contains two optimizations:

* In appendEscapedString requires a boxing operation to transform Char into Character for the Set.contains(java.lang.Object) operation, even is the escapedChars isEmpty (the default)

- The patch eliminates this check when the settings.escapeChars isEmpty using && to short circuit the Set.contains method 

* In  Meta memoize there is an extra lookup and Option creation as both Map.contains and Map.apply are implemented using Map.get for the case where the key has already been cached (one for contains and another for get, which is creating additional Option instances). 

- The patch uses getOrElse to avoid the second lookup (the or else populates the cache as a side effect).

### DIFF
--- a/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
+++ b/core/json/src/main/scala/net/liftweb/json/JsonAST.scala
@@ -827,7 +827,8 @@ object JsonAST {
         case '\n' => "\\n"
         case '\r' => "\\r"
         case '\t' => "\\t"
-        case c if ((c >= '\u0000' && c < '\u0020')) || settings.escapeChars.contains(c) =>
+        // Set.contains will cause boxing of c to Character, try and avoid this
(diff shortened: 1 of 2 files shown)


Shadowfiend:
Would you mind amending the commit to also include the notes you've included on the PR description? Looks like some great sleuthing… I'm feeling pretty 🤦‍♂️ y about `appendEscapedString` :/

What kind of performance boosts are we seeing here? Should we be considering a point release for 3.1?

chriswebster:
Added the notes from the PR.

Shadowfiend:
:heart:

Shadowfiend:
Given two approvals I think we'll roll 'er on in.

PR #1882, opened 2017-07-22
The Extraction mkMapping and decompose methods have code to determine…

chriswebster:
… if a tuple is being used and whether this should be considered as formatted as array. The tuple_? method is a Seq that contains the classes for the 22 varieties of tuples and the tuple_? method does a O(n) scan on this frequently during the case class serialization / deserialization process. Currently, the code is performing contains first and then checking whether Tuples are formatted as arrays (where the default is false). This change reverses the order to short circuit this check.

https://groups.google.com/forum/#!topic/liftweb/E8tHa_5MWt8


farmdawgnation:
✅  @chriswebster has [signed the contributors file](https://github.com/lift/framework/blob/master/contributors.md#name-22)
✅  There is a ML thread where a committer has asked for this fix.

This is golden for merge. Thank you for the contribution!

farmdawgnation:
(Going to wait a beat on the merge since we're having some issues cutting the 3.2.0-M1 build, however.

farmdawgnation:
Oh, JK, 3.2.0-M1 is out! Huzzah! Merging!

PR #1884, opened 2017-07-22
Fix example in lift-json's Xml helper.

Shadowfiend:
We had some outdated example code that used `map` instead of `mapField`.

Spotted [on Stack Overflow](https://stackoverflow.com/questions/43100997/compilation-error-in-code-snippet-by-lift-json-3-0-1-to-convert-json-arrays-into).


Shadowfiend:
Bundled in a fix for a spec that just failed, which was expecting ordering out of three parallel-y running `LAFuture`s. Bad plan! :)

PR #1885, opened 2017-07-23
Validate that closing a session from a comet doesn't NPE

### LINKED ISSUE #1145: NullPointerException when trying to run LiftSession.destroySession with mock req instance in comet
Lift throws a NullPointerException when attempting to run destroySession from a mock req instance with a LiftSession stub, as when trying to run:

for (liftSession <- S.session) {
  liftSession.destroySession
}

See https://groups.google.com/d/msg/liftweb/-/xOauoxgBzmUJ for discussion.

farmdawgnation:
Closes #1145.

In the days of ye old Lift comet actors were using mock request instances. This meant that it was possible to get yourself into the somewhat surprising situation of triggering an NPE from a comet if you attempted to delete a session even though comets are associated with valid sessions.

We've since stopped that practice and now seem to plug in an `Empty` for the request when we're inside a `Comet` as of #1613. As such I was unable to reproduce this issue from a unit test. Decided to open up a PR with a unit test to ensure we don't accidentally regress in the future.

### DIFF
--- a/web/webkit/src/test/scala/net/liftweb/http/CometActorSpec.scala
+++ b/web/webkit/src/test/scala/net/liftweb/http/CometActorSpec.scala
@@ -94,15 +94,27 @@ object CometActorSpec extends Specification {
     }
 
     "be able to invoke destroySession without causing an NPE" in {
+      var didRun = false
+      var didThrow = false
+
       case object BoomSession
       val comet = new SpecCometActor {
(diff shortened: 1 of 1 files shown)


Shadowfiend:
Just going to merge this guy, since it's just a spec, the spec passed, and it's for behavior we want to be checking.

PR #1887, opened 2017-07-25
Fixing stackoverflow in typechecker while building tests

### LINKED ISSUE #1883: Investigate stack overflows in typechecker while building tests
We've been having some issues building the past few milestones wherein we end up with stack overflows while trying to build the lift-common tests in the typechecker. It seems to be specific to OpenJDK Java 1.8.0_131 in Linux. We should investigate what's up with this so that our build process isn't totally painful for @fmpwizard.

karma4u101:
This PR closes #1883 

After increasing the JVM stack size by adding -Xss256m to liftish INTERNAL_OPTS I can no longer reproduce the #1883 issue. 
I also bumped the memory allocation pool to -Xmx2048m although I am not sure that was strictly necessary (as shouldn't normally increasing the stack size).

### DIFF
--- a/liftsh
+++ b/liftsh
@@ -21,7 +21,7 @@ if test -f ~/.liftsh.config; then
 fi
 
 # Internal options, always specified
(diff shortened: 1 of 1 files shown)


farmdawgnation:
:+1: from me.

farmdawgnation:
Barring any further comments, I'm shipping it!

PR #1889, opened 2017-07-26
Safe settings

joescii:
**[Mailing List](https://groups.google.com/forum/#!topic/liftweb/-RQdaEkDCjM) thread**:

As described in the ML thread above, this is an attempt to make dealing with the mutability of `LiftRules` settings safer for the application developer. This mechanism can be used for new settings. Older settings cannot benefit without breaking API changes. 

I introduce a new class called `LiftRulesSetting[T]` which extends from the `ValueHolder` family of Lift types, hence its API is familiar to Lift developers with simple methods such as `set` and `get` to access the contained value.

The `LiftRulesSetting[T]` container type assists the application developer by defining exceptions for three different troublesome scenarios: 
1. A setting that is written after it has already been read: `LiftRulesSettingWrittenAfterReadException`
2. A setting that is written after Lift boot is complete: `LiftRulesSettingWrittenAfterBootException`
3. A setting which has been written twice with different values: `LiftRulesSettingWrittenTwiceException`

When one of these scenarios are encountered, the exception is created with the stack trace of the previous offending access. For instance, `LiftRulesSettingWrittenAfterReadException` will contain the stack trace of the last read. Or `LiftRulesSettingWrittenAfterBootException` will contain the stack trace of the first time it was set. (perhaps both traces should be included? I could probably make one exception and stacktrace the "cause" of the other). The idea is that the exception should do the best job we can to help the developer resolve the conflict.

What is actually done with the exception is configured with `LiftRules.settingsExceptionFunc` which is of type `LiftRulesSettingException => Unit`. I've added two reasonable implementations with `LiftRules.throwSettingsException` and `LiftRules.warnOnSettingsException`. The flexibility of the function setting will allow for an application to ignore a purposeful violation as the exceptions contain the name of the setting being changed in a suspect manner.


web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -142,6 +142,8 @@ object LiftRules extends LiftRulesMocker {
    */
   type StatelessTestPF = PartialFunction[List[String], Boolean]
 
+  type StackTrace = Array[StackTraceElement]
```
Shadowfiend: This seems like it'll do more harm than good tbh.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -1582,6 +1584,10 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
 
   private def logSnippetFailure(sf: SnippetFailure) = logger.info("Snippet Failure: " + sf)
 
+  val throwSettingsException: LiftRulesSettingException => Unit = e => throw e
+  val warnOnSettingsException: LiftRulesSettingException => Unit = e => logger.warn("LiftRules setting safety violation!!!", e)
```
Shadowfiend: Given the simplicity of these, I feel like we should just pick the one we want as default and let the user update it as they desire.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -2295,3 +2301,80 @@ trait FormVendor {
   private object requestForms extends SessionVar[Map[String, List[FormBuilderLocator[_]]]](Map())
 }
 
+abstract class LiftRulesSettingException(settingName: String, stackTrace: LiftRules.StackTrace, message: String) extends Exception(message) {
```
Shadowfiend: This, and its attendant subclasses, should probably live in its own file. I know we've embraced putting 18000 classes in each file in Lift, but IMO we should move away from that :)

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -2295,3 +2301,80 @@ trait FormVendor {
...
+  setStackTrace(stackTrace)
+}
+
+case class LiftRulesSettingWrittenAfterReadException(settingName: String, stackTrace: LiftRules.StackTrace, message: String)
+  extends LiftRulesSettingException(settingName, stackTrace, message)
+
+case class LiftRulesSettingWrittenAfterBootException(settingName: String, stackTrace: LiftRules.StackTrace, message: String)
+  extends LiftRulesSettingException(settingName, stackTrace, message)
+
+case class LiftRulesSettingWrittenTwiceException(settingName: String, stackTrace: LiftRules.StackTrace, message: String)
```
Shadowfiend: Rather than repeating the name in the exception, should we think about putting `WrittenAfterReadException`, etc, inside the `LiftRulesSetting` singleton?

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -2295,3 +2301,80 @@ trait FormVendor {
...
+  extends LiftRulesSettingException(settingName, stackTrace, message)
+
+
+class LiftRulesSetting[T](val name: String, val default: T) extends LiftValue[T] with HasCalcDefaultValue[T] {
```
Shadowfiend: Some scaladocs would be 👌, particularly advising when you would use this vs another approach.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -2295,3 +2301,80 @@ trait FormVendor {
...
+  import LiftRules.StackTrace
+
+  private [this] var v: T = default
+  private [this] var lastSet: Option[Exception] = None
+  private [this] var lastRead: Option[Exception] = None
+
+  private [this] def writeAfterReadMessage =
+    s"LiftRules.$name was set AFTER already being read! " +
```
Shadowfiend: I don't know that the emphasis on `AFTER` gives us much here.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -2295,3 +2301,80 @@ trait FormVendor {
...
+    s"Your Lift application may be booting into an inconsistent state. " +
+    s"Review the stacktrace below to see where the value was last read. " +
+    s"Consider re-ordering your settings in boot. "
```
Shadowfiend: Not sure if we should be this prescriptive, since I could see this also happening because something ran *after* boot.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -2295,3 +2301,80 @@ trait FormVendor {
...
+  private [this] def writtenTwiceMessage1(newVal: T) =
+    s"LiftRules.$name was set to $v then later set to $newVal. " +
+    s"This could potentially cause your Lift application to run in an inconsistent state. " +
+    s"Review the stacktrace below to see where LiftRules.$name was first set to $v. "
+
+  private [this] def writtenTwiceMessage2(newVal: T) =
+    s"Review the stacktrace below to see where LiftRules.$name was later set to $newVal. "
+
+  private [this] val toIgnore = Set("LiftRulesSetting", "LiftValue")
+  private [this] def trimmedStackTrace(t: Throwable): StackTrace =
```
Shadowfiend: Believe we've generally not put a space between `private` and `[this]`. Nitpick, though.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -2295,3 +2301,80 @@ trait FormVendor {
...
+
+  private [this] def writtenTwiceMessage1(newVal: T) =
+    s"LiftRules.$name was set to $v then later set to $newVal. " +
+    s"This could potentially cause your Lift application to run in an inconsistent state. " +
+    s"Review the stacktrace below to see where LiftRules.$name was first set to $v. "
+
+  private [this] def writtenTwiceMessage2(newVal: T) =
+    s"Review the stacktrace below to see where LiftRules.$name was later set to $newVal. "
+
+  private [this] val toIgnore = Set("LiftRulesSetting", "LiftValue")
```
Shadowfiend: This seems like it could live directly inside the `trimmedStackTrace` helper, no?

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -2295,3 +2301,80 @@ trait FormVendor {
...
+  private [this] def trimmedStackTrace(t: Throwable): StackTrace =
+    t.getStackTrace.dropWhile(e => toIgnore.find(e.getClassName contains _).isDefined)
```
Shadowfiend: Why not use fully-qualified names and make this `dropWhile(toIgnore contains _)`?

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -2295,3 +2301,80 @@ trait FormVendor {
...
+
+
+  override def set(value: T): T = {
+    lastSet.collect { case e if v != value =>
+      val e1 = LiftRulesSettingWrittenTwiceException(name, trimmedStackTrace(e), writtenTwiceMessage1(value))
+      LiftRules.settingsExceptionFunc.get.apply(e1)
+
+      val e2 = LiftRulesSettingWrittenTwiceException(name, trimmedStackTrace(e), writtenTwiceMessage2(value))
```
Shadowfiend: What's with this causing two exceptions?

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -2295,3 +2301,80 @@ trait FormVendor {
...
+      LiftRules.settingsExceptionFunc.get.apply(e)
+    }
+
+    // TODO: Skip if the value is the same?
+    lastRead.foreach { e =>
+      val e2 = LiftRulesSettingWrittenAfterReadException(name, trimmedStackTrace(e), writeAfterReadMessage)
+      LiftRules.settingsExceptionFunc.get.apply(e2)
+    }
+
+    lastSet = Some(new Exception)
```
Shadowfiend: You can stash the stack trace instead of an `Exception` via `Some((new Exception).getStackTrace)`. Generally it seems like that's our goal throughout, so we should be operating in terms of stack traces rather than exceptions.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -2295,3 +2301,80 @@ trait FormVendor {
...
+  private [this] def writtenTwiceMessage2(newVal: T) =
+    s"Review the stacktrace below to see where LiftRules.$name was later set to $newVal. "
+
+  private [this] val toIgnore = Set("LiftRulesSetting", "LiftValue")
+  private [this] def trimmedStackTrace(t: Throwable): StackTrace =
+    t.getStackTrace.dropWhile(e => toIgnore.find(e.getClassName contains _).isDefined)
+
+
+  override def set(value: T): T = {
+    lastSet.collect { case e if v != value =>
```
Shadowfiend: I generally don't love the use of functions that are meant to be collectors/mappers for side-effecting purposes. I know it lets us use a PF here, but I think there's little added cost to using `foreach` and `case _ => // if we've not set this property before, we're fine` to make it clear we're not trying to do anything with the resulting value.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
farmdawgnation: Ditto to the above comment. This is consistent with what we do with other error handling behavior.

joescii:
Thanks for the feedback guys! I must have something misconfigured because I have not gotten any email notifications for any of this activity. I'm going to simply address the points we agree with via code change, and leave comments where I need clarification.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
joescii: Really? I dig simple type aliases like this, but please do inform what you see could be troublesome here.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
joescii: LOL, the same thoughts went through my mind when I tossed these in here with the rest. :)

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
joescii: This is definitely an area I needed some feedback on. I'll try something very short and sweet.

joescii: Upon second thought, I'm going to trim the overly descriptive content and let you take a look.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
joescii: Ah yes, you caught me making an old micro-optimization that the JVM probably doesn't need any help doing.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
joescii: Ah yeah... I'm bouncing among several different projects with different conventions :)

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
joescii: Take a look at this stacktrace: 
```text
at net.liftweb.http.LiftRulesGuardedSetting.get(LiftRulesGuardedSetting.scala:71)
at net.liftweb.util.LiftValue$class.is(ValueHolder.scala:79)
at net.liftweb.http.LiftRulesGuardedSetting.is(LiftRulesGuardedSetting.scala:17)
at bootstrap.liftweb.Boot.boot(Boot.scala:62)
[...]
```

And this stacktrace:
```text
at net.liftweb.http.LiftRulesGuardedSetting.get(LiftRulesGuardedSetting.scala:71)
at net.liftweb.util.LiftValue.is(ValueHolder.scala:79)
at net.liftweb.util.LiftValue.is$(ValueHolder.scala:79)
at net.liftweb.http.LiftRulesGuardedSetting.is(LiftRulesGuardedSetting.scala:17)
at bootstrap.liftweb.Boot.boot(Boot.scala:62)
[...]
```

Notice the difference in `LiftValue`'s class name, as far as the JVM is concerned. The first stacktrace includes one of Scala's `$` insertions to dance around the JVM's nominative typing details.

The difference between the stacktraces? The first is Scala 2.11.11 but the second is Scala 2.12.2. 

For us to use FQN, we have to account for the Scala compiler. I don't feel too awesome about that.

joescii: BTW, I should have mentioned that your proposal works flawlessly on 2.12. I'd have to add `net.liftweb.util.LiftValue$class` for 2.11

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
joescii: Yeah, I'm very much in agreement with you. I suspect my use of `collect` for a side-effect this way happened as the code evolved, and I didn't reevaluate.

joescii:
So @farmdawgnation [doesn't like flagging setting one of these twice](https://github.com/lift/framework/pull/1889#pullrequestreview-54536383) and @Shadowfiend [doesn't understand why there are two exceptions](https://github.com/lift/framework/pull/1889#discussion_r131521843). 

The concern I hoped to address here is the chance that a module sets a value and the developer unknowingly changes that value. I suspect this is a low-probability issue, but I was inspired to cover as many bases as I could imagine.

Drop this check altogether?

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
joescii: Yeah, I agree this is better. Originally I was concerned about processing too many unnecessary stacktraces, but I've already ironed out the performance concerns such that this change looks safe.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
Shadowfiend: It adds a layer of indirection for a very simple type, resulting in confusion about what the type is. I'd rather see `Array[StackTraceElement]` than `StackTrace` -> look up what that means -> realize it's just an `Array[StackTraceElement]` any day of the week.

Might be personal preference though. I almost never use type aliases.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
Shadowfiend: We can look up FQNs at runtime though, right? Just `Set(LiftRulesSetting.getClass.getName, LiftValue.getClass.getName)` or somesuch? If not, I'm down with this---it's not a huge deal anyway.

web/webkit/src/main/scala/net/liftweb/http/LiftRulesGuardedSetting.scala
```diff
@@ -0,0 +1,126 @@
...
+import net.liftweb.util.{HasCalcDefaultValue, LiftValue}
+
+object LiftRulesGuardedSetting {
+  type StackTrace = Array[StackTraceElement]
+
+  /**
+    * Base class for all possible violations which LiftRulesGuardedSetting warns you about
+    * @param settingName
+    * @param stackTrace
+    * @param message
```
Shadowfiend: Down to document these parameters if they seem nonobvious, but empty `@param`s make me quake in me boots. But not from fear :laughing:

Shadowfiend: Down to document these parameters if they seem nonobvious, but empty @params make me quake in me boots. But not from fear 😆

web/webkit/src/main/scala/net/liftweb/http/LiftRulesGuardedSetting.scala
```diff
@@ -0,0 +1,126 @@
...
+    */
+  abstract class SettingViolation(settingName: String, stackTrace: StackTrace, message: String) extends Serializable {
+    /**
+      * Converts this violation into an Exception which can be handed to a logger for clean message printing
+      * @return
```
Shadowfiend: Same for empty `@return`s.

joescii: Alright, fine on both. This is where I kinda hate scaladoc because it wants me to add more words to my already descriptive names. LOL

joescii:
"Killer", huh? :D I wish I could have y'all review all of my code :)

web/webkit/src/main/scala/net/liftweb/http/LiftRulesGuardedSetting.scala
Shadowfiend: For what it's worth, I almost never specify `@return` or `@param` for this reason. My functions tend to be tight enough and the parameter and function name descriptive enough that these should be unnecessary.

Not always the case, but that's how I deal with the annoynace :)

Shadowfiend: For what it's worth, I almost never specify `@return` or `@param` for this reason. My functions tend to be tight enough and the parameter and function name descriptive enough that these should be unnecessary.

Not always the case, but that's how I deal with the annoyance :)

Shadowfiend:
🙇 can't ask for a greater compliment.

Reviewing 5-10 PRs a week means I have to continually come up with new ways to express approval so nobody gets bored :D

web/webkit/src/main/scala/net/liftweb/http/LiftRulesGuardedSetting.scala
```diff
@@ -0,0 +1,116 @@
...
+      val e2 = SettingWrittenTwice(name, currentStackTrace, writtenTwiceMessage2(value))
+      LiftRules.guardedSettingViolationFunc.get.apply(e2)
+    }
+
+    if(LiftRules.doneBoot) {
+      val e = SettingWrittenAfterBoot(name, currentStackTrace, writeAfterBootMessage)
+      LiftRules.guardedSettingViolationFunc.get.apply(e)
+    }
+
+    // TODO: Skip if the value is the same?
```
joescii: Should we skip the warning if the value matches what has already been read?

joescii:
I'm about to push a commit to remove the set-twice violation. My only outstanding question is [here](https://github.com/lift/framework/pull/1889#pullrequestreview-56936344).

web/webkit/src/main/scala/net/liftweb/http/LiftRulesGuardedSetting.scala
farmdawgnation: No, I don't think we should. Though it may be the same for some set of inputs it may not be the same for some other set of inputs. It would be my luck that this would happen in development vs production.

joescii: That's a good point. I'll remove the comment.

joescii:
Ship it?

farmdawgnation:
I think so. =)

PR #1903, opened 2017-08-25
Provide context path to comet session reload function

### LINKED ISSUE #1901: Context path not considered in default noCometSessionCmd
**[Mailing List](https://groups.google.com/forum/#!forum/liftweb) thread:** https://groups.google.com/d/msg/liftweb/uFp0jxZVe_I/RsYTSrLoCgAJ

**[Example project](https://app.assembla.com/wiki/show/liftweb/Posting_example_code) link:** https://github.com/VlachJosef/Lift-3.1.0-redirect-issue

When we run into a situation where we expect a comet on a page, but there is no comet on a page, we automatically execute a `noCometSessionCmd` which should force the page to reload _somewhere_ and re-init all of the comets that are needed to run the page.

The default `noCometSessionCmd` doesn't take context path into account, so if you're hosting your Lift app inside an application container and it's mounted somewhere other than `/` - this default command will not work correctly. See the example project above for an example of this.

This is a regression from 2.6, so we'll want to patch Lift 3.0 and Lift 3.1.

farmdawgnation:
closes #1901

This _attempts_ to fix the issue with context paths when comet sessions are lost. For some reason it doesn't appear to actually fix the issue... I'm at the point where I need a second pair of eyes on this. :/


web/webkit/src/main/resources/toserve/lift.js
```diff
@@ -76,8 +76,12 @@
       },
       cometGetTimeout: 140000,
       cometFailureRetryTimeout: 10000,
-      cometOnSessionLost: function() {
-        window.location.href = "/";
+      cometOnSessionLost: function(contextPath) {
+        if (contextPath === null || contextPath === undefined) {
+          window.location.href = "/";
+        } else {
+          window.location.href = contextPath;
```
Shadowfiend: This could really just be simplified to `window.location.href = contextPath || "/"`.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -776,7 +776,7 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
    * lift.cometOnSessionLost reloads the current page by default.
    */
   val noCometSessionCmd = new FactoryMaker[JsCmd](
-    () => JsCmds.Run("lift.cometOnSessionLost()")
+    () => JsCmds.Run(s"lift.cometOnSessionLost('${S.contextPath}')")
```
Shadowfiend: Should we urlencode this guy and escape `'`s?

farmdawgnation:
@Shadowfiend Alright, I think this is ready for another look

farmdawgnation:
@Shadowfiend hate to break it to you but in my testing the urlencoding breaks everything. It looks like the convention is for the context path to include the `/` which, when urlencoded, produces different results. Going to back that particular change out. We can leave in the quote escape though.

Shadowfiend:
That makes sense---in JS, that's the difference between `encodeURI` and `encodeURIComponent`, I guess. It's possible browsers automatically handle that for `window.location` assignments, as well. Since context path probably won't be much more than a path, I think this is fine.

PR #1906, opened 2017-08-29
Finish up the snippet timer implementation

### LINKED ISSUE #1460: Feature which optionally times the performance of snippet evaluation
## Original text:

This feature optionally times the evaluation of snippets as discussed in this thread.

https://groups.google.com/forum/?fromgroups=#!topic/liftweb/yy8gxcjH_bs

I think this is worth writing up a quick wiki as its discoverability is zero and it is quite a useful feature (I used it to track down a performance problem in my app). Where should I be submitting documentation to?

Cheers,

Noel

farmdawgnation:
This finishes up the snippet timer implementation started by @NoelKennedy.

Closes #1460.

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
@@ -852,6 +852,9 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
   @volatile var snippetNamesToSearch: FactoryMaker[String => List[String]] =
       new FactoryMaker(() => (name: String) => name :: Nil) {}
 
+
+  val snippetTimer = new FactoryMaker[SnippetTimer](() => NoOpSnippetTimer){}
+
   /**
    * Implementation for snippetNamesToSearch that looks first in a package named by taking the current template path.
    * For example, suppose the following is configured in Boot:
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
@@ -647,10 +647,10 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
    *
    * The HTML Lift emitted would then look like:
    *
-   * {{{
+   * <pre>
    * <span id="lift-event-js-F827001738725NKMEQNF"
    *       data-lift-removed-attributes="onclick">Do something!</span>
-   * }}}
+   * </pre>
    *
    * This makes it possible to replace the old CSS with with similar
    * matching for the `data-lift-removed-attributes` attribute:
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
@@ -862,7 +862,7 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
    * Since this is a FactoryMaker you can programmatically override it for an individual request
    * or session as you see fit. You can also implement your own timer!
    */
-  val snippetTimer = new FactoryMaker[SnippetTimer](() => NoOpSnippetTimer){}
+  val snippetTimer = new LiftRulesGuardedSetting[Option[FactoryMaker[SnippetTimer]]]("snippetTimer", None)
 
   /**
    * Implementation for snippetNamesToSearch that looks first in a package named by taking the current template path.
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
@@ -853,6 +853,15 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
       new FactoryMaker(() => (name: String) => name :: Nil) {}
 
 
+  /**
+   * Snippet timers are used to time and record the execution time of snippets. We provide
+   * two default implementations for you:
+   *   - NoOpSnippetTimer that does nothing
+   *   - LoggingSnippetTimer that logs snippet times.
+   *
+   * Since this is a FactoryMaker you can programmatically override it for an individual request
+   * or session as you see fit. You can also implement your own timer!
+   */
   val snippetTimer = new FactoryMaker[SnippetTimer](() => NoOpSnippetTimer){}
 
   /**
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
@@ -859,10 +859,28 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
    *   - NoOpSnippetTimer that does nothing
    *   - LoggingSnippetTimer that logs snippet times.
    *
-   * Since this is a FactoryMaker you can programmatically override it for an individual request
+   * To enable snippet timing, invoke `LiftRules.installSnippetTimer`. Once enabled you can use
+   * the snippet timer like a regular `FactoryMaker`. If you only want snippet timing for certain
+   * sessions or requests, invoke `installSnippetTimer` with the `NoOpSnippetTimer` and change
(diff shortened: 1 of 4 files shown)


web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -647,10 +647,10 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
    *
    * The HTML Lift emitted would then look like:
    *
-   * {{{
+   * <pre>
    * <span id="lift-event-js-F827001738725NKMEQNF"
    *       data-lift-removed-attributes="onclick">Do something!</span>
-   * }}}
+   * </pre>
```
Shadowfiend: 👎👎👎👎👎👎👎👎👎👎

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -852,6 +852,18 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
...
+  /**
+   * Snippet timers are used to time and record the execution time of snippets. We provide
+   * two default implementations for you:
+   *   - NoOpSnippetTimer that does nothing
+   *   - LoggingSnippetTimer that logs snippet times.
+   *
+   * Since this is a FactoryMaker you can programmatically override it for an individual request
+   * or session as you see fit. You can also implement your own timer!
+   */
+  val snippetTimer = new FactoryMaker[SnippetTimer](() => NoOpSnippetTimer){}
```
Shadowfiend: Worth noting that FactoryMakers can have significant overhead. Tempted to say this is best done as a var that can perhaps hold an `Option[FactoryMaker]` or something of the sort.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
farmdawgnation: Yeah, I'm not wild about it either, but the file was (and remains) totally unusable in my tooling. I'm honestly not sure it shouldn't be given that the intent of this syntax is to include scala code, not HTML.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
farmdawgnation: Ah, good call.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
Shadowfiend: Is that the case? Many uses of this include the scala REPL prompt, for example, which is not valid scala code. Additionally, the [scaladoc docs](https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html) just say ‘code blocks’, not ‘Scala code blocks’. It would seem surprising to me for scaladoc to have an overly narrow definition of code blocks, but not surprising for tooling to fail to take this flexibility into account, fwiw. Could def be wrong though.

farmdawgnation: Hmmm. Yeah, guess I'm wrong on that then. Whatevs. I'll revert.

farmdawgnation:
@Shadowfiend This is ready for another look. I decided to use one of @joescii's fancy new `LiftRulesGuardedSetting` things. LMKWYT!

farmdawgnation:
I think we have a bit of a problem here.

`FactoryMaker` is inherited from a `LiftRules` through the `Factory` trait. It's attached to the _instance_ of `LiftRules` and not the singleton. I tried to do something like this in order to set the snippetTimer:

```scala
    if (! Props.productionMode) {
      LiftRules.snippetTimer.set(Some(
        new FactoryMaker[SnippetTimer](LoggingSnippetTimer){}
      ))
    }
```

No luck because `FactoryMaker` isn't visible from `Boot`. I tried adding the `Factory` trait into `Boot` but that has a subtly different type and resulted in:

```
[error] /Users/matt/Projects/farmdawgnation/judicial-manager/src/main/scala/bootstrap/liftweb/Boot.scala:62: type mismatch;
[error]  found   : <error>
[error]  required: _1.FactoryMaker[net.liftweb.http.SnippetTimer] where val _1: net.liftweb.http.LiftRules
[error]         new realInstance.FactoryMaker[SnippetTimer](LoggingSnippetTimer){}
[error]         ^
[error] two errors found
```

So, it looks like we either need to revert this to an `Option[SnippetTimer]` or use a `FactoryMaker`...

Shadowfiend:
Could this be `new LiftRules.FactoryMaker[SnippetTimer](LoggingSnippetTimer)`?

farmdawgnation:
No, because the `FactoryMaker` class is tied to the _instance_ of the `LiftRules` class, not the type of `LiftRules`. To the JVM they look like two different types.

Shadowfiend:
Oh me lordy. That's nasty.

Perhaps a helper on `Factory` would be handy…

```scala
  def buildFactoryMaker[T : Manifest](default: Vendor[T]) = {
    new FactoryMaker[T](default) {}
  }
```

Then you could do `LiftRules.buildFactoryMaker` here. That's all I can really think of though… In Java, it seems like we'd be able to do something like `LiftRules.new FactoryMaker[String](...) {}`, but that doesn't seem to be legal Scala...

farmdawgnation:
I don't think that would work either.

I'd have to do something like `LiftRules.realInstance.buildFactoryMaker` to be legal unless I'm mis-remembering. I think ultimately, though, the right answer is to change how we declare `FactoryMaker` to make it a proper first-class citizen. I'm just debating about whether or not this would be a breaking change. I'm inclined to think that it wouldn't. It would break binary compatibility, but not API compatibility unless I'm missing something.

Shadowfiend:
Hmm… I believe the current representation is tied into how it tracks its internal data. Changing that seems like it could be a lot of work in a potentially-dangerous and mission-critical corner of the code base for relatively little gain.

`LiftRules.realInstance` is already abstracted away by the implicit conversion from `LiftRulesMocker` to `LiftRules`, which is how we can access everything else on the `LiftRules` class on the `LiftRules` singleton instead.

Whiiiiich actually revealed to me the situation here.

```scala
  val currentRules = LiftRules.realInstance
  val maker = new currentRules.FactoryMaker[String](...) {}
```

I actually think this weirdness can be contained. Rather than all this dancing about, we just provide an `installSnippetTimer` function:

```scala
  def installSnippetTimer(default: SnippetTimer): Unit = {
    val factoryMaker = new FactoryMaker[String](default) {}
    snippetTimer.set(Some(factoryMaker))
  }
```

This subsumes all the instantiation stuff, and once the timer is installed you can use the `FactoryMaker` as usual by accessing `LiftRules.snippetTimer.value.foreach(_.default.set)`, .request.set, etc. The setting + Option + FactoryMaker combo is the real annoyance here, but I don't know if it's avoidable.

farmdawgnation:
That would work, but I'm already not wild about the number of different shapes for setting things in `LiftRules`. That said, overhauling that is way outside of scope for a minor point release.

farmdawgnation:
I'll take a look at implementing that soonish, with an eye toward hopefully getting it in for M2.

farmdawgnation:
Looks like this works!

```
20:46:39.293 Snippet CompDashboard.render evaluation took 8 Milliseconds
20:46:39.296 Snippet embed evaluation took 2 Milliseconds
20:46:39.298 Snippet CompMeta.name evaluation took 1 Milliseconds
20:46:39.299 Snippet CompMeta.status evaluation took 1 Milliseconds
20:46:39.302 Snippet Msgs evaluation took 2 Milliseconds
20:46:39.305 Snippet surround evaluation took 23 Milliseconds
20:46:39.306 Snippet Menu.title evaluation took 0 Milliseconds
20:46:39.307 Snippet CurrentUser.rootLink evaluation took 1 Milliseconds
20:46:39.308 Snippet CurrentUser.userName evaluation took 1 Milliseconds
20:46:39.308 Snippet CurrentUser.logoutLink evaluation took 0 Milliseconds
```

PR #1908, opened 2017-08-29
Support limiting MetaProtoExtendedSession cookies to context path

### LINKED ISSUE #1758: Set default cookie path for MetaProtoExtendedSession to contextPath
As suggested in https://groups.google.com/d/msg/liftweb/ZWmPdpIG1JA/dfAzbXiTBgAJ the cookie path for `MetaProtoExtendedSession` should be configurable and/or default to contextPath.

farmdawgnation:
This PR add support for limiting the extended sessions provided by `MetaProtoExtendedSession` to the current context path instead of `/`.  To use this, add `with ContextPathExtendedCookie` to your `MetaProtoExtendedSession`.

Closes #1758.

### DIFF
--- a/persistence/mapper/src/main/scala/net/liftweb/mapper/ProtoExtendedSession.scala
+++ b/persistence/mapper/src/main/scala/net/liftweb/mapper/ProtoExtendedSession.scala
@@ -143,6 +143,14 @@ KeyedMetaMapper[Long, T] with ProtoSessionCookiePath {
  * to the context path for your application. This is useful if you have
  * multiple applications on a single application server and want to ensure
  * their cookies don't cross-pollinate.
+ *
+ * Example usage:
+ *
+ * {{{
+ * case class AppExtendedSession extends ProtoExtendedSession[AppExtendedSession]
+ * object MetaAppExtendedSession extends MetaProtoExtendedSession[AppExtendedSession]
+ *   with ContextPathExtendedCookie
+ * }}}
  */
 trait ContextPathExtendedCookie extends ProtoSessionCookiePath {
   override def sessionCookiePath = S.contextPath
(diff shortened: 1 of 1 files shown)


persistence/mapper/src/main/scala/net/liftweb/mapper/ProtoExtendedSession.scala
```diff
@@ -59,8 +59,12 @@ trait UserIdAsString {
   def userIdAsString: String
 }
 
+trait ProtoSessionCookiePath {
+  def sessionCookiePath: String = "/"
```
Shadowfiend: Worth adding a quick scaladoc here? Also, does this need to live in its own trait?

farmdawgnation: It does, _or_ `ContextPathExtendedCookie` would need to extend `MetaProtoExtendedSession` which semantically seemed _more_ confusing. If you disagree then it's probably worth discussing.

Can add a scaladoc here though.

Shadowfiend: Biggest annoyance with that is that `ContextPathExtendedCookie` would need to be generic. I think this works.

persistence/mapper/src/main/scala/net/liftweb/mapper/ProtoExtendedSession.scala
```diff
@@ -59,6 +59,13 @@ trait UserIdAsString {
   def userIdAsString: String
 }
 
+/**
+ * The root trait for defining the session cookie path for extended sessions
+ * that defines the default session cookie path: "/".
+ *
+ * Extend this trait then mix it into an extended session singleton to change
```
Shadowfiend: Do you ever need to extend this trait explicitly? Can you not just change this directly in your extended session singleton?

farmdawgnation: You could directly make the change, so long as you only wanted to make the change to a single instance. Which, I suppose given how these are used, that's all you'd do. I'll remove the remark.

farmdawgnation:
If I don't hear any further feedback by Monday afternoon, I'll go ahead and merge this.

PR #1910, opened 2017-08-31
Improve LAPinger documentation for clarity

### LINKED ISSUE #1778: LAPinger ScalaDoc is badly worded
Discussed in this thread:

https://groups.google.com/forum/#!topic/liftweb/elN2uamESOc

I would recommend wording something more like:

The LAPinger object schedules an actor to be pinged "delay" miliseconds in the future. The schedule method return a ScheduledFuture object which can be cancelled if necessary.

farmdawgnation:
Closes #1778.

### DIFF
--- a/core/actor/src/main/scala/net/liftweb/actor/LAPinger.scala
+++ b/core/actor/src/main/scala/net/liftweb/actor/LAPinger.scala
@@ -34,8 +34,8 @@ object ThreadPoolRules {
 }
 
 /**
- * The ActorPing object schedules an actor to be ping-ed with a given message at specific intervals.
- * The schedule methods return a ScheduledFuture object which can be cancelled if necessary
+ * The LAPinger object schedules a LiftActor to be pinged "delay" miliseconds in the future.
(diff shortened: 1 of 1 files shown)


core/actor/src/main/scala/net/liftweb/actor/LAPinger.scala
```diff
@@ -34,8 +34,8 @@ object ThreadPoolRules {
 }
 
 /**
- * The ActorPing object schedules an actor to be ping-ed with a given message at specific intervals.
- * The schedule methods return a ScheduledFuture object which can be cancelled if necessary
+ * The LAPinger object schedules a LiftActor to be pinged "delay" miliseconds in the future.
```
Shadowfiend: The original phrasing seems clearer to me here (class names aside). `delay` doesn't really have a meaning at the level of the singleton that I can see.

farmdawgnation:
@Shadowfiend Check the latest commit here and let me know what you think.

PR #1911, opened 2017-08-31
Consider LiftRules.cometCreation while building comets

### LINKED ISSUE #1902: LiftRules.cometCreation is not considered when building comets
**[Mailing List](https://groups.google.com/forum/#!forum/liftweb) thread:** https://groups.google.com/d/msg/liftweb/uFp0jxZVe_I/_LFJ51hwCgAJ

The `LiftRules.cometCreation` value is not considered when building comets as of Lift 3.0. This means that if you try to leverage this value for comet creation, it'll just stop working once you've upgraded from 2.6.

This is a regression from Lift 2.6, so we should also patch Lift 3.0 and Lift 3.1.

Assigning to @Shadowfiend for now since he has some familiarity with the issue and seemed willing to tackle it from the exchange on the ML.

farmdawgnation:
This fixes a bug that appeared in Lift 3 where we stopped considering cometCreation while creating
comets from creation info.

Decided to take my own stab at this one @Shadowfiend - let me know how I did!

Do still need to test it throughly though.

Closes #1902.

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
@@ -2487,8 +2487,9 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
   /**
    * As with `findOrBuildComet[T]`, but specify the type as a `String`. If the
    * comet doesn't already exist, the comet type is first looked up via
-   * `LiftRules.cometCreationFactory`, and then as a class name in the comet
-   * packages designated by `LiftRules.buildPackage("comet")`.
+   * `LiftRules.cometCreationFactory`, then `LiftRules.cometCreation`, and
+   * finally  as a class name in the comet packages designated by
+   * `LiftRules.buildPackage("comet")`.
    */
   private[http] def findOrCreateComet(
     cometType: String,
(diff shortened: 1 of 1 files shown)


web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
```diff
@@ -2568,12 +2569,14 @@ class LiftSession(private[http] val _contextPath: String, val underlyingId: Stri
...
-  // attempting to use LiftRules.cometCreationFactory and then building it by
+  // attempting to use `LiftRules.cometCreationFactory` and then attempting to
+  // find a match in `LiftRules.cometCreation`. Failing those, this will build it by
   // class name. Return a descriptive Failure if it's all gone sideways.
   //
   // Runs some base setup tasks before returning the comet.
   private def buildCometByCreationInfo(creationInfo: CometCreationInfo): Box[LiftCometActor] = {
-    LiftRules.cometCreationFactory.vend.apply(creationInfo) or {
+    LiftRules.cometCreationFactory.vend.apply(creationInfo) or
+    LiftRules.cometCreation.toList.find(_.isDefinedAt(creationInfo)).map(_.apply(creationInfo)) or {
```
Shadowfiend: Wonder if we could use `NamedPF.applyBox(creationInfo, LiftRules.cometCreation.toList)` here.

farmdawgnation:
@Shadowfiend ready for another round here

Shadowfiend:
Sweet. That seem like a good change to you as well?

Also: thoughts on potentially introducing a test here?

farmdawgnation:
> Sweet. That seem like a good change to you as well?

Yeah, it reads a lot smoother. It might be a hair more magical but nothing that scares me too too much.

> Also: thoughts on potentially introducing a test here?

How would you feel about filing a follow up ticket for that? This has already cost us a good bit of time and I'd like to make the M2 deadline with the fix.

Shadowfiend:
Up to you on the ticket, I'm down to leave this as-is. I'm responsible for there being no test coverage here anyway heh <_<

PR #1920, opened 2017-09-20
optimization of StringHelpers.randomString

andreak:
- Optimize randomString by calling _random.nextBytes for only the bytes we need.
  We use 6 bits per byte and store the 2 leftover bits in a byte (Int actually).
  After 3 iterations we have (3*2) enough bits for a new random char so we use
  these instead of consuming a new random-byte. So for every 4-th iteration we
  use these accumulated random-bits.

Some non-scientific testing shows 30-35% improvement in running-time.

### DIFF
--- a/core/util/src/main/scala/net/liftweb/util/StringHelpers.scala
+++ b/core/util/src/main/scala/net/liftweb/util/StringHelpers.scala
@@ -175,6 +175,7 @@ trait StringHelpers {
     * @return the generated string
     */
   def randomString(size: Int): String = {
+    if (size < 1) return ""
     val charArray = new Array[Char](size)
     val numRandomBits = 6
     val bytesNeeded = math.ceil((numRandomBits * size) / 8.0).toInt
--- a/core/util/src/main/scala/net/liftweb/util/StringHelpers.scala
+++ b/core/util/src/main/scala/net/liftweb/util/StringHelpers.scala
@@ -170,28 +170,42 @@ trait StringHelpers {
   def clean(in : String) =  if (in == null) "" else in.replaceAll("[^a-zA-Z0-9_]", "")
 
   /**
-   * Create a random string of a given size.  5 bits of randomness per character
-   * @param size size of the string to create. Must be a positive integer.
-   * @return the generated string
-   */
+    * Create a random string of a given size.  6 bits of randomness per character
+    * @param size size of the string to create. Must be a positive integer.
+    * @return the generated string
+    */
   def randomString(size: Int): String = {
-    def addChar(pos: Int, lastRand: Int, sb: GoodSB): GoodSB = {
-      if (pos >= size) sb
-      else {
-        val randNum = if ((pos % 6) == 0) {
-          _random.synchronized(_random.nextInt)
-        } else {
-          lastRand
-        }
-
-        sb.append((randNum & 0x1f) match {
-          case n if n < 26 => ('A' + n).toChar
-          case n => ('0' + (n - 26)).toChar
-        })
-        addChar(pos + 1, randNum >> 5, sb)
+    val charArray = new Array[Char](size)
+    val numRandomBits = 6
+    val bytesNeeded = math.ceil((numRandomBits * size) / 8.0).toInt
+    val bytesArray = new Array[Byte](bytesNeeded)
+    _random.nextBytes(bytesArray)
+    var carry: Int = 0
+    var carryCount: Int = 0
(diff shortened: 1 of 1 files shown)


core/util/src/main/scala/net/liftweb/util/StringHelpers.scala
```diff
@@ -170,28 +170,43 @@ trait StringHelpers {
...
+        val charFromCarry = carry.toByte
+        carry = 0
+        carryCount = 0
+        charFromCarry
       }
-    }
-    addChar(0, 0, new GoodSB(size)).toString
+      charArray(idx) = byteToUse & 0x3f match { // Zero all but 6 first bits
+        case n if n < 26 => ('A' + n).toChar
+        case n => ('0' + (n % 10)).toChar // Ensure 0 < n < 10
```
Shadowfiend: I'm concerned that we're doing a little too much trickery here. Doesn't the mod actually lose entropy by fitting the random value into a very small range? See for example https://crypto.stackexchange.com/a/22769 . The old approach instead aligns itself to a binary range, ensuring we use the full provided entropy.

One call to `nextFuncName` *should* really only result in two `_random.nextInt`, since `nextFuncName` does a `randomString(6)`. Is most of the gain here from this change in approach or from the loss of synchronization? And if it's the approach, is it the structure of the code or the approach itself that has resulted in the difference? This code also happens to be a bit harder to grok IMO, so if we can stick with the old approach sans synchronization, I wouldn't be the least bit annoyed.

Also worth noting: the `String` constructor can directly take a byte array, provided you furnish a `Charset` as well (you don't have to, but you're at the mercy of the platform if you don't). Not sure if that's faster than converting to `Char` along the way.

andreak: If not mod(10), how should we generate the numers 0-9, ensuring the random-value is within range?
Also note that the old code did _not_ generate the numbers 6-9, because the highest value 5 bits can represent is 31, meaning the highest number generated was ('0' + (31 - 26)).toChar = 5.

I tried replacing the char-array with a byte-array:
```
		val charArray = new Array[Byte](size)
...
			charArray(idx) = byteToUse & 0x3f match { // Zero all but 6 first bits
				case n if n < 26 => (65 + n).toByte
				case n => (48 + (n % 10)).toByte // Ensure 0 < n < 10
			}

...
		new String(charArray) // all chars are [A-Z0-9] so no point providing Charset
```
But it was actually ~8% slower.

andreak: strace'ing the process on Linux shows that sun.security.provider.NativePRNG reads 32 random bytes at a time from /dev/urandom, regardless of how many bytes you request. Subsequent calls to nextInt/nextBytes only reads more when it needs to.

farmdawgnation:
@Shadowfiend Ping! the last responses look to be from @andreak so I think the ball is in your court here.

core/util/src/main/scala/net/liftweb/util/StringHelpers.scala
Shadowfiend: Not generating the numbers 6-9 isn't relevant if we're using all the available randomness. 5 bits of randomness per character = `[A-Z0-5]`. 6 bits of randomness would be expected to result in 64 possible characters, but that's not happening. I don't think we can say that we're generating 6 bits of randomness if we restrict the range of the numbers, that's really my only concern. I think we can probably say we're generating >5 bits of randomness? And the lost remaining randomness may be a problem.

Fundamentally I'm worried for two reasons:
 - Secure randomness is not easy.
 - This randomness sits at the root of the security of the identifiers we generate, which sit at the root of the security of a Lift application; if we've sacrificed randomness by using more bits because we can't fully utilize those bits, we've actually worsened security for performance, which I don't think is a reasonable tradeoff in such a core function.

My understanding of secure randomness and how mod affects randomness is actually not something I'm *super*-confident in, so if you're comfortable with the theoretical underpinnings and tell me this doesn't produce a less secure system, I'm in.

andreak: Agreed, this optimization seems not worth it, let's close this issue and move one. Thanks for taking time looking at it.

Shadowfiend: If we got a measurable boost in parallel invocation from dropping `synchronized`, I would be totally down to include just that.

PR #1921, opened 2017-10-03
Session serialization support

joescii:
This PR shuffles a few things around, adds some `@transient` annotations, and adds some default constructors to aid in the success of [lift-cluster](https://github.com/joescii/lift-cluster). Unless I screwed up something, this PR is essentially a refactor to anyone NOT using lift-cluster. 

Relevant (but really long and old) ML thread: https://groups.google.com/forum/#!topic/liftweb/KHjbjev8A0E

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
@@ -199,6 +199,41 @@ object LiftSession {
     }
   }
 
+  private[http] class DataAttrNode(liftSession: LiftSession) {
+    val dataAttributeProcessors = LiftRules.dataAttributeProcessor.toList
+
+    def unapply(in: Node): Option[DataAttributeProcessorAnswer] = {
+      in match {
+        case element: Elem if dataAttributeProcessors.nonEmpty =>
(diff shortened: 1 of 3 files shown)


web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala
```diff
@@ -199,6 +199,41 @@ object LiftSession {
     }
   }
 
+  private[http] class DataAttrNode(liftSession: LiftSession) {
```
farmdawgnation: Just verifying: these were merely promoted to first class objects, right? No other changes?

joescii: Correct. I don't recall the exact issue now, but as they were defined they would not deserialize cleanly. Simply making them a class rather than an embedded object allows the deserialized object to re-instantiate as expected.

farmdawgnation: That actually makes sense. It’s possible to run into similar issues with lift-json extraction. Cool.

farmdawgnation:
As there hasn't been any further feedback on this, I'm gonna :shipit:

Issue #1258, opened 2012-04-17
RFE: required HTML5 attribute support on Fields


robmv:
Use BaseField.required_? to add the "required" HTML 5 attribute when serializing to HTML 5

Group discussion thread: https://groups.google.com/forum/?fromgroups#!topic/liftweb/Vtw0Eib8Avs

Shadowfiend:
This should be pretty easy, but will require manually going through the fields that extend `BaseField` to add the `required` attribute as appropriate in their `toForm` implementations.

PR #1925, opened 2017-11-05
#1258 - WIP. Add html 'required' attribute if BaseField.required_? is set to true

Bhashit:
I had some time, and I was sort of waiting for a response on another issue I was working on. And since we are looking for a release in less than two weeks, I thought I would go ahead create a PR for this without asking about it first. If that's inappropriate, please decline this right away.

Otherwise, does this look like the right approach here? Wanted to ask before I went ahead with changing a whole bunch of files.

### DIFF
--- a/core/util/src/main/scala/net/liftweb/util/BaseField.scala
+++ b/core/util/src/main/scala/net/liftweb/util/BaseField.scala
@@ -18,7 +18,8 @@ package net.liftweb
 package util
 
 import common._
-import xml.NodeSeq
+
+import xml._
 
 /**
  * Defines the association of this reference with a markup tag ID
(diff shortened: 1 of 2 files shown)


core/util/src/main/scala/net/liftweb/util/BaseField.scala
```diff
@@ -113,6 +114,14 @@ trait SettableField extends ReadableField with SettableValueHolder {
    */
   def toForm: Box[NodeSeq]
 
+  protected def html5RequiredAttribute: MetaData = {
```
Shadowfiend: Only thing I would say is that, since making something `protected` makes it part of public API, we may want to consider making this a little more generic. Perhaps `additionalHtmlAttributes`?

Bhashit: That sounds useful. But yeah, good catch. I'll try and see if there's a way I can avoid adding to the public API.

Bhashit: Unless you want to add that method to the public API, how about making it `private [liftweb]` in some companion object. For ex. as a method in `SettableField` where the `required` attribute is defined.

Shadowfiend: I think starting as `private[liftweb]` is fine. If we want it to be public API we can adjust later. Can't take that back :)

farmdawgnation:
Hey there, I'm closing PR as orphaned since it's been more than a year since it's had any updates. Feel free to reopen when you have time to update it!

PR #1933, opened 2018-01-16
Issue#1919: Asset root path should be configurable

heharkon:
Added 'assetPath' variable to LiftRules, with default value '/assets'. Rest of the path structure is still convention based, not configurable.

This will address the issue where the Lift template projects are storing the lazy loading gif under assets folder and the framework library expects it to reside at the root level.

On the other hand, also if the default place for the assets is in the way of the application for a reason or another, it is now easier to move somewhere else.

Added also my name to the contributors.md file, as instructed in the documentation.

**[Mailing List](https://groups.google.com/forum/#!topic/liftweb/-2Mh4D2L-cM) thread**:

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
@@ -1335,13 +1335,20 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
   }
 
   /**
-    * The root path for assets folder. For example to store
-    * asset files under 'myassets' folder, use value '/myassets/'. Assets
-    * under this configurable path follow conventions, for example
-    * the default lazy loading spinner ('ajax-loader.gif') is under
-    * 'images' folder.
+    * The root path for the assets folder.
+    *
+    * This applies also to the URL domain, appended after the deployed context path.
+    * In app developer's work area this folder resides under the 'webapp' folder.
+    *
+    * For example to store asset files under 'myassets' folder,
+    * use value '/myassets/'. Assets under this configurable path follow
+    * conventions, for example the default lazy loading spinner ('ajax-loader.gif')
+    * is under 'images' folder.
+    *
+    * Thus the URL to the assets folder would be:
+    * 'http://<domain name>/<context path>/myassets/'
     */
-  @volatile var assetPath: String = "/"
+  @volatile var assetRootPath: String = "/"
 
   /**
    * Contains the URI path under which all built-in Lift-handled requests are
--- a/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
@@ -1335,6 +1335,13 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
   }
 
   /**
+    * The root path for assets folder. Assets under this configurable path
+    * follow conventions, for example the default lazy loading spinner
+    * ('ajax-loader.gif') is under 'images' folder.
+    */
+  @volatile var assetPath: String = "/assets"
+
+  /**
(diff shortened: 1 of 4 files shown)


web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -1334,6 +1334,13 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
     }
   }
 
+  /**
+    * The root path for assets folder. Assets under this configurable path
+    * follow conventions, for example the default lazy loading spinner
+    * ('ajax-loader.gif') is under 'images' folder.
+    */
+  @volatile var assetPath: String = "/assets"
```
farmdawgnation: I'm kind of torn over whether or not this default represents a breaking change or not. I'm inclined to say that we should default this to `/` until Lift 4.0.

Open to other opinions, though.

heharkon: You're right, it would be a nasty surprise when upgrading to a new version of lift. I'll update the PR soon with the default path set to /. Thanks.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
```diff
@@ -1334,6 +1334,15 @@ class LiftRules() extends Factory with FormVendor with LazyLoggable {
...
   }
 
+  /**
+    * The root path for assets folder. For example to store
+    * asset files under 'myassets' folder, use value '/myassets/'. Assets
+    * under this configurable path follow conventions, for example
+    * the default lazy loading spinner ('ajax-loader.gif') is under
+    * 'images' folder.
+    */
+  @volatile var assetPath: String = "/"
```
eltimn: Would it make sense to have this default to `""` instead of removing the `/` from the links?

heharkon: @eltimn I was thinking that option too, but eventually leaned towards this way. I was thinking that slash would make it clear that it is about root path, and also when with a folder value it would need to be there. 

But, I'm happy with either way, just let me know. :)

eltimn: I'm fine either way, really. I just wanted to bring it up and see if anyone else had an opinion.

heharkon: There is also a third possible approach, I wrote to the mailing list about it:
https://groups.google.com/d/msg/liftweb/bf07dlUfWpc/cUG6KXeiAQAJ

farmdawgnation:
I didn't merge this because it didn't seem like we've landed on an agreement yet. What are we thinking on how this should go?

farmdawgnation:
Oh sorry, I see the link about the ML. That conversation i still ongoing. We'll wait to see what happens.

web/webkit/src/main/scala/net/liftweb/http/LiftRules.scala
Shadowfiend: I'm down for the current approach. One thought though: should this be `assetRoot` or `assetPathRoot`?

Also, the scaladocs don't make it immediately clear that this is something that is relative to the domain root; that is, that this goes directly into HTML references and therefore `/` means something other than “project root” or any number of other possibilities. It's also worth noting that this will in turn be rewritten to take context path into account, if that is set, for maximum fun.

heharkon: Yes, maybe it would be better to make those things more clear, the naming and also scaladoc. I'll try to find time to change them today after work.

joescii:
This would be _super_ handy for splitting up a Lift app into modules behind an nginx proxy!

PR #1935, opened 2018-01-18
Avoid extra array creation when creating a list of a single item

chriswebster:
Mailing List thread:
https://groups.google.com/forum/#!topic/liftweb/jD3OG9jD_hs

The list creation in findDirectByName was using variable argument constructor which was performing unnecessary array allocations as part of the read case class flow. This change creates the single item list directly bypassing the extra array allocation and optimizing performance.

The benchmark results were as follows:

before change
3.2.0-RC1
[info] LiftJsonBenchmark.readCaseClass thrpt 10 443843.164 ± 5996.863 ops/s

after change
[info] LiftJsonBenchmark.readCaseClass thrpt 10 584527.678 ± 11315.888 ops/s


core/json/src/main/scala/net/liftweb/json/JsonAST.scala
```diff
@@ -118,7 +118,7 @@ object JsonAST {
      * nothing is found, you'll get a `JNothing`.
      */
     def \(nameToFind: String): JValue = {
-      findDirectByName(List(this), nameToFind) match {
+      findDirectByName(this :: Nil, nameToFind) match {
```
farmdawgnation: I'm debating about whether or not this is an obscure enough change to document somehow. I'll keep thinking on it.

Shadowfiend: I think it's worth leaving a note here, in the style of “Use :: instead of List() to avoid an extra array allocation.”

chriswebster: done

PR #1936, opened 2018-01-19
Use single value append to avoid the overhead of variable arguments

chriswebster:
https://groups.google.com/forum/#!topic/liftweb/jD3OG9jD_hs

The newValue method in JsonParser was using a variable argument method when adding a single item to a ListBuffer which triggered extra array allocations. This change uses a single item method bypassing the extra array allocation and optimizing performance.

before change
3.2.0-RC1
[info] LiftJsonBenchmark.readCaseClass thrpt 10 443843.164 ± 5996.863 ops/s

After patch
[info] LiftJsonBenchmark.readCaseClass  thrpt   10  551319.586 ± 6361.040  ops/s


Shadowfiend:
:bow: really appreciate these tweaks you keep contributing @chriswebster !

PR #1939, opened 2018-02-04
ajk-service-request-timer

andreak:
Refactor timing of servicing a request out to ServiceRequestTimer.
The default-implementation, StandardServiceTimer, gives the same
logging as the old parameter logServiceRequestTiming=true

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/provider/HTTPProvider.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/provider/HTTPProvider.scala
@@ -132,6 +132,9 @@ trait HTTPProvider {
   }
 
   private def postBoot {
+    if (!LiftRules.logServiceRequestTiming) {
+      LiftRules.installServiceRequestTimer(NoOpServiceTimer)
+    }
     try {
       ResourceBundle getBundle (LiftRules.liftCoreResourceName)
     } catch {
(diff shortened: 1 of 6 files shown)


andreak:
Note that the codacy-report complaining about
```
Usage of get on optional type.
LiftRules.serviceRequestTimer.get.vend.logTime(req, resp)(doService)
```
https://app.codacy.com/app/Lift/framework/file/14407628970/issues/source?bid=2087836&fileBranchId=6459480#l162

is wrong, as this isn't an Option but a ValueHolder.

web/webkit/src/main/scala/net/liftweb/http/provider/HTTPProvider.scala
```diff
@@ -132,6 +132,9 @@ trait HTTPProvider {
   }
 
   private def postBoot {
+    if (!LiftRules.logServiceRequestTiming) {
+      LiftRules.installServiceRequestTimer(NoOpServiceTimer)
+    }
```
farmdawgnation: I'm not sure I quite understand this line. If request timing is off, we install the no-op timer in the new place?

andreak: What I'm proposing is to remove ```logServiceRequestTiming``` and only use ```ServiceRequestTimer```, but we cannot make such a breaking change (removing```logServiceRequestTiming```) in 3.x can we? So to not break existing apps the line in ```postBoot``` installs the ```NoOpServiceTimer``` implementation which will give the same behavior as today's ```logServiceRequestTiming=false```. Currently ```logServiceRequestTiming``` defaults to ```true```, which corresponds to having ```StandardServiceTimer``` as default, which is the default in this PR.

PR #1940, opened 2018-02-09
Changing a method of lift actor.

zhongsigang:
An actor method was implemented incorrectly.

### DIFF
--- a/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
+++ b/core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
@@ -171,7 +171,7 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler, context: Box[LAFutur
   /**
    * Has the future been aborted
    */
-  def aborted_? = synchronized {satisfied}
+  def aborted_? = synchronized {aborted}
 
   /**
    * Java-friendly alias for completed_?.
--- a/core/actor/src/main/scala/net/liftweb/actor/LiftActor.scala
+++ b/core/actor/src/main/scala/net/liftweb/actor/LiftActor.scala
@@ -130,7 +130,6 @@ trait SpecializedLiftActor[T] extends SimpleActor[T]  {
     */
 
     def remove() {
-      val newPrev = prev
       prev.next = next
       next.prev = prev
     }


core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
```diff
@@ -171,7 +171,7 @@ class LAFuture[T](val scheduler: LAScheduler = LAScheduler, context: Box[LAFutur
   /**
    * Has the future been aborted
    */
-  def aborted_? = synchronized {satisfied}
+  def aborted_? = synchronized {aborted}
```
farmdawgnation: Oh my. This.... is a pretty bad bug. 🤔 

@lift/committers Thoughts on incorporating this fix back to 3.0 since this certainly isn't the _intended_ behavior?

farmdawgnation: Oh wait, we... we never actually flip `aborted` to true as far as I can see, either. It looks like we should do that in `abort()` yeah?

farmdawgnation:
Jut realized the committers group includes a lot of folks who are inactive 🤔. Sorry folks, was in my github flow and didn't think about it. May be worth having a committers emeritus group or something for folk that are inactive.

core/actor/src/main/scala/net/liftweb/actor/LAFuture.scala
Shadowfiend: We do, in the `fail` method, which `abort` also falls.

Shadowfiend: So, re: backporting this, the problem is it really changes a fundamental behavior… By literally flipping the previous behavior <_<

But, the existing behavior is *so wrong* that it might still be worth it? This is a tough call.

joescii: Yeah I'm over here watching and eating popcorn cuz I ain't got a clue.

Shadowfiend: Backports are generally

Shadowfiend: Per our support doc, ‘API-incompatible changes will only ship as a part of a major release’… This… is pretty API-incompatible…

So perhaps the big question isn't “do we backport this to Lift 3.0” but… Do we merge it into Lift 3.3 at all?

My feeling is yes to both… But it's a pretty fuzzy feeling.

farmdawgnation: I'm leaning toward "merging for Lift 3.3, not backporting, and including BIG BOLD TEXT in the release notes for 3.3" - mostly because this never worked as intended. I'd also love a test to make sure we don't do something foolish like this in the future. :(

fmpwizard: +1 not backporting it

Shadowfiend: :thumbsup: I'm down with that.

Re: something foolish like this, thus code review :)

zhongsigang:
Unfortunately, I developed it on windows, and it's not easy to build a test
environment.

Thanks.

2018-03-04 1:21 GMT+08:00 Matt Farmer <notifications@github.com>:

PR #1942, opened 2018-03-01
Upgraded jBCrypt to v0.4

### LINKED ISSUE #1941: Upgrade jBcrypt to v0.4
**[Mailing List  thread](https://groups.google.com/d/msg/liftweb/EYarSCv_rro/eB-RBZoiAwAJ)**

jBCrypt had a vulnerability that was fixed in v0.4, but Lift is using a copied version of v0.3m.

eltimn:
Fixes #1941 

I did confirm that the copy we had was the same as the previously published 0.3 version.

I didn't test this against any apps as I'm not using either of the fields anywhere. However, this jar was published 3 years ago, so I think it's probably safe to use.

Lastly, we should probably publish a 2.6.4 version, but I don't see a branch on Github for 2.6 anywhere and wasn't sure how to go about backporting.


Shadowfiend:
No no... He's got a point...

Shadowfiend:
Added #1945 to put the public API back in as a deprecated pass-through.

Shadowfiend:
:100: merging.

We need to coordinate point releases of our various branches, as well.

Shadowfiend:
:100: merging.

We need to coordinate point releases of our various branches, as well.

PR #1943, opened 2018-03-01
Upgrade mongo-java-driver to v3.6.3

eltimn:


eltimn:
Bug fixes plus compatibility with server v3.6: https://docs.mongodb.com/ecosystem/drivers/driver-compatibility-reference/#java-driver-compatibility

I just realized they have newer jars. I may update the dependencies to use them instead of the "uber" jar we're using.

PR #1944, opened 2018-03-01
Code gardening

eltimn:
* Updated build.sbt to use newer syntax
* Use JavaConverters instead of JavaConversions in mongodb and mongodb-record modules


eltimn:
Yes, one small step :)

PR #1945, opened 2018-03-03
Bridges of BCrypt: Reintroduce pass-through deprecated net.liftweb.util.BCrypt

Shadowfiend:
@eltimn had a moment to go ahead and do this to hopefully save you the time :)

Merges into #1942 .

### DIFF
--- a/core/util/src/main/java/net/liftweb/util/BCrypt.java
+++ b/core/util/src/main/java/net/liftweb/util/BCrypt.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2007-2011 WorldWide Conferencing, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


core/util/src/main/java/net/liftweb/util/BCrypt.java
```diff
@@ -0,0 +1,87 @@
...
+ * This class is a passthrough to org.mindrot.jbcrypt.BCrypt, provided for
+ * backwards compatibility as we kept a copy in the Lift codebase before it was
+ * available for package dependencies. Prefer using org.mindrot.jbcrypt.BCrypt.
+ */
+@Deprecated
+public class BCrypt {
+    /**
+     * @see org.mindrot.jbcrypt.BCrypt#hashpw(String,String)
+     */
+    @Deprecated
```
farmdawgnation: A message explaining the deprecation would probably be a good idea for all of these.

Shadowfiend: 62.3% sure the Java deprecated annotation supports neither a message nor a version.

farmdawgnation: Oh right this is a java file. Sigh.

arigoldx: ahHA!

_this_ is why i'm seeing

> method hashpw in class BCrypt is deprecated: see corresponding Javadoc for more information.

:facepalm:

made me think that `hashpw` was deprecated up the chain. 😅

PR #1950, opened 2018-06-03
Various bits of build file Gardening

farmdawgnation:
This PR mostly consists of version bumps in our dependencies in anticipation of the Lift 3.3.0-RC1 release in a few weeks. There are a few minor, unrelated changes as well.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -36,7 +36,7 @@ object Dependencies {
   lazy val jbcrypt                = "org.mindrot"                % "jbcrypt"            % "0.4"
   lazy val joda_time              = "joda-time"                  % "joda-time"          % "2.10"
   lazy val joda_convert           = "org.joda"                   % "joda-convert"       % "2.0.2"
-  lazy val htmlparser             = "nu.validator.htmlparser"    % "htmlparser"         % "1.4"
+  lazy val htmlparser             = "nu.validator"               % "htmlparser"         % "1.4.11"
   lazy val mongo_java_driver      = "org.mongodb"                % "mongodb-driver"     % "3.6.3"
   lazy val mongo_java_driver_async  = "org.mongodb"              % "mongodb-driver-async" % "3.6.3"
   lazy val paranamer              = "com.thoughtworks.paranamer" % "paranamer"          % "2.8"
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -35,7 +35,7 @@ object Dependencies {
   lazy val javamail               = "javax.mail"                 % "mail"               % "1.4.7"
   lazy val jbcrypt                = "org.mindrot"                % "jbcrypt"            % "0.4"
   lazy val joda_time              = "joda-time"                  % "joda-time"          % "2.10"
-  lazy val joda_convert           = "org.joda"                   % "joda-convert"       % "1.8.1"
+  lazy val joda_convert           = "org.joda"                   % "joda-convert"       % "2.0.2"
   lazy val htmlparser             = "nu.validator.htmlparser"    % "htmlparser"         % "1.4"
   lazy val mongo_java_driver      = "org.mongodb"                % "mongodb-driver"     % "3.6.3"
   lazy val mongo_java_driver_async  = "org.mongodb"              % "mongodb-driver-async" % "3.6.3"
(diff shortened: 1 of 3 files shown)


project/Dependencies.scala
```diff
@@ -25,30 +25,30 @@ object Dependencies {
...
   lazy val commons_httpclient     = "commons-httpclient"         % "commons-httpclient" % "3.1"
   lazy val javamail               = "javax.mail"                 % "mail"               % "1.4.7"
   lazy val jbcrypt                = "org.mindrot"                % "jbcrypt"            % "0.4"
-  lazy val joda_time              = "joda-time"                  % "joda-time"          % "2.9.2"
-  lazy val joda_convert           = "org.joda"                   % "joda-convert"       % "1.8.1"
-  lazy val htmlparser             = "nu.validator.htmlparser"    % "htmlparser"         % "1.4"
-  lazy val mongo_java_driver      = "org.mongodb"                % "mongodb-driver"     % "3.6.3"
-  lazy val mongo_java_driver_async  = "org.mongodb"              % "mongodb-driver-async" % "3.6.3"
+  lazy val joda_time              = "joda-time"                  % "joda-time"          % "2.10"
+  lazy val joda_convert           = "org.joda"                   % "joda-convert"       % "2.0.2"
```
Shadowfiend: Looks like a major version bump. Any concerns around that?

farmdawgnation: Hmm, well the only thing this touches is the json-ext package, so none major from the framework's point of view. I guess in theory any of these could break transitive binary/API compatibility, but given how poorly developers tend to enforce semver it may be impossible to preserve that with _any_ dependency bump. :/

farmdawgnation: To be clear, certainly worth investigating their change log when I get a chance though.

farmdawgnation:
Looks like the htmlparser bump broke template processing. I'm going to revert that change.

project/Dependencies.scala
```diff
@@ -47,7 +47,7 @@ object Dependencies {
   lazy val squeryl                = "org.squeryl"                % "squeryl"            % "0.9.5-7"   cross CVMappingAll
   lazy val slf4j_api              = "org.slf4j"                  % "slf4j-api"          % slf4jVersion
   lazy val scala_xml              = "org.scala-lang.modules"     %% "scala-xml"         % "1.0.6"
-  lazy val rhino                  = "org.mozilla"                % "rhino"              % "1.7.7.1"
+  lazy val rhino                  = "org.mozilla"                % "rhino"              % "1.7.10"
```
arigoldx, line 50: I know this was a million years ago but any idea for the reason behind it?

No biggie -- I'm just currently in the flail-in-the-dark phase of bumping us to Lift 3.4.3 and am going error by eror 😅 .

farmdawgnation: I think most of this PR was just getting us up to date with regard to our outdated dependencies.

arigoldx: iiiiinterestingly enough:

we upgraded to lift 3.3 and our tests started failing. adding `exclude("org.mozilla", "rhino"),` to our lift dependencies got 'em to pass O___o.

not sure if we're just postponing the inevitable but i found it noteworthy.

arigoldx: that's what got me asking about _why_ this updated dependency -- we were only successful after we reverted it.

farmdawgnation: It's possible that you have other dependencies that rely on there being an older version of rhino present and Lift bringing in a newer version transitively causes issues for you. I'd recommend taking a look at what other dependencies might ask for Rhino (or look for any direct dependencies on it in your project) to try and determine whether or not that's what is happening. If you use Rhino directly, there's a good chance you'll have code to update.

Shadowfiend: There is direct usage in that codebase fwiw 😉

arigoldx: > It's possible that you have other dependencies that rely on there being an older version of rhino present

ooh yeah. I think that's why it the `exclude` worked, yah?

> and Lift bringing in a newer version transitively causes issues for you.

arigoldx: > It's possible that you have other dependencies that rely on there being an older version of rhino present and Lift bringing in a newer version transitively causes issues for you.

ooh yeah. I think that's why it the `exclude` worked, yah?

I suppose I expected something Rhino-y in this PR. That is, instead of just bumping a version, maybe something like `rhino.oldFunction -> rhino.newFunction`. On the other hand, version bumps to stay current -- with no code changes -- are also legit.

> There is direct usage in that codebase fwiw 😉

'preciate it :D Suppose I might have to Actually Dig In:tm: & learn myself some Clojure 😎.

arigoldx: Caveat -- I say "Clojure" 'cause I think that's how we're using it. Prolly shoulda [read the README first](https://github.com/mozilla/rhino) 😅 .

PR #1953, opened 2018-11-18
Add a JDK11 and JDK12 build to the Travis matrix

farmdawgnation:
Given that JDK 11 is the new LTS release of the JDK, we should ensure our tests pass in it. The runtime bytecode is guaranteed to be compatible, but there are some changes to the default classpath that could affect our behavior.


farmdawgnation:
Ah, right. sbt.

Shadowfiend:
🙃

farmdawgnation:
Sure enough we're failing on JDK11!

```
[error] Error: Total 309, Failed 0, Errors 1, Passed 308, Pending 27
[error] Error during tests:
[error] 	net.liftweb.mongodb.record.MongoFieldSpec
[error] (lift-webkit / Test / compileIncremental) Compilation failed
[error] (lift-mongodb-record / Test / test) sbt.TestsFailedException: Tests unsuccessful
```

I'm taking a look.

farmdawgnation:
Couldn't repro the mongodb record failure locally so I'm guessing that was transient. Fixed the compile issues in webkit.

farmdawgnation:
Holy crap the build is green

farmdawgnation:
Ok, JDK 12 is now _also_ in the mix. JDK 13 is fresh off the presses but we'll give it a bit before we try that out.

.travis.yml
```diff
@@ -1,39 +1,34 @@
 language: scala
-
 dist: trusty
-
 sudo: false
-
 scala:
   - 2.11.11
```
Shadowfiend: Looks like `2.11.12` is a thing?

web/webkit/src/test/scala/net/liftweb/http/HtmlNormalizerSpec.scala
```diff
@@ -153,7 +153,7 @@ class HtmlNormalizerSpec extends Specification with XmlMatchers with Mockito {
 
       js.toJsCmd must be matching("""(?s)\Qlift.onEvent("lift-event-js-\E[^"]+\Q","event",function(event) {doStuff;});
         |lift.onEvent("hello","event",function(event) {doStuff2;});
-        |lift.onEvent("lift-event-js-\E[^"]+\Q","event",function(event) {doStuff3;});\E""".stripMargin('|').lines.mkString("\n").r
+        |lift.onEvent("lift-event-js-\E[^"]+\Q","event",function(event) {doStuff3;});\E""".stripMargin('|').linesIterator.mkString("\n").r
```
Shadowfiend: This turns out to be because of this guy: https://github.com/scala/bug/issues/11125

A note from that issue:

> note that it isn't safe to rely on the compiler to catch bad uses of .lines, you want to search your source code and find all of them
> 
> yes, often the compiler will catch it, e.g. s.lines.toVector fails to compile because Java's Stream doesn't have toVector
> 
> but more devious is something like s.lines.filter(...) == .... Java's Stream has filter, and == isn't type-safe, so if you're lucky you'll get a "are unrelated: they will most likely never compare equal" warning ... if you're lucky... and if you're lucky enough to spot the warning in your code's big pile of warnings it already generates that you tune out now...
> 
> I guess a Scalafix rewrite could help here, and wouldn't have false positives if you are using .lines method on something other than a String

build.sbt
```diff
@@ -7,8 +7,8 @@ homepage in ThisBuild              := Some(url("http://www.liftweb.net"))
 licenses in ThisBuild              += ("Apache License, Version 2.0", url("http://www.apache.org/licenses/LICENSE-2.0.txt"))
 startYear in ThisBuild             := Some(2006)
 organizationName in ThisBuild      := "WorldWide Conferencing, LLC"
-scalaVersion in ThisBuild          := "2.12.6"
-crossScalaVersions in ThisBuild    := Seq("2.12.6", "2.11.11")
+scalaVersion in ThisBuild          := "2.12.8"
+crossScalaVersions in ThisBuild    := Seq("2.12.8", "2.11.11")
```
Shadowfiend: Ditto re: 2.11.12.

farmdawgnation: I'm happy to version bump that later but this has been languishing awhile, let's not hold it up.

.travis.yml
farmdawgnation: I'm happy to version bump that later but this has been languishing awhile, let's not hold it up.

web/webkit/src/test/scala/net/liftweb/http/HtmlNormalizerSpec.scala
farmdawgnation: Oy vey.

farmdawgnation: Oh I just realized this probably means we've got to hunt all uses of lines down. ffffff

Shadowfiend: It feels like I can pick this particular issue up in the 2.13 PR, since I *think* that'll catch issues? Not sure.

PR #1958, opened 2019-04-02
Upgrade nu.validator html parser

nodefactory-bk:
Liftweb has a slightly ancient nu.validator version.

Specifically it breaks vuetify data-table since it removes <td> tags
inside <template>.

I know I haven't discussed this on the mailing list but since the change is small and passes all lift tests I'm hoping I don't get shot for it =)

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -36,7 +36,7 @@ object Dependencies {
   lazy val jbcrypt                = "org.mindrot"                % "jbcrypt"            % "0.4"
   lazy val joda_time              = "joda-time"                  % "joda-time"          % "2.10"
   lazy val joda_convert           = "org.joda"                   % "joda-convert"       % "2.1"
-  lazy val htmlparser             = "nu.validator.htmlparser"    % "htmlparser"         % "1.4"
+  lazy val htmlparser             = "nu.validator"               % "htmlparser"         % "1.4.12"
   lazy val mongo_java_driver      = "org.mongodb"                % "mongodb-driver"     % "3.7.1"
   lazy val mongo_java_driver_async  = "org.mongodb"              % "mongodb-driver-async" % "3.7.1"
   lazy val paranamer              = "com.thoughtworks.paranamer" % "paranamer"          % "2.8"


Shadowfiend:
I'm having trouble finding who is publishing these point releases. The [original site](https://about.validator.nu/htmlparser/) notes the last release as 1.4, and the [Mercurial repo](https://hg.mozilla.org/projects/htmlparser) seems to have more recent commits but its last tag is for the 1.4 release. Is it [this fork](https://github.com/validator/htmlparser)? What does 1.4.12 include with respect to 1.4?

andreak:
Why not upgrade to latest from https://github.com/validator/validator ?

farmdawgnation:
It looks to be that fork, but they're not tagging releases and may not be pushing the commits that have the version bumps at all? I'm inclined to think this is safe but have very little to go on in terms of validating that this is the same code that's getting released. @sideshowbarker seems to be the maintainer of the fork — maybe he can shed some light on the validity of making this change / dangers thereof?

sideshowbarker:
Assuming the releases being referred to here at the maven packages with groupId=nu.validator and artifact=htmlparser:

https://repo1.maven.org/maven2/nu/validator/htmlparser/

…then yeah I am the one doing those releases.

As far as those releases being safe to use: The same code is what I use as the HTML parser for the W3C Nu Html Checker https://validator.w3.org/nu/ — so it’s functioning in that context.

I have not used it myself outside that context, but I don’t have reason to believe it’d have any problems.

One difference from upstream that you should be aware of is that the fork doesn’t expose an XOM API. But that shouldn’t affect you unless (unless of course if you’re actually using the XOM API).

There are some other differences from upstream:

https://github.com/validator/htmlparser/compare/master...validator-nu

…but all the differences there except for the *Add .classpath, .settings, .project, .mailmap* change are changes to bring the parser into conformance with the current requirements in the HTML spec.

As far as tagging, going forward I can start tagging every time I do a release. The version numbers just reflect that fact I never bump the minor version number (the 1.4) but always only the patch number.

That seems appropriate since the fork introduces no API changes (well, other than completely dropping the XOM API, but I’m not aware of anybody that’s been using that anyway).

farmdawgnation:
Awesome, thanks. I think with some certification that these builds are legit, I'm fine to start using it provided things work. The CI build is failing, so I'm not sure what's up with that offhand. Will try to take a look at it before too long.

Shadowfiend:
Looks like it was one of our favorite `TimeHelpers` errors. Restarted the build.

PR #1959, opened 2019-04-23
Fix crudify render all

ReviakinAleksey:
**Mailing List thread**: https://groups.google.com/forum/#!topic/liftweb/as0h7yPlyoY

Template rendered for `crud.all` Snippet in net.liftweb.proto.Crudify render empty `td` elements instead of `TheCrudType` filed values.
This PR introduce issue fix and test for this functional.

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -281,3 +281,6 @@ Henrik Härkönen
 
 ### Email: ###
 heharkon@iki.fi
+
+### Email: ###
+it dot blackdog at gmail dot com
--- a/persistence/proto/src/main/scala/net/liftweb/proto/Crudify.scala
+++ b/persistence/proto/src/main/scala/net/liftweb/proto/Crudify.scala
@@ -644,7 +644,7 @@ trait Crudify {
         pointer <- fieldsForList
         field <- computeFieldFromPointer(c, pointer).toList
       } yield {
-        ".value *" #> field.asHtml
+        ".row-item *" #> field.asHtml
       }
     }
   }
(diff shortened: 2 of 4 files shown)


Shadowfiend:
@ReviakinAleksey sorry for taking so long to reply. This is a great PR! Love the new test 😍 Only issue I see is it looks like the test isn’t compiling in CI: https://travis-ci.org/lift/framework/jobs/523455236#L2271

ReviakinAleksey:
> @ReviakinAleksey sorry for taking so long to reply. This is a great PR! Love the new test 😍 Only issue I see is it looks like the test isn’t compiling in CI: https://travis-ci.org/lift/framework/jobs/523455236#L2271

Hello @Shadowfiend, unfortunately I only test my code inside IDEA IDE, and mess up PR in first lines:
line number 5 `import net.liftweb.http.{LiftRules, LiftSession, WithLiftContext}` which produce `object WithLiftContext is not a member of package net.liftweb.h` compilation error in `sbt test`.
I use `WithLiftContext` to render `Crudify` snippet in most simplest way.

Could you please give me a help hand with this PR and suggest proper way to test snippet output?

Shadowfiend:
Ah, I see! `WithLiftContext` is part of testkit, which isn't a test dependency of the `proto` module's tests. This'll need some sbt fiddling… Not sure if I'll have time to do it but I'm going to try!

ReviakinAleksey:
I think I found proper solution, test now is using `S.statelessInit(Req.nil)` method instead of `WithLiftContext` trait.

Shadowfiend:
Nice! Awesome adjustment 😎 Let's rock and roll!

PR #1962, opened 2019-06-19
Scala 2.13

Shadowfiend:
https://groups.google.com/d/msg/liftweb/StXaQu_xB2Y/kVauqosGBQAJ

Poking at Scala 2.13, still very much WIP right now and getting a solid cascade of compile errors :tada:

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,8 +7,8 @@ homepage in ThisBuild              := Some(url("http://www.liftweb.net"))
 licenses in ThisBuild              += ("Apache License, Version 2.0", url("http://www.apache.org/licenses/LICENSE-2.0.txt"))
 startYear in ThisBuild             := Some(2006)
 organizationName in ThisBuild      := "WorldWide Conferencing, LLC"
-scalaVersion in ThisBuild          := "2.12.6"
-crossScalaVersions in ThisBuild    := Seq("2.12.6", "2.11.11")
+scalaVersion in ThisBuild          := "2.13.0"
+crossScalaVersions in ThisBuild    := Seq("2.13.0","2.12.6", "2.11.11")
 
 libraryDependencies in ThisBuild ++= Seq(specs2, specs2Matchers, specs2Mock, scalacheck, scalatest)
 
@@ -87,7 +87,7 @@ lazy val json =
     .settings(
       description := "JSON Library",
       parallelExecution in Test := false,
-      libraryDependencies ++= Seq(scalap(scalaVersion.value), paranamer)
+      libraryDependencies ++= Seq(scalap(scalaVersion.value), paranamer, scala_xml)
     )
 
 lazy val documentationHelpers =
--- a/build.sbt
+++ b/build.sbt
@@ -150,6 +150,7 @@ lazy val webkit =
       description := "Webkit Library",
       parallelExecution in Test := false,
       libraryDependencies ++= Seq(
+        scala_parallel_collections,
         commons_fileupload,
         rhino,
         servlet_api,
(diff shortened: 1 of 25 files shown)


eltimn:
I wonder if [Scala Steward](https://github.com/fthomas/scala-steward) would be helpful here. Has anyone tried using that at all?

Shadowfiend:
I haven't, though often “keeping libraries up to date” is incompatible with “staying backwards compatible” in the Scala world. That said, it's probably worth a shot.

I don't think it'll help for this *particular* upgrade though, since it doesn't seem to do much beyond poking you to update to a particular version of a library, and the specific issue here is that it's compile-error-o'clock :grin:

Shadowfiend:
Sweet, looks like scalaz had a 7.2 release compatible with Scala 2.13 since I opened this, which takes care of a wide swath of the issues 🙌

Shadowfiend:
'Bout to dive into webkit. Rebased onto latest master, most things were pretty straightforward. However, this guy seems like a regression that may be worth reporting upstream:

```
package net.liftweb
package util

import java.util.concurrent._
import Helpers.TimeSpan
```

Here, `Helpers` is resolved as `java.util.concurrent.Helpers`, which is *not accessible*. Instead, it should be resolving to `net.liftweb.util.Helpers`, which is both in scope *and* accessible. For now I've disambiguated by prefixing with `util.`.

Shadowfiend:
Heh. Forging full-steam ahead of course I forgot the possibility for 2.13 vs <2.13 source incompatibility. A few places we'll be able to get around that, some others we may need to get creative on forking sources or similar. More to come, though possibly not today.

Shadowfiend:
Fixed a mistake I made with implicits in `Box`, and got `lift-webkit` cooperating. Couple of notes:
- Scala `Buffer`s (and a few other collection types I think) lost `--` and `-` in Scala 2.13. Unclear if that was intentional, need to open an issue on the Scala repo.
- The Scala 2.13 compiler seems to be a little wobblier about disambiguating certain implicits and method signatures. That led to [this `flatMap` invocation change](https://github.com/lift/framework/pull/1962/commits/5ba6943a3fe08ab709e81fccb1ccbbc12b7aa268#diff-b03ffa476fa6a48ce2212116c4094e15R2326), and is also triggering some interesting issues for `lift-mapper` that I haven't dug into super-deeply yet:
```
[error] …/lift/framework/persistence/mapper/src/main/scala/net/liftweb/mapper/ProtoTag.scala:36:7: ambiguous reference to overloaded definition,
[error] both method find in trait KeyedMetaMapper of type (by: net.liftweb.mapper.QueryParam[ModelType]*)net.liftweb.common.Box[ModelType]
[error] and  method find in trait KeyedMetaMapper of type (key: Any)net.liftweb.common.Box[ModelType]
[error] match argument types (net.liftweb.mapper.Cmp[ModelType,String])
[error]       find(By(name, tag)) match {
[error]       ^
```

Will tackle those next.

farmdawgnation:
Just circling back to a thought here: per-module addition of 2.13 support would probably be more manageable and reviewable. I'm also happy to deploy only some modules as 2.13 if we decide to cut a release before everything is ported over.

arashi01:
@Shadowfiend if there is any legwork needed to get per-module 2.13 support happy to help out too.

Shadowfiend:
Two things need to happen, I think:
- We need a variant of `crossScalaVersions` that includes 2.13 and one that doesn't. We'll override per-module using `.settings(crossScalaVersions := supports213)` or similar.
- I need to do some history massaging to get decent commit messages and separation of commits since the current WIP history was left bananas while I figured things out.

If you have cycles, that would be great---I'm not sure if I will before next week at the latest (and obviously I haven't the past few weeks). I think if you want to tinker then open a new PR with some per-module magic, go for it. Most of the base modules are ~easy, and lift-json in particular was a cakewalk (just an import tweak).

Shadowfiend:
Closing this in favor of #1973, which I just merged.

PR #1971, opened 2019-09-05
2.13 Incremental - Extended

arashi01:
**[Mailing List](https://groups.google.com/forum/#!topic/liftweb/dDef6oO2mFM) thread**:

- Allow Scala 2.13 to be removed from builds on a per-module basis (Initially disable Scala 2.13 for all modules). Implemented as an sbt AutoPlugin to allow adding `.noScala13` to project instances to disable Scala 2.13.

- Upgrade Specs2 to current release. Update tests for compatibility.

- Update dependencies to versions with Scala 2.13 availability.

- Bump sbt from 1.2.6 to 1.2.8

See also #1962


Shadowfiend:
Thanks for picking up this work! A couple of thoughts/questions about the approach:

- At a glance, it feels like this is less obvious and more code to delete later vs simply defining an additional variable in `build.sbt` (say, `scala212Only` and adding a `.settings(crossScalaVersions := scala212Only)` directly in each build that doesn't use the default. Quite possible I'm missing a niggly detail that makes this not work, but in that vein, what motivated the plugin approach?
-  Also, was there a particular reason you went with 2.13 as a default that needed to be overridden? Not a huge deal, just curious.
- Lastly, this should be multiple atomic commits vs one large commit with several largely orthogonal changes.

Thanks again for putting in the cycles here; I should be able to give quicker feedback this week until we get this merged.

Shadowfiend:
Had a few extra cycles, so I illustrated the alternate approach in #1973 , with a few thoughts in that PR description. Happy to discuss, as this approach may have advantages I don't understand---and commits in that PR are fair game for cherry-picking here if this is the better approach.

arashi01:
@Shadowfiend okay great. Regarding the plugin vs alternate approach, either should work fine. To cover the points specifically

- For me the plugin approach seemed simpler as it is (for me) least invasive to the main build.sbt build file. Since it will be in use _hopefully_ only temporarily the idea is that once the `noScala13` calls are removed for all modules we simply delete the plugin section.
- Only reason I went with Scala 2.13 for `scalaVersion` is that is what I saw set as default in the other pull request 2.13 branch so I assumed that policy was to target latest supported Scala version as default.

I have no problem picking up from your approach at #1973 though. Both should be perfectly fine. I should have a fair amount of time later today to look into it. Need to investigate issues with Scala 2.11 tests I'm experiencing locally when running the testsuite also.

arashi01:
@Shadowfiend Managed to get some time to work on getting a few more modules working. Based off your branch at  #1973 so should be easy enough to cherry-pick whatever you find useful.

So far have everything except persistence projects and lift-json-scalaz7 compiling and testing against Scala 2.13.

Have a test failing locally in TimeHelpersSpec which I need to investigate. Also testsuite hangs on Scala 2.11 with lift-util and lift-webkit (maybe more?).

arashi01:
Got to enabling a few more modules. Remaining modules _not_ enabled for Scala 2.13:
- lift-json-scalaz7
- lift-mapper
- lift-squeryl-record
- lift-mongodb
- lift-mongodb-record

Shadowfiend:
Pushed a quick fix to ensure we're using Scala 2.12.9 on all the 2.12 Travis builds (I fixed the main build, but not the OpenJDK 11+12 builds). The non-2.11 test failures seem to be around some flappy tests, so we'll see if they keep complaining. I'm poking at 2.11 locally to see if I can reproduce.

Shadowfiend:
The Scala 2.11 issue is an instance of scala/scala-parser-combinators#197. Going to see if I can fork tests for lift-webkit and fix it.

Shadowfiend:
Okayyyy in my branch, trying a version bump to sbt 1.3.0 and specs2 4.7.0, as well as dropping a workaround for a bug that was ostensibly fixed in 2.12.9. If all of that works out, I'll pull the commits into this branch.

Shadowfiend:
@arashi01 I ended up having to go pretty deep on this branch in the end, so I'm going to close this PR in favor of #1973 after all. Mostly it's this branch + a bunch of extra commits fiddling with silliness.

arashi01:
@Shadowfiend Okay great.

PR #1973, opened 2019-09-07
2.13 incremental

Shadowfiend:
This is an incremental setup based on work by @arashi01 that allows
for only certain modules to have 2.13 builds enabled. As of now, these
are almost all modules except for lift-json-scalaz7, lift-mapper,
lift-squeryl-record, lift-mongodb, and lift-mongodb-record.

Build-related updates include:
- Version bumps for specs2, sbt, scala-xml, scala-parser-combinators, and scalatest.
- Proper travis builds for 2.13.
- A deliberate version-bump strategy of “minimum needed to compile”. It felt like further bumping should be a separate thread.


Shadowfiend:
Poking at Travis to try and get a build triggered, hold tight folks <_<

Shadowfiend:
Still a confusing compilation failure in 2.13.0/OpenJDK 11/12 vs 8… Not sure what's up there yet, though the error itself is trivially fixable.

We're also running into an issue with 2.13 where specs2 specs declared as singletons (`object .. extends Specification`) don't have their specs run.

Shadowfiend:
Ah, I know why the JDK issue is coming up. JDK 11/12 have an inaccessible `Helpers` class in `java.util.concurrent`. In `lift-utill`'s `Schedule` class, a reference to `Helpers` in the context of a wildcard `java.util.concurrent._` import is leading the compiler to try to resolve the inaccessible `Helpers` class instead of the package-local `net.liftweb.util.Helpers`. We can import explicitly to avoid this.

Shadowfiend:
Looks like going singletons->classes broke more things, of course, and across Scala versions to boot 😬

Shadowfiend:
Ok, so there is a remaining issue in Scala 2.13 for `lift-json`:

```
[error] ! Multidimensional array extraction example
[error]  net.liftweb.json.MappingException: Can't find constructor for class net.liftweb.json.MultiDim (Meta.scala:227)
[error] net.liftweb.json.Meta$.fail(Meta.scala:227)
[error] net.liftweb.json.ScalaSigReader$.$anonfun$readConstructor$1(ScalaSig.scala:25)
[error] net.liftweb.json.ScalaSigReader$.readConstructor(ScalaSig.scala:25)
[error] net.liftweb.json.Meta$Reflection$.term$1(Meta.scala:334)
[error] net.liftweb.json.Meta$Reflection$.typeParameters(Meta.scala:351)
…
```

Unfortunately, this seems to be happening due to an underlying change in behavior in scalap. I've opened scala/bug#11747 to see if we can figure out the underlying cause.

Shadowfiend:
Ok, looks like we're in passing territory here.

persistence/record/src/main/scala/net/liftweb/record/MetaRecord.scala
```diff
@@ -148,7 +148,7 @@ trait MetaRecord[BaseRecord <: Record[BaseRecord]] {
 
     fieldList = {
       val ordered = fieldOrder.flatMap(f => tArray.find(_.metaField == f))
-      ordered ++ (tArray -- ordered)
+      ordered ++ (tArray --= ordered)
```
Shadowfiend: This is a weird one that we should look at in a follow-up PR (and before we cut a non-SNAPSHOT release from this), I think.

project/plugins.sbt
```diff
@@ -2,7 +2,7 @@ DefaultOptions.addPluginResolvers
 resolvers += Resolver.typesafeRepo("releases")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-web" % "1.4.3")
-addSbtPlugin("org.jetbrains" % "sbt-idea-plugin" % "2.1.3")
+//addSbtPlugin("org.jetbrains" % "sbt-idea-plugin" % "2.1.3")
```
Shadowfiend: We should either uncomment or delete this, I think, but that can happen in a follow-up.

Shadowfiend:
@farmdawgnation if you have a chance, would love a glance here. The diff isn't huge, but I think the main focus areas for this first PR to go out the door are the approach to 2.13 vs 2.12-only modules in build.sbt, the tweaks in `Box`, and the `NamedCometDispatcher` version fork.

arashi01:
@Shadowfiend Maybe try bump to Scala 2.13.1 / 2.12.10, and sbt 1.3.2 while we're at it? Fingers crossed it should go okay!

andreak:
Keep on rocking!

PR #1976, opened 2019-10-15
Delegate equals comparison in Record to Field

### LINKED ISSUE #1975: Delegate equals comparison in Record to Field
Currently, Record.equals directly compares the `valueBox` and `name` of each field. This makes it impossible to override that functionality in individual Fields. Because of this, PatternField does not work properly in this regard.

The proposed solution is to delegate the comparison to the individual Field, using the current code as the default.

eltimn:
Fixes #1975

### DIFF
--- a/persistence/mongodb-record/src/main/scala/net/liftweb/mongodb/record/BsonRecord.scala
+++ b/persistence/mongodb-record/src/main/scala/net/liftweb/mongodb/record/BsonRecord.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2011-2018 WorldWide Conferencing, LLC
+ * Copyright 2011-2019 WorldWide Conferencing, LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


persistence/mongodb-record/src/main/scala/net/liftweb/mongodb/record/BsonRecord.scala
```diff
@@ -57,6 +57,7 @@ trait BsonRecord[MyType <: BsonRecord[MyType]] extends Record[MyType] {
   /**
     * Pattern.equals doesn't work properly so it needs a special check. If you use PatternField, be sure to override equals with this.
     */
+  @deprecated("PatternField now has a properly functioning `equals` method.", "3.4.0")
```
farmdawgnation:
```suggestion
  @deprecated("PatternField now has a properly functioning `equals` method.", "3.4.1")
```

eltimn:
Uh, I just noticed that commit suggestion button. I haven't used that yet. Looks really handy.

PR #1977, opened 2019-10-18
Update lift-mongodb and lift-mongodb-record for deprecations in the m…

### LINKED ISSUE #1830: Deprecate use of com.mongodb.DBObject
The mongo-java-driver has deprecated many classes in addition to DBObject. Need to prepare for when those don't exist any more.

eltimn:
…ongo-java-driver.

Fixes #1830 

The mongo-java-driver has deprecated a lot of their classes and are going to remove them in the next release. This is long overdue.

One of the big changes is the use of a codec model for encoding and decoding data to/from the db into a language type. I took advantage of that to change how Records are stored. A RecordCodec is now used, which means data is put directly into a Record instead of first being put into a DBObject (or the newer Document) and then converting from there. I didn't do any benchmarks, but I would think that's a decent performance improvement.

I'm not 100% sure if I'm done with this yet, but I wanted to get this up. I need to set it aside for a couple of weeks while I get back to my job. The main thing I'm still thinking about doing is replacing the BsonParser with a Codec for JValue. I have a lot of it written already, but I'm not sure it's worth the effort to finish it.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -257,8 +257,9 @@ lazy val mongodb =
   persistenceProject("mongodb")
     .dependsOn(json_ext, util)
     .settings(
+      crossScalaVersions := crossUpTo213,
       parallelExecution in Test := false,
-      libraryDependencies ++= Seq(mongo_java_driver, mongo_java_driver_async),
+      libraryDependencies ++= Seq(mongo_scala_driver, mongo_java_driver, mongo_java_driver_async),
       initialize in Test := {
         System.setProperty(
           "java.util.logging.config.file",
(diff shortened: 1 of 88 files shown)


farmdawgnation:
Any word here Tim?

eltimn:
Got real busy at work. I'm hoping to get back to this in the next couple of weeks.

farmdawgnation:
No judgement here. Toddler is keeping me plenty busy too. :joy:

eltimn:
Superceded by #1984

PR #1978, opened 2019-12-28
Disable the js auto-rewrite

farmdawgnation:
**Mailing List thread**:
https://groups.google.com/d/msg/liftweb/urqpaxe-0U8/204X5JzqDQAJ

This PR removes the auto-rewriting we do for minified JS files.

In previous versions of Lift we would minify the JS files in advance using YUI compressor. This fell by the wayside with the recent sbt upgrades due to compatibility issues. Newer solutions, e.g. sbt-web, seem to favor doing the minification on the build of the final web artifact and combining everything into a single file so I've had some difficulty getting it to play nice with our existing JS files.

In an effort to get a working Lift build again, I've favored disabling the automatic rewrite until we can revisit the issue in more depth.

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/ResourceServer.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/ResourceServer.scala
@@ -47,9 +47,12 @@ object ResourceServer {
   }
 
   @volatile var pathRewriter: PartialFunction[List[String], List[String]] = rewriter orElse {
-    case "lift.js" :: Nil => List("lift-min.js")
-    case "json.js" :: Nil => List("json2-min.js")
-    case "json2.js" :: Nil => List("json2-min.js")
+    // These automatic rewrites have been disabled since we're not currently
+    // minifying our internal js files on build. There's an open question in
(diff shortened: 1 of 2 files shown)


Shadowfiend:
I think we have tests that make sure the rewrites are in place, so you'll need to `.pendingUntilFixed` those or something.

I wish there was a quick, straightforward way to fix this, but I think it would involve gluing the `sbt-web` minification stuff with the `resourceManaged` magic that the YUI compressor plugin does… Pretty yakky, and probably time-consuming. I'm okay with sticking to this solution until someone has time to do a proper pass.

eltimn:
Since this is super easy to do with Node, I wrote a quick npm script to minify the js files. Your call whether we want to use it or not, but I figured this was better than removing the minified files, if only until someone wants to continue fighting sbt.

#1979

farmdawgnation:
Commented on the other PR as well, but I think once we have the npmsh bits wired into the package task for the webkit build then we're good to go. I don't want to introduce a change that requires me to remember extra steps during the release process because... I'm very likely to forget them.

PR #1979, opened 2019-12-29
Added minify npm script

eltimn:
In case we want to use this instead of removing the minified js files.

### DIFF
--- a/npmsh
+++ b/npmsh
@@ -3,6 +3,8 @@
 set -ev
 
 cd web/webkit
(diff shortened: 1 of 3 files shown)


farmdawgnation:
I think this is going to end up being the correct solution once it's linked into the sbt package task. In the meantime I'm going to go ahead and merge my change to disable the minification so that the webkit starts working again. Also since I don't know when I'll have a chance to dig in here again. 😬

PR #1982, opened 2020-05-31
Lift-mapper with Scala 2.13 support

lvitaly:
- Lift-mapper was updated for supporting Scala 2.13
- All deprecation warnings of compiler were fixed
- Scala versions were bumped to 2.12.11 and 2.13.2
- Sbt version was bumped to 1.3.12

### DIFF
--- a/persistence/mapper/src/main/scala-2.13/net/liftweb/mapper/ManyToMany.scala
+++ b/persistence/mapper/src/main/scala-2.13/net/liftweb/mapper/ManyToMany.scala
@@ -1,0 +1,241 @@
+/*
+ * Copyright 2006-2011 WorldWide Conferencing, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.liftweb
+package mapper
+
+import common._
+
+import scala.annotation.tailrec
+import scala.language.existentials
+
+/**
+ * Add this trait to a Mapper to add support for many-to-many relationships
+ * @author nafg
+ */
+trait ManyToMany extends BaseKeyedMapper {
+  this: KeyedMapper[_, _] =>
+
+  private[this] type K = TheKeyType
+  private[this] type T = KeyedMapperType
+
+  private var manyToManyFields: List[MappedManyToMany[_,_,_]] = Nil
+
+  /**
+   * An override for save to propagate the save to all children
+   * of this parent.
+   * Returns false as soon as the parent or a one-to-many field returns false.
+   * If they are all successful returns true.
+   */
+  abstract override def save: Boolean = {
+    super.save && manyToManyFields.forall(_.save)
+  }
+
+  /**
+   * An override for delete_! to propogate the deletion to all children
+   * of this parent.
+   * Returns false as soon as the parent or a one-to-many field returns false.
+   * If they are all successful returns true.
+   */
+  abstract override def delete_! : Boolean = {
+    super.delete_! &&
+      manyToManyFields.forall( _.delete_!)
+  }
+
+
+  /**
+   * This is the base class to extend for fields that track many-to-many relationships.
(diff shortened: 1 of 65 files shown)


persistence/mapper/src/main/scala-2.13/net/liftweb/mapper/ManyToMany.scala
```diff
@@ -83,85 +83,79 @@ trait ManyToMany extends BaseKeyedMapper {
...
+            otherFK(newJoin)(_.apply(e))
+            newJoin
+          }{ removedJoin =>
+            removedJoins = removedJoins filter removedJoin.ne
+            removedJoin // well, noLongerRemovedJoin...
           }
-        case Some(join) =>
-          join
-      }
+      }(join => join)
```
farmdawgnation, line 112: What tests is this block covered under?

persistence/mapper/src/main/scala/net/liftweb/mapper/MappedField.scala
```diff
@@ -363,20 +356,18 @@ trait MappedField[FieldType <: Any,OwnerType <: Mapper[OwnerType]] extends Typed
...
    * field.set(new_value) <br />
    * are all the same
    */
-  def update[Q <% FieldType](v: Q) {
+  def update[Q](v: Q)(implicit implFn: Q => FieldType): Unit = {
     this.set(v)
   }
 
-  def apply[Q <% FieldType](v: Q): OwnerType = {
+  def apply[Q](v: Q)(implicit implFn: Q => FieldType): OwnerType = {
```
farmdawgnation, line 370: Why is the implicit required in this new version?

persistence/mapper/src/main/scala/net/liftweb/mapper/MetaMapper.scala
```diff
@@ -1672,13 +1666,11 @@ object In {
...
-            OuterMapper <: Mapper[OuterMapper]]
-  (_outerField: MappedField[JoinTypeA, OuterMapper],
-   _innerField: MappedField[JoinTypeA, InnerMapper],
-   qp: Zoom*): InThing[OuterMapper]
-  = {
+  def apply[InnerMapper <: Mapper[InnerMapper], JoinTypeA, Zoom, OuterMapper <: Mapper[OuterMapper]](
+    _outerField: MappedField[JoinTypeA, OuterMapper],
+    _innerField: MappedField[JoinTypeA, InnerMapper],
+    qp: Zoom*
+  )(implicit ev: Zoom => QueryParam[InnerMapper]): InThing[OuterMapper] = {
```
farmdawgnation, line 1673: Looks like we've also added implicits here that I'm not totally clear on the use of.

farmdawgnation:
Sorry it took so long for me to get around to this. Personal life has been a bit of a mess lately so I appreciate the patience. I've left some questions about some implicits I saw added in particular. Otherwise things look Fine™, but I'm not a lift-mapper expert so I'm not sure I'm the best judge. I'd love to know how confident you are that the changes you made are covered by tests.

persistence/mapper/src/main/scala/net/liftweb/mapper/MappedField.scala
lvitaly: `Q <% FieldType` is a view bound syntax which is deprecated in Scala 2.13.x

> * **View bounds deprecated**
>   * Deprecated: A <% B (#6500)
>   * Use instead: (implicit ev: A => B)

https://github.com/scala/scala/releases/tag/v2.13.0

persistence/mapper/src/main/scala/net/liftweb/mapper/MetaMapper.scala
lvitaly: As previously described, the deprecated view bound syntax was replaced by implicit function.

> * **View bounds deprecated**
>   * Deprecated: A <% B (#6500)
>   * Use instead: (implicit ev: A => B)

https://github.com/scala/scala/releases/tag/v2.13.0

persistence/mapper/src/main/scala/net/liftweb/mapper/MappedField.scala
farmdawgnation: Ah, good to know

ShaneDelmore:
Just a heads up, if this doesn't land for a few more days you may want to retest with Scala 2.12.12.   I ran some json tests against Scala 2.12.11 that succeed, but in 2.12.12-bin-cbe6ade (the RC for 2.12.12).  Unfortunately I don't have a detailed repro, just had enough time to dig a little bit and see that lazy val symbols: ScalaSigParsers.symbols is null in at least one case, which appears to be an unexpected condition.  I tested with lift-json 3.4.1 and have no idea if the issue is present with these changes or not.

PR #1988, opened 2020-12-13
Final Scala 2.13 Pieces: documentation-helpers, json-scalaz7

farmdawgnation:
This PR updates documentation-helpers and  json-scalaz7 to support Scala 2.13, and gets us to a ~full 2.13 build of the entire framework~ a 2.13 build excluding squeryl-record. There's a bit of build file cleanup we'll want to do after this is merged, but in favor of keeping the PR easy to understand I focused on expanding the existing patterns we have to 2.13 support.


eltimn (approved):
I don't really know much about scalaz, but l trust you know what you're doing.

PR #1990, opened 2021-04-06
Adding support for SameSite cookies

csaltos:
**[Mailing List](https://groups.google.com/g/liftweb/c/VzWNc7PWdIo) thread**:

SameSite is a required field for nowadays web security and since Servlet API does not support it yet, this PR tries to give it support at Scala level for Lift via Netty.

We are going to use this patch for Lift at Talenteca.com and I hope you find this PR valuable for other people too.

Best regards,

Carlos Saltos

### DIFF
--- a/web/webkit/src/main/scala/net/liftweb/http/provider/HTTPCookie.scala
+++ b/web/webkit/src/main/scala/net/liftweb/http/provider/HTTPCookie.scala
@@ -50,10 +50,17 @@ case class HTTPCookie(name: String,
 
   /**
    * Returns a new HTTPCookie that preserve existing member values but sets the httpOnly attribute
-   * @param flagHttpOnly - should the cookie be flagged as HTTP Only (only works in Servlet 3.0 containers)
+   * @param flagHttpOnly - should the cookie be flagged as HTTP Only
    * @return HTTPCookie
    */
   def setHttpOnly(flagHttpOnly: Boolean): HTTPCookie = copy(httpOnly = Full(flagHttpOnly))
+
+  /**
+   * Returns a new HTTPCookie that preserve existing member values but sets the sameSite attribute
+   * @param newSameSite - the new cookie SameSite value
+   * @return HTTPCookie
+   */
+  def setSameSite(newSameSite: SameSite.Value): HTTPCookie = copy(sameSite = Full(newSameSite))
 
   /**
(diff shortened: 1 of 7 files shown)


farmdawgnation (changes requested):
I'm not comfortable adding a netty dependency, but it looks like ServerCookieEncoder doesn't require a ton of other classes. Any chance you're game to implement a modified copy of it in Lift that can stand alone?

csaltos:
Yes, of course, Matt, good point !! ... I will gladly try to do it next week.

csaltos:
Ready dear Matt, now it works with an internal cookie encoder and I also added some tests that are important.

Please let me know if it's OK now ... and if it needs further improvements, please let me know, I will do them gladly.

Best regards,

Carlos Saltos

web/webkit/src/main/scala/net/liftweb/http/provider/HTTPCookie.scala
```diff
@@ -37,19 +41,27 @@ case class HTTPCookie(name: String,
...
+                       sameSite : Box[SameSite.Value] = Empty) extends java.lang.Cloneable {
   override def clone(): HTTPCookie = {
     super.clone()
     copy()
   }
 
   /**
    * Returns a new HTTPCookie that preserve existing member values but sets the httpOnly attribute
-   * @param flagHttpOnly - should the cookie be flagged as HTTP Only (only works in Servlet 3.0 containers)
+   * @param flagHttpOnly - should the cookie be flagged as HTTP Only
```
farmdawgnation, line 53: This change is because we're now encoding our own cookies and are no longer dependent on the servlet encoding?

farmdawgnation (commented):
One question but this looks good overall. Going to give others a day or two to look at it before merging. Thank you!

farmdawgnation (approved):

web/webkit/src/main/scala/net/liftweb/http/provider/HTTPCookie.scala
csaltos: Yes, exactly, now with our own new handling of the cookies we do not have the limitation of the Servlet version ... it just works for all versions.

csaltos:
Thank YOU, dear Matt !! :+1::sunglasses:

csaltos:
Thanks for the merge !! :+1::sunglasses:

PR #2026, opened 2024-11-11
Remove squeryl and JSON scala combinators parser

fanf:
Remove dependencies that are not used anymore but still referenced, causing problems for futur upgrades: 

 - squeryl is a dead project for years now, and it causes import problem in intellij. Since `persistance` module is deleted, there's no reason to keep it, 
 - `JSON` in scala-combinator was removed post-1.1 branch as a "sample, confused for a real parsed" and should not be used. It will make impossible update to newer version of the llib. See: https://github.com/scala/scala-parser-combinators/issues/99

The changes are minimal and all tests pass.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -55,7 +55,7 @@ lazy val core: Seq[ProjectReference] =
 lazy val common =
   coreProject("common")
     .settings(
-      description := "Common Libraties and Utilities",
+      description := "Common Libraries and Utilities",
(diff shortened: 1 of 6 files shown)


pabloazul:
Looks reasonable to me. I built it with sbt 1.6.2 and java 11.0.2.

farmdawgnation (approved):

PR #2031, opened 2025-08-11
feat(lift-common)!: scala 3 support

farmdawgnation:
Implementation of a lift-common build that works in Scala 3.

- [x] Updated existential type with `any`
- [x] Explicitly invoke an `ev` implicit to get the compiler working
- [x] Fix indentation in LRU
- [x] Update dependencies
- [x] Change `logFailure` to public so that overrides work correctly
- [x] Fully replace scalatest-mock since it doesn't exist in Scala 3
- [x] Fix side effects of dependency upgrades throughout lift


dpp (approved):
Awesome!

